PR #4, opened 2016-04-07
Add basic lua implementation

SnirkImmington:
We have a Lua thread! We can ask it to do things! We can register Rust and Lua things for it to do!
* Dependency on `hlua`
* Lua module with thread spawned on `compositor-ready`
* Basic `LuaQuery` and `LuaResponse` enums to send messages to the thread
* `lua::send` to send messages
* Lua thread which receives requests and runs Lua code
* `lib/lua` folder with a utils library and `init.lua` to prepare the thread
* `lua::funcs` module containing Rust functions to be called from Lua (such as in `funcs::input`). This module has `#![deny(dead_code)]` so functions written here will cause a compile error if they are not registered.
* `lua::funcs::init` which adds all the Rust functions into the proper Lua APIs.
* Key binding `C-p` which prints the pointer position using Lua
* Key binding `C-l` which sleeps the Lua thread (demonstrating the main thread continues, Lua requests queue up)

This has the skeleton of the infrastructure we need for way-cooler. This is a huge part of the capability we need with the ability to have keybindings and Lua functions. After expanding on this API (add UUIDs on Lua messages, get back results, add connections to registry) we can start working on the config system, and adding more Rust and Lua functions to our API.


src/lua/funcs/input.rs
```diff
@@ -0,0 +1,31 @@
...
+
+/// Sets the position of the mouse
+pub fn pointer_set_position(in_x: f64, in_y: f64) -> Result<(), String> {
+    if in_x < 0f64 {
+        return Err("Invalid negative x parameter!".to_string());
+    }
+    if in_y < 0f64 {
+        return Err("Invalid negative y parameter!".to_string());
+    }
+    //let &resolution = WlcOutput::focused().get_resolution();
```
Timidger: Should probably delete this comment if it's not needed

src/lua/funcs/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+#[forbid(dead_code)]
+#[forbid(unused_variables)]
+
+mod input;
+
+/// Register all the Rust functions for the lua libraries
+pub fn register_libraries(lua: &mut Lua) {
+
+    // A macro for adding lua functions... could be a method but whatever
+    macro_rules! lua_function {
```
Timidger: If this could be a method, it probably should be. I don't know Rust's stance on that grey line for functions that could be functions or macros, but to take a leaf out of Scheme/Lisp's book I'd say if there is something that could be a function that is instead written as a macro, you should probably write it as a function. Macros are hard to understand, and add an unnecessary level of complexity in certain cases.

Now I don't know Rust's syntax for macros, but from what I can see is that it makes a new function then automatically binds it with lua.set. Can that be a function pointer/lambda of some kind? Because it looks like you pass the function at `hlua::function0($name);`, so why do we need to have it defined? And if so, we could have it be returned as a lambda/function pointer. If you have good reasons to keep it in I'm all for hearing them, but I think this is something important to consider.

SnirkImmington:
Added 05100c5 in response to your comments, removed the macro for now and the comment you commented on.

Given the way we have to manually access the tables in order to register the functions, I don't think a lua macro would be as automatic, unless it took in a reference to which table we register the function with.

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+}
+
+/// Sends a value to the lua thread.
+///
+/// # Panics
+/// * If the lua thread is currently not running (check `lua::thread_running()`)
+/// * If the sender was unable to send, which should only happen
+/// if the lua thread isn't running.
+pub fn send(query: LuaQuery) {
```
Timidger: Do we need to expose this function? `try_send` is already public, if there is a case where some rust code should panic then it can be handled by the caller (by simply unwrapping the result of `try_send`, for example).

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    trace!("Initializing...");
+    let (query_tx, query_rx) = channel::<LuaQuery>();
+    let (answer_tx, answer_rx) = channel::<LuaResponse>();
+    {
+        let mut sender = SENDER.lock().unwrap();
+        let mut receiver = RECEIVER.lock().unwrap();
+
+        *sender = query_tx;
+        *receiver = answer_rx;
+    }
```
Timidger: I'm confused, what is the purpose of this block? Aren't `sender` and `receiver` being dropped at the end of this block? They are locks on the sender/receiver, they are reassigned to `query_tx` and `answer_tx` respectively and then nothing happens?

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    thread::spawn(move || {
```
Timidger: This is a pretty long block of code, should we maybe not make it a lambda? It's not particularly unreadable, just maybe it makes more sense to be a separate function? On the fence about this though, it seems fine

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        trace!("thread: Inside thread!");
+        let sender = answer_tx;
+        let receiver = query_rx;
+        let mut lua = Lua::new();
+        trace!("thread: Loading core libraries...");
+        lua.openlibs();
+        trace!("thread: loading way-cooler library...");
+        lua.execute_from_reader::<(), File>(
+            File::open("lib/lua/init.lua").unwrap()
```
Timidger: Location of init.lua should probably be in a constant that's can easily be configured

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        );
+        trace!("thread: loading Rust libraries...");
+        funcs::register_libraries(&mut lua);
+        trace!("Finished loading libraries!");
+        // Only broadcast running after loading libs
+        *RUNNING.write().unwrap() = true;
+        trace!("thread: Testing hello world...");
+        let mut file = File::create("/tmp/init.lua").unwrap();
```
Timidger: Don't assume there is a tmpfs at /tmp/. Perhaps use a crate like _tempfile_ or some other mechanism to have it write to a valid temporary place that the OS can choose.

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+                None => {
+                    warn!("thread: Unable to get variable {}", varname);
+
+                    thread_send(&sender, LuaResponse::Variable(None));
+                }
+            }
+        },
+
+        LuaQuery::SetValue { name: name, val: val } => {
+        },
```
Timidger: If you're not going to fill these in, might want to make it panic with a not implemented (I think there is a macro to work like that, I used it before)

src/lua/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        LuaQuery::EmptyArray(name) => {
+
```
Timidger: If you're not going to fill these in, might want to make it panic with a not implemented (I think there is a macro to work like that, I used it before)

src/registry.rs
```diff
@@ -60,3 +55,38 @@ pub fn get_or_default(name: &RegKey, value: RegVal) -> RegVal {
     }
 }
 
+/// Set a value in the registry.
+///
+/// If the key already exists, returns the old value.
+/// Returns `None` for new keys.
```
Timidger: These seem like confusing return values. Why should I care what the old value is when I set it?

src/lua/mod.rs
SnirkImmington: So there are four messengers:

1. Lua thread->send
2. Lua thread<- receive
3. outside->send
4. outside<-receive

The outside ones are declared as lazy statics, and the lua thread ones are moved into the thread.

> `let sender = answer_tx;`
> `let receiver = query_rx;`

However, the lazy static macro requires a thing to be initialized. One way to do it is to have them be `Option<Sender>` and `Option<Receiver>`, and ~~~unwrap~~~ `match` the values in `try_send`. I chose to be lazy and initialize them to one end of a channel (and immediately drop the other half) before the thread's `init()` fixed them.

I'll change them to `Option` types.

src/lua/mod.rs
SnirkImmington: This was some testing code I wrote, should definitely be removed before the merge.

src/registry.rs
SnirkImmington: This passes along `HashMap`'s `insert` method's return types. Most of our code will probably ignore it too.

src/lua/mod.rs
Timidger: Ah, I see. Thanks, it would be more clear if it's an Option, though I think a comment would also have sufficed.

src/lua/mod.rs
SnirkImmington: At the moment it's within the source code anyway, can configure

PR #6, opened 2016-04-13
Registry upgrades - into a Lua-compatable Json-encoded key/value store

SnirkImmington:
#### Registry upgrades
Previously, the registry was a key-value store of String => an enum of primitive types. Now, the registry can contain any struct which, using `rustc-serialize`, can convert to a `Json` structure and be `Decode`d back.

In addition, a `registry` table was added to the Lua thread. It has a metatable which allows Lua to access values from it (i.e. `foo = registry["bar"]`). Lua is currently not allowed to mutate values in the registry. Registry values include an "access" enum, which may in the future include a lua-readonly option.

#### More detailed list of commit changes:

- Add dependency to `rustc-serialize`
- Remove direct `libc` dependency 
- Changes registry values to `RegistryValue` struct
- Move registry types into `registry/types.rs`
- Add `lua/funcs/lua_registry.rs` for Lua `registry` metamethods and `Json` -> `AnyLua` conversion
- Fixed many warnings and organized some imports better
- Add some tests to the registry, lua thread, and lua/registry interop

#### Possible additions

_We'll do 'em in `lua-async`_

- [x] More tests never hurt - 5 threads reading
- [X] Add more `HashMap` methods to the registry - `len()` and `remove()`?
- [X] ~~Supposedly, we should be able to key the HashMap using `&str` instead of `&String`~~ Lifetimes don't work with a `'static` HashMap
- [X] Implement lua-readonly and lua-acessable in the Lua registry
- [x] `registry::get` and `registry::get_json` should also return the `RegistryAccess` - now return a tuple


src/lua/funcs/lua_registry.rs
```diff
@@ -0,0 +1,131 @@
...
+use registry::RegistryAccess;
+
+use hlua::{LuaTable};
+use hlua::any::AnyLuaValue;
+
+use rustc_serialize::json::{Json, ToJson};
+use std::ops::Deref;
+pub fn index(table: AnyLuaValue, lua_key: AnyLuaValue)
+             -> Result<AnyLuaValue, &'static str> {
+    if let AnyLuaValue::LuaString(key) = lua_key {
```
Timidger: Is the reason we need this check (instead of encoding this in the function signature) because this is a function that will be called from lua and so it can be any lua value?

src/lua/funcs/lua_registry.rs
```diff
@@ -0,0 +1,131 @@
...
+            Err("No value found for that key!")
+        }
+    }
+    else {
+        Err("The registry can only be indexed with keys!")
+    }
+}
+
+// Prevent lua from changing the registry?
+pub fn new_index(table: AnyLuaValue, key: AnyLuaValue, val: AnyLuaValue)
```
Timidger: If we don't have an implementation for that yet, perhaps we should just leave it out for now so we don't need to remove it later? We can always add it in later.

src/lua/funcs/lua_registry.rs
```diff
@@ -0,0 +1,131 @@
...
+    Err("You can't set the registry!")
+}
+
+fn convert_json(json: Json) -> AnyLuaValue {
+    match json {
+        Json::String(val) => AnyLuaValue::LuaString(val),
+        Json::Boolean(val) => AnyLuaValue::LuaBoolean(val),
+        Json::F64(val) => AnyLuaValue::LuaNumber(val),
+        Json::I64(val) => AnyLuaValue::LuaNumber((val as i32) as f64),
+        Json::U64(val) => AnyLuaValue::LuaNumber((val as u32) as f64),
```
Timidger: These conversions are hilarious

src/lua/funcs/mod.rs
```diff
@@ -8,14 +8,29 @@ use hlua::{Lua, LuaTable, Function};
...
+        let mut wm: LuaTable<_> = lua.get("wm").unwrap();
+        let mut pointer: LuaTable<_> = wm.get("pointer").unwrap();
 
-    pointer.set("get_position", hlua::function0(input::pointer_get_position));
-    pointer.set("set_position", hlua::function2(input::pointer_set_position));
+        pointer.set("get_position", hlua::function0(input::pointer_get_position));
+        pointer.set("set_position", hlua::function2(input::pointer_set_position));
+    }
+
+    // Initialize registry at the end?
```
Timidger: Is this comment necessary :P? Or at least the question mark?

src/lua/funcs/mod.rs
```diff
@@ -8,14 +8,29 @@ use hlua::{Lua, LuaTable, Function};
...
+    init_registry(lua);
+}
+
+#[inline] // It's called once
+fn init_registry(lua: &mut Lua) {
+    let mut reg_table: LuaTable<_> = lua.get("registry").unwrap();
+    let mut meta_reg = reg_table.get_or_create_metatable();
+    meta_reg.set("__metatable", "Turtles all the way down");
+    meta_reg.set("__index", hlua::function2(lua_registry::index));
+    meta_reg.set("__newindex", hlua::function3(lua_registry::new_index));
```
Timidger: Explain all of what you're doing here to me later, I don't understand the meta-table and other hacker you can do in Lua that well at all.

src/lua/mod.rs
```diff
@@ -124,14 +129,17 @@ pub fn init() {
 fn thread_init(sender: Sender<LuaResponse>, receiver: Receiver<LuaQuery>) {
     trace!("thread: initializing.");
     let mut lua = Lua::new();
+    //unsafe {
+    //    hlua_ffi::lua_atpanic(&mut lua.as_mut_lua().0, thread_on_panic);
+    //}
```
Timidger: Why is this code commented out and not deleted?

src/lua/tests.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::time::Duration;
+use std::thread;
+
+use super::*;
+
+#[test]
+fn big_lua_test() {
+    println!("Testing lua thread...");
+    super::init();
+    thread::sleep(Duration::from_secs(5));
```
Timidger: what happens if you don't sleep and try to send immediately? If bad things happen, is there no way to buffer?

src/lua/tests.rs
```diff
@@ -0,0 +1,31 @@
...
+    thread::sleep(Duration::from_millis(500u64));
+    assert!(!thread_running())
+}
+
+fn test_variable() {
+    try_send(LuaQuery::Execute("hello = 'hello world!'".to_string()))
+        .unwrap();
+
+    try_send(LuaQuery::Execute("assert(hello == 'hello world!')".to_string()))
+        .unwrap();
```
Timidger: Does this panic if the assert fails? Or is this just running it on the thread and not caring about the output "value" from the executed line?

src/main.rs
```diff
@@ -22,7 +25,7 @@ mod callbacks;
 mod lua;
 
 /// Callback to route wlc logs into env_logger
-extern "C" fn log_handler(level: LogType, message_ptr: *const libc::c_char) {
+extern "C" fn log_handler(level: LogType, message_ptr: *const i8) {
```
Timidger: Why'd you switch to a raw pointer to an i8? A char in C could be an unsigned or signed char, the standard doesn't specify ([source](http://stackoverflow.com/questions/2054939/is-char-signed-or-unsigned-by-default)). I see no reason why we shouldn't use `libc::c_char` here?

src/lua/funcs/lua_registry.rs
SnirkImmington: I don't think I've tested thoroughly what happens with invalid types on hlua calling Rust functions.
I think I set it up this way because it's the metatable reference, and a metatable could be indexed with anything (similar issues arise with Lua -> Json conversion). 

The upcoming `async-lua

src/registry/tests.rs
```diff
@@ -0,0 +1,111 @@
...
+    x: i32,
+    y: i32
+}
+
+impl ToJson for Point {
+    fn to_json(&self) -> Json {
+        let mut map = BTreeMap::new();
+        map.insert("x".to_string(), Json::I64(self.x as i64));
+        map.insert("y".to_string(), Json::I64(self.y as i64));
+        Json::Object(map)
```
Timidger: Why does the `Json` library make this have to be a BTreeMap? Seems like an odd data structure to use for this, why not just anything that utilizes the `Entry` trait?

src/registry/tests.rs
```diff
@@ -0,0 +1,111 @@
...
+    });
+    let tx5 = tx.clone();
+    thread::spawn(move || {
+        read_thread(String::from("test_point"), point, tx5);
+    });
+
+    let mut result = true;
+
+    for i in 0..5 {
+        result = result && rx.recv().unwrap();
```
Timidger: Ha. Nice. As soon as one fails, the rest won't be read.

src/lua/funcs/lua_registry.rs
SnirkImmington: This uses hlua 0.1.7, which does not have `AnyLuaValue::LuaNil`.

Not implementing the metatable's new_index would allow scripts to write data onto the registry table. I _think_ the `index` metamethod would be called before the data was accessed, but this provides a solid error for now and lets us figure out permissions later.

src/lua/funcs/mod.rs
SnirkImmington: Registry was already initialized in utils.lua, so it's not necessary :P

src/lua/funcs/mod.rs
SnirkImmington: We are overloading three metatable methods: `__index`, `__index_new`, and `__metatable`.

So a metatable (for the purposes of the registry table) is basically a list of operators. For example, to make the registry look like a Lua table, we use the metatable to implement it's "operator\_[]\_get" ( `x = registry[key]` uses the `index` method) and set "operator_[]_set" (`registry[key] = val` uses the `index_new` method).

The `__metatable` method is a way to prevent users from changing the `registry` to become an actual table using a new metatable.

Metatables also let you introduce OOD patterns and things like polymorphism - you can use the metatable to say things like "every time an access is made to a table, try to access the field on this other table first" - a la superclasses.

(This is similar to Javascript objects having an `__proto__` field, except it's designed in a way that you cannot access a table's metatable from the table itself).

The registry may be the only structure with a metatable, if you need other Rust data you'll probably use a function.

src/lua/mod.rs
SnirkImmington: I might want to bring it back one day?

src/lua/tests.rs
SnirkImmington: Well what happens is when you send you get a "Lua thread not initialized". The `sender` channel isn't initialized until the thread is up and running, and the send will also fail if `lua::is_running()` returns false - buffering isn't designed to work while the thread is dead right now (especially because the lua state will be different after a restart).

I think the only use for it would be testing, but we could write a method that busy-waits/sleeps until `lua::is_running()` and then sends.

src/lua/tests.rs
SnirkImmington: It panics if the assert fails. This is testing whether telling the lua thread to terminate (while nothing else is querying the thread) can be done in a timely manner.

src/main.rs
SnirkImmington: Because that would mean adding in the libc dependency (which we do indirectly depend on). It's also no different - the libc definition is `pub type c_char = i8;` - if this type was different from libc's, the log handler and `rustwlc::pointer_to_string` would both have compiler errors.

I know there's no real _cost_ in using libc and fixing this, but given that it's code that we'll never touch again and we have 0 other lines that could use libc I'd rather keep it off of our dependency list. If we ever need any other definition from it, I will gladly add it and replace this.

src/lua/mod.rs
Timidger: Ah, is this not previous code you commented out? I see, if it was I was going to say you could just use version control and revert the commit that removed it, but I see now that you are adding this as a comment. Ok, that's fine then.

src/lua/tests.rs
Timidger: It would be nice to have that as a testing method, if you want we could expose it but it should definitely be as another option since if all requests did that then we would have the threads never die if something goes wrong

src/lua/tests.rs
Timidger: I think you're looking at the wrong test, I'm talking about `test_variable`, whereas the previous test was testing termination, this one is testing if you can bind `hello` to "hello world!"

src/main.rs
Timidger: [It isn't for arm](https://doc.rust-lang.org/libc/arm-unknown-linux-gnueabihf/libc/type.c_char.html). libc should deal with platform-specific issues with dealing with C. If we keep it in its current state and try to compile on ARM (say for a raspberry pi) then we will get a compile error.

src/lua/tests.rs
SnirkImmington: Okay, upgrading the Lua API to smooth out restarts would include some work from `async-lua`.

src/lua/tests.rs
SnirkImmington: Oh, good point, with the `async-lua` changes I should be receiving the response and checking that the response is an okay instead of `LuaError("Assert failed)` or something.

src/registry/tests.rs
SnirkImmington: They'll die on their own after sending, but good point I guess

SnirkImmington: I'll change it to 
```rust
result = result && match rx.recv() { Some(val) => val, Err(_) => false };
```

src/registry/tests.rs
SnirkImmington: One thing I noticed is `BTreeMap` keeps things sorted. If you had two `HashMaps` with an "x" and "y" key, especially if they used different `Hasher`s, you couldn't just iterate through the keys in order and check when you're comparing two Json objects.

PR #7, opened 2016-04-16
Async Lua - upgrades to the Lua thread

SnirkImmington:
Communication with the Lua thread has been revamped. A successful request of the Lua thread returns a `Receiver` which the requester can use to get back the needed data at any time, or `LuaResponse::Ping` for some requests. In addition, `hlua` and `lazy_static` were updated.

### Changes

- `LuaQuery` and `LuaResponse` rewritten
- Update `hlua`, `lazy_static`, `libc`
- `lua::try_send` was rewritten, responses can be asynchronously received
- Most of the `LuaQuery` messages were implemented
- Methods which modify Lua use a `Vec<String>` method for names (as "foo.bar.baz" does not access the two inner tables with `hlua::Lua::get`)
- `convert_json` and its tests were moved to `lua/mod.rs` and its allocation was optimized

### Remaining Work
- [x] ~~Implement `LuaQuery::{SetValue, NewTable}`~~ Use `LuaQuery::ExecWithLua`
- [x] Remove the `try` from functions where it doesn't refer to blocking
- [x] ~~I don't think `LuaQuery::Invoke` _really_ works~~
- [ ] Tests for all the `LuaQuery`s


SnirkImmington:
Lua requests should start getting merged together into ones like `LuaQuery::ExecWithLua`.

Also, if the lua thread handlers get any bigger they should be moved into their own partially-reexported module.

src/lua/mod.rs
```diff
@@ -218,61 +202,173 @@ fn thread_handle_message(sender: &Sender<LuaResponse>,
...
-            match var_result {
-                Some(var) => {
-                    thread_send(&sender, LuaResponse::Variable(Some(var)));
+            if varname.len() == 0 {
+                thread_send(request.reply, LuaResponse::InvalidName);
+                return;
+            }
+            // Table[0] String had to be cloned, it'd be nice if Rust let us
+            // borrow out parts of memory
+            match lua.get::<AnyLuaValue, _>(varname[0].clone()) {
```
Timidger: If you want to borrow out the part of the memory, why can't you used `varname[0].borrow()`?

src/lua/mod.rs
```diff
@@ -218,61 +202,173 @@ fn thread_handle_message(sender: &Sender<LuaResponse>,
...
+            // Non-string keys are not allowed
+            _ => {
+                return Err(());
+            }
         }
     }
+
+    // Gauss' trick
+    let desired_sum = ((arr.len()) * (arr.len() + 1)) / 2;
+    if counter != desired_sum as f64 {
```
Timidger: This is clever, but what's stopping you from your counter being a `u32` and just incrementing by one every time you see a number and just comparing that number to the length of the array?

src/lua/mod.rs
```diff
@@ -218,61 +202,173 @@ fn thread_handle_message(sender: &Sender<LuaResponse>,
...
+                          -> Result<Json, ()> {
+    let mut json_obj: BTreeMap<String, Json> = BTreeMap::new();
+
+    for (key, val) in obj.into_iter() {
+        match key {
+            AnyLuaValue::LuaString(text) => {
+                json_obj.insert(text, try!(lua_to_json(val)));
+            },
+            AnyLuaValue::LuaNumber(ix) => {
+                json_obj.insert(format!("{}", ix), try!(lua_to_json(val)));
```
Timidger: Not using `ix.to_string`? I'm not sure if there is any additional cost to `format!`, but it might be easier to read/understand.

src/lua/types.rs
```diff
@@ -0,0 +1,154 @@
...
+    Error(hlua::LuaError),
+    /// A function is returned
+    Function(hlua::functions_read::LuaFunction<String>),
+    /// Pong response from lua ping
+    Pong,
+}
+
+impl LuaResponse {
+    /// Whether this response is an InvalidName or Error
+    pub fn is_err(&self) -> bool {
```
Timidger: So you're using `Result` semantics for LuaResponse to test if the response was a valid type? Why can't we just have LuaResponse's wrapped in Results<> and not have to add a layer an InvalidName? Or did you want to do to that but keep this and have this "error" spell out that the other side sent something wrong and the other result type means a different error has a occured?

src/registry/types.rs
```diff
@@ -4,38 +4,43 @@ use std::sync::Arc;
...
-pub enum RegistryAccess {
-    Public,
-    Lua,
-    Private
+bitflags! {
+    /// Access permissions for items in the registry
+    pub flags AccessFlags: u8 {
+        const LUA_PRIVATE = 0b00000000,
+        const LUA_ACCESS  = 0b00000001,
+        const LUA_WRITE  = 0b00000010,
```
Timidger: Why don't you just write this as hex? `0x0`, `0x1`, `0x2`, etc.?

src/lua/mod.rs
SnirkImmington: If any of the indicies are missed, an Object is returned instead of an Array. Perhaps a better solution is to index those with a Null.

src/registry/types.rs
SnirkImmington: They're flags (the convention is `1 << 0,  1 << 1, 1 << 2`) but I like how the binary representation shows where the ones are. It would look better aligned though.

Timidger: I think it would probably be easier to read if we don't use the full binary format. Those ones all string together. Either the `1 << 0` or the hex don't have that problem.

Timidger:
There seems to be conflicts? Is it from the age of the pull request?

SnirkImmington:
Conflicts are because `logging-hotfix` was merged in first.

PR #8, opened 2016-04-20
Logging hotfix

Timidger:
Updated logging so that it uses a string. LibC dependency removed. The build will fail because I tested against the changes done in `rust-wlc`. Once those changes are merged in and the crate updated, this should be merged in.


Timidger:
My bad, wrong branch

PR #9, opened 2016-04-20
Updated to new logging (that doesn't use a char*)

Timidger:
Updated logging so that it uses a string. LibC dependency removed. The build will fail because I tested against the changes done in rust-wlc. Once those changes are merged in and the crate updated, this should be merged in.


SnirkImmington:
I've updated the crate, and it appears at https://crates.io/crates/rustwlc/0.3.2, but https://github.com/rust-lang/crates.io-index isn't updating.

PR #11, opened 2016-04-29
Lua registry - utilities to convert to AnyLuaValue, registry values are AnyLuaValue

SnirkImmington:
#### Project:
- Remove dependency `rustc_serialize`

#### Convert:
- Add `convert` module with `ToTable`, `FromTable`, `LuaDecoder`
- `ToTable` and `FromTable` traits can convert an `AnyLuaValue` to and from `T` without copies
- `LuaDecoder` is able to convert most primitive types, unsorted `Vec<T>`, and named struct fields recursively

#### Macros:
- Add `macros.rs` which provides `lua_convertible!` for creating a struct that has `ToTable` and `FromTable` without using nightly features

#### Registry:
-  Registry uses `AnyLuaValue` types


src/convert/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+//! Also see macros.
+
+pub mod serialize;
+pub mod deserialize;
+
+pub use self::serialize::ToTable;
+pub use self::deserialize::
+{FromTable, LuaDecoder, ConverterError, ConvertResult};
+
+// Tests for serialize <-> deserialize compatability
```
Timidger: If this is tests explicitly for this, shouldn't it be in deseralize.rs? Why is it in mod.rs?

src/convert/serialize.rs
```diff
@@ -0,0 +1,72 @@
...
+
+numeric_impl!(usize, isize);
+numeric_impl!(i8, i16, i32);
+numeric_impl!(u8, u16, u32);
+numeric_impl!(f32, f64);
+
+// Implementation for &Ts which use Copy syntax
+impl<'a, T> ToTable for &'a T where T: Copy + ToTable {
+    fn to_table(self) -> AnyLuaValue {
+        self.clone().to_table()
```
Timidger: Why are you `clone`ing here when `T` implements copy?

src/lua/funcs/lua_registry.rs
```diff
@@ -1,21 +1,21 @@
...
 use std::ops::Deref;
 
 pub fn index(table: AnyLuaValue, lua_key: AnyLuaValue) -> AnyLuaValue {
-    if let AnyLuaValue::LuaString(key) = lua_key {
-        let maybe_json = registry::get_json(&key);
-        if let Some(json_pair) = maybe_json {
-            let (access, json) = json_pair;
+    if let LuaString(key) = lua_key {
+        let maybe_lua = registry::get_lua(&key);
+        if let Some(lua_pair) = maybe_lua {
```
Timidger: Since you only use `maybe_lua` here, maybe it would be cleaner to just do `if let Some(lua_pair) = registry::get_lua(&key) {`. Less mental overhead when you don't have to remember another variable.

src/lua/funcs/lua_registry.rs
```diff
@@ -25,13 +25,11 @@ pub fn index(table: AnyLuaValue, lua_key: AnyLuaValue) -> AnyLuaValue {
...
-    if let AnyLuaValue::LuaString(key) = lua_key {
-        if let Ok(json_val) = lua_to_json(val) {
-            let ref mut reg = registry::write_lock();
-            if let Some(mut reg_val) = (*reg).get_mut(&key) {
-                reg_val.set_json(json_val);
-                return Ok(());
-            }
+    if let LuaString(key) = lua_key {
+        let ref mut reg = registry::write_lock();
+        if let Some(mut reg_val) = (*reg).get_mut(&key) {
```
Timidger: unnecessary dereferencing on `reg`, and once again it might be clearer to do away with that variable altogether?

src/lua/funcs/lua_registry.rs
```diff
@@ -25,13 +25,11 @@ pub fn index(table: AnyLuaValue, lua_key: AnyLuaValue) -> AnyLuaValue {
...
                  -> Result<(), &'static str> {
-    if let AnyLuaValue::LuaString(key) = lua_key {
-        if let Ok(json_val) = lua_to_json(val) {
-            let ref mut reg = registry::write_lock();
-            if let Some(mut reg_val) = (*reg).get_mut(&key) {
-                reg_val.set_json(json_val);
-                return Ok(());
-            }
+    if let LuaString(key) = lua_key {
+        let ref mut reg = registry::write_lock();
```
Timidger: Why are you taking this as a `ref`?

src/macros.rs
```diff
@@ -0,0 +1,72 @@
+//! Warning: extreme macros
+
+// TODO visibility
+// Usage:
```
Timidger: Are these supposed to be doc comments?

src/macros.rs
```diff
@@ -0,0 +1,72 @@
...
+//
+// struct Bar {
+//     name: String,
+//     point: Point
+// }
+//
+// }
+macro_rules! lua_convertible {
+    (  $(#[$attr:meta])*
+       struct $name:ident { $($fname:ident : $ftype:ty, )+  }  ) => {
```
Timidger: The indentation here is a little confusing, at first I though there were multiple arms, but brace highlighting tells me there is just one. Perhaps you should indent after this line to better illustrate that?

src/registry/mod.rs
```diff
@@ -45,30 +45,25 @@ pub fn write_lock<'a>() -> RwLockWriteGuard<'a, RegMap> {
...
 pub fn get<K, T>(name: &K) -> Result<(AccessFlags, T), RegistryError>
-where T: Decodable, String: Borrow<K>, K: Hash + Eq + Display {
-    let maybe_json = get_json(name);
-    if let Some(json_pair) = maybe_json {
-        let (access, json_arc) = json_pair;
-        let mut decoder = json::Decoder::new(json_arc.deref().to_json());
-        match T::decode(&mut decoder) {
+where T: FromTable, String: Borrow<K>, K: Hash + Eq + Display {
+    let maybe_lua = get_lua(name);
+    if let Some(lua_pair) = maybe_lua {
```
Timidger: Unnecessary `maybe_lua`, perhaps `if let Some(lua_pair) = get_lua(name)` is better?

src/registry/mod.rs
```diff
@@ -77,10 +72,10 @@ where T: Decodable, String: Borrow<K>, K: Hash + Eq + Display {
 }
 
 /// Set a key in the registry to a particular value
-pub fn set<T: ToJson>(key: String, flags: AccessFlags, val: T) {
-    trace!("set: {}", key);
-    let ref mut write_reg = *write_lock();
+pub fn set<T: ToTable>(key: String, flags: AccessFlags, val: T) {
+    trace!("set: {:?} {}", flags, key);
     let regvalue = RegistryValue::new(flags, val);
+    let ref mut write_reg = *write_lock();
```
Timidger: Why are you taking the reference of a dereferenced value? Just do:
`let mut write_reg = write_lock();`

src/macros.rs
```diff
@@ -0,0 +1,72 @@
+//! Warning: extreme macros
```
Timidger: Also, perhaps we shouldn't dump our macros in a file called `macros.rs`? It'd be better to logically put them in a place that makes sense, in this case for the lua conversion.

src/registry/mod.rs
```diff
@@ -45,30 +45,25 @@ pub fn write_lock<'a>() -> RwLockWriteGuard<'a, RegMap> {
...
 }
 
-/// Gets a Json object from a registry key
-pub fn get_json<K>(name: &K) -> Option<(AccessFlags, Arc<Json>)>
+/// Gets a Lua object from a registry key
+pub fn get_lua<K>(name: &K) -> Option<(AccessFlags, Arc<AnyLuaValue>)>
 where String: Borrow<K>, K: Hash + Eq + Display {
-    trace!("get_json: {}", *name);
+    trace!("get_lua: {}", *name);
     let ref reg = *read_lock();
```
Timidger: Taking ref of a dereferenced value, just use `let reg = read_lock();`

PR #12, opened 2016-05-06
Added workspaces feature, basics of tree for tiling layout

Timidger:


src/compositor.rs
```diff
@@ -0,0 +1,65 @@
+//! Compositor: wm state including the layout, helm, and interactions
```
Timidger: Fix this so it would compile if we were to use it, leave it in and expand later

src/layout/tree.rs
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+    }
+    None
+
+}
+
+fn get_current_workspace<'a>(tree: &'a Tree) -> Option<&'a mut Node> {
+    if let Some(container) = tree.get_active_container() {
+        //if let Some(child) = container.get_ancestor_of_type(ContainerType::Workspace) {
```
SnirkImmington: Does this work properly?

Timidger: The commented out code _does_ work **if** get_active_container actually did what it did, right now I've hacked it so it just returns the active workspace, so that commented out code would never work (since we that code is trying to find that workspace). I should have fixed that up so it's easier to see that, sorry. I can get to that later tonight, though feel free to rename things to make more sense if you'd like/can

SnirkImmington:
We need to add more tests to the tree itself, and I assume the tree API is going to change a bit (i.e. with some of the borrow root methods). However, I think we should add the tests in the big cleaning branch and work on the API when we implement containers.

PR #22, opened 2016-05-20
Registry properties

SnirkImmington:
Properties were added to the registry! This is similar to programming language properties, and is implemented by registering Rust methods to be called when a registry key is read or written.

- Add `Property` variant to registry values
- Rename `RegistryValue` -> `RegistryField`
- Unify registry return types as `RegistryResult`
- Rename flags to `AccessFlags::{READ, WRITE}`
- Reduce `trace!` spam from using the registry (but add printing of values)
- Rename `registry::{get, set}_data` to `registry::{get, set}_struct`
- Split up registry tests

### DIFF
--- a/src/registry/mod.rs
+++ b/src/registry/mod.rs
@@ -17,7 +17,7 @@ pub use self::types::*; // Export constants too
 #[cfg(test)]
 mod tests;
 
-pub type RegMap = HashMap<String, RegistryValue>;
+pub type RegMap = HashMap<String, RegistryField>;
 
 lazy_static! {
     /// Registry variable for the registry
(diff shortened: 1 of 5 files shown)


SnirkImmington:
Latest comment fixes warnings, clear to merge in.

src/registry/mod.rs
```diff
@@ -49,77 +56,169 @@ pub fn write_lock<'a>() -> RwLockWriteGuard<'a, RegMap> {
...
+                        None => return Err(RegistryError::InvalidOperation)
+                    }
+                }
+                else {
+                    return Err(RegistryError::WrongKeyType);
+                }
+            }
+        }
+    }
+    func(json);
```
Timidger: Why do you set the func on L165 and then call it down here, instead of simply calling func on L165 and returning `Ok(None)`?

PR #24, opened 2016-06-06
Finalization of Layout Tree

Timidger:
The internals of the tree have been moved over from the very error-prone and very buggy unsafe implementation that utilized raw pointers to using petgraph, a 3rd party library. The implementation essentially uses an adjacency list that only invalidates the last index when a node is removed. 

* Added workspace switching
* Added the ability to send containers to different workspaces
* Consolidated the active containers
* Severely increased the stability of the layout (shouldn't crash anymore, hopefully)


Timidger:
Tried to merge to master, my bad

Issue #19, opened 2016-05-17
Tiling features


SnirkImmington:
These features are desired of the way-cooler tiling algorithm (see [way-cooler/tree/tiling](https://github.com/Immington-Industries/way-cooler/tree/tiling)):

### Workspaces
- [x] Can have names or use numbers
- [ ] Can be generated on the fly or exist as a static list
- [x] Each monitor can have its own workspace list, 
- [ ] or the list is global across all outputs 
- [x] Can move containers between workspaces

### Containers
- [x] Views are assigned to containers
- [x] Containers can tile horizontally or vertically
- [x] Containers can be nested
- [x] Containers can be controlled i3 style
- [ ] Can move containers to other containers
- [ ] Custom tiling patterns so container keybindings aren't needed

### Views
- [x] Try not to tile views that don't want to be tiled
- [ ] Floating views

Timidger:
A lot of this stuff is still up in the air, but I'm going to go ahead and close it since we have the majority of the features and they should exist as separate issues now.

PR #29, opened 2016-06-21
Tiling Support

Timidger:
This pull request adds the following features:
* Windows are auto-tiled
    - Can tile windows vertically or horizontally
    - Allows you to nest containers, allowing infinite flexibility in how you want to structure your windows
    - Modelled off i3's tiling, see [here](https://i3wm.org/docs/userguide.html#_tree) for more details
* Allows you to switch the focus of containers without the mouse
* Removed support to move and resize individual windows
* Adds the following commands (with default key bindings):
    - `horizontal_vertical_switch`, (alt+e): Switch the current container between horizontal and vertical layouts. If the current container is a view, this switches the parent's layout.
    - `split_vertical`, (alt+v): Takes the current container and adds it as the only child to a new vertically tiled container at the place where the active container was at originally. This is the primary way to make a new sub container with a vertical layout.
    - `split_horizontal`, (alt+h): Takes the current container and adds it as the only child to a new horizontally tiled container at the place where the active container was at originally. This is the primary way to make a new sub container with a horizontal layout.
    - `focus_{up,down,left,right}`, (alt+\<arrow key\>): Switch focus to the window to the {up,down,left,right} of the active container.
    - `move_to_workspace_{index}`, (alt+\<number\>): Move the active container to a different workspace.
    - `remove_active`, (alt+q): Remove the active container (the one currently focused).

This crosses out some features in #19

### DIFF
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -80,7 +80,7 @@ pub extern fn view_destroyed(view: WlcView) {
     trace!("view_destroyed: {:?}", view);
     if let Ok(mut tree) = tree::try_lock_tree() {
         tree.remove_view(&view);
-        tree.update_active_of(ContainerType::Container);
+        tree.update_active_of(ContainerType::Workspace);
     } else {
         warn!("Could not delete view {:?}", view);
     }
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -62,11 +62,18 @@ pub extern fn output_render_post(output: WlcOutput) {
 pub extern fn view_created(view: WlcView) -> bool {
     trace!("view_created: {:?}: \"{}\"", view, view.get_title());
     let output = view.get_output();
-    view.set_mask(output.get_mask());
     if let Ok(mut tree) = tree::try_lock_tree() {
+        if tree.get_active_container().is_none() {
+            warn!("Could not create view, there is no focus \
+                    so Way-Cooler doesn't know where to put it");
+            return false;
+        }
+        view.set_mask(output.get_mask());
         let v_type = view.get_type();
         if v_type != ViewType::empty() {
             view.focus();
+            // Now focused on something outside the tree, have to unset the active container
+            tree.unset_active_container();
             return true
         }
         tree.add_view(view.clone());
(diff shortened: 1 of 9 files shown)


Timidger:
I'll add tests, though the code shouldn't change. Look over it when you get a chance.

src/callbacks.rs
```diff
@@ -42,7 +43,12 @@ pub extern fn output_focus(output: WlcOutput, focused: bool) {
 pub extern fn output_resolution(output: WlcOutput,
                             old_size_ptr: &Size, new_size_ptr: &Size) {
     trace!("output_resolution: {:?} from  {:?} to {:?}",
-             output, *old_size_ptr, *new_size_ptr);
+           output, *old_size_ptr, *new_size_ptr);
+    // Update the resolution of the output and it's children
```
SnirkImmington: its

src/callbacks.rs
```diff
@@ -57,11 +63,28 @@ pub extern fn view_created(view: WlcView) -> bool {
     trace!("view_created: {:?}: \"{}\"", view, view.get_title());
     let output = view.get_output();
     if let Ok(mut tree) = tree::try_lock_tree() {
-        tree.add_view(view.clone());
-        drop(tree);
+        if tree.get_active_container().is_none() {
+            warn!("Could not create view, there is no focus \
+                    so Way-Cooler doesn't know where to put it");
```
SnirkImmington: Very trivial complaint, but do we want way-cooler spelled like this to show up in logs?

Timidger: It can be re-worded so it doesn't have to mention its own existence

src/callbacks.rs
```diff
@@ -57,11 +63,28 @@ pub extern fn view_created(view: WlcView) -> bool {
...
+            return false;
+        }
         view.set_mask(output.get_mask());
-        view.bring_to_front();
-        view.focus();
+        let v_type = view.get_type();
+        if v_type != ViewType::empty() {
+            view.focus();
+            // Now focused on something outside the tree, have to unset the active container
+            if tree.get_active_container().is_some() {
```
SnirkImmington: Can these two ifs just be
```
if !tree.active_is_root() {
    tree.unset_active_container();
}
return true
```

src/callbacks.rs
Timidger: thanks, for things like that that don't really matter, feel free to just change that yourself

src/callbacks.rs
Timidger: I believe you meant the next two lines. In any case, yes that is correct

SnirkImmington:
I'll not subject you to finicky corrections.

SnirkImmington:
Code seems great to me, I'll merge in after the tests.

PR #31, opened 2016-07-02
Way-Cooler no longer segfaults when closing with multiple windows

Timidger:
When multiple windows were open on Way-Cooler while it was being closed (either via a SIGINT signal or via the terminate command), it would mysteriously segfault. The issue was traced to the tree attempting to do operations on the views that were now invalidated by wlc. To remedy this, in the `compositor_terminating` callback the tree will now properly invalidate itself without attempting to close any views.


SnirkImmington:
:+1:

PR #32, opened 2016-07-02
Added tree as json property to registry

Timidger:
* Implemented `ToJson` for `LayoutTree`
* Added the key `tree_layout` to registry that gets the tree's structure as a JSON object


SnirkImmington:
:+1:

PR #39, opened 2016-07-25
Configure via init.lua

SnirkImmington:
Way Cooler reads the file `init.lua` from one of several sources (`$WAY_COOLER_CONFIG`, `$XDG_CONFIG_DIR/way-cooler/init.lua`, `$HOME/.config/way-cooler`, `/etc/way-cooler/init.lua`). The init file can register keybindings and access the Way Cooler API directly.

**Not implemented**
- Graceful handling of not having a init file
- Useful errors/reporting when there's something wrong with the init file
- Loading without a init file/using a backup file
- Reloading the init file (Sending Lua the restart message might work)
- Separation of Lua startup & reading init file (there's an `if !cfg!(test)` around reading init file though)
- Testing of the Lua/Rust code surrounding the init file


Timidger:
Have to re-add `alt+esc` command hard coded into the rust code

config/init.lua
```diff
@@ -0,0 +1,103 @@
...
+  -- (if you're used to mod+arrow to switch workspaces you may wanna change them)
+  key({ mod }, "left", "layout.focus_left"),
+  key({ mod }, "right", "layout.focus_right"),
+  key({ mod }, "up", "layout.focus_up"),
+  key({ mod }, "down", "layout.focus_down"),
+
+  -- Split containers
+  key({ mod }, "h", "layout.split_horizontal"),
+  key({ mod }, "v", "layout.split_vertical"),
+  key({ mod, "Shift" }, "e", "layout.switch_horizontal_vertical"),
```
Timidger: Why'd you add a "shift" modifier to this command?

config/init.lua
```diff
@@ -0,0 +1,103 @@
...
+
+
+  -- Quit
+  key({ mod, "Shift" }, "q", "quit"),
+}
+
+-- Add Mod + X bindings to switch to workspace X, Mod+Shift+X send active to X
+for i = 1, 9 do
+  table.insert(keys,
+               key({ mod          }, tostring(i), "switch_workspace_" .. i))
```
Timidger: There's some crazy amount of white space going on in here

lib/lua/lua_init.lua
```diff
@@ -0,0 +1,69 @@
...
+    }
+end
+
+local use_key = ", use the `key` or `config.key` method to create a keybinding"
+
+-- Converts a list of modifiers to a string
+local function keymods_to_string(mods, key)
+    table.insert(mods, key)
+    local turn = table.concat(mods, ',')
+    print(turn)
```
Timidger: You probably should clear out your debug print statements, even if you don't use the function yet.

src/commands/defaults.rs
```diff
@@ -143,11 +144,22 @@ fn focus_down() {
 }
 
 fn launch_terminal() {
-    let term = env::var("WAYLAND_TERMINAL")
-        .unwrap_or("weston-terminal".to_string());
+    warn!("Got {:?}", registry::get_data("terminal"));
```
Timidger: Remove debug print statement

src/commands/defaults.rs
```diff
@@ -143,11 +144,22 @@ fn focus_down() {
...
+    let command: String;
+    if let Ok((flags, data)) = registry::get_data("terminal").map(|d| d.resolve()) {
```
Timidger: This could be simplified, I suggest something like the following:

```
let command = registry::get_data("terminal").map(|d| d.resolve()).and_then(|(_, data)| {
        data.as_string().map(|str_ref| str_ref.to_string())
            .ok_or(RegistryError::KeyNotFound)
    }).unwrap_or("weston_terminal".to_string());
```

src/keys/keypress.rs
```diff
@@ -0,0 +1,85 @@
...
+use rustwlc::KeyMod;
+use rustwlc::xkb::{Keysym, NameFlags};
+use std::hash::{Hash, Hasher, SipHasher};
+
+use std::fmt::Result as FmtResult;
+use std::fmt::{Display, Formatter};
+
+/// Structure containing keys which are pressed
+/// to trigger a keybinding.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
Timidger: Should this be Copy?

src/keys/keypress.rs
```diff
@@ -0,0 +1,85 @@
...
+    modifiers: KeyMod,
+    /// The keys pressed
+    key: Keysym
+}
+
+impl KeyPress {
+    /// Creates a new KeyPress struct from a list of modifier and key names
+    pub fn from_key_names(mods: &[&str], key: &str) -> Result<KeyPress, String> {
+        super::keymod_from_names(mods).and_then(|mods| {
+            let proper_key = if let Some(found_key) = super::NAME_MAPPING.get(key) {
```
Timidger: `unwrap_or`

src/keys/keypress.rs
```diff
@@ -0,0 +1,85 @@
...
+                found_key
+            }
+            else {
+                key
+            };
+
+            // Parse a keysym for each given key
+            let key_sym = try!(
```
Timidger: I suggest

```
            let key_sym = try!(
                Keysym::from_name(key.to_string(), NameFlags::None)
                    .or_else(|| Keysym::from_name(key.to_string(), NameFlags::CaseInsensitive))
                    .ok_or(format!("Invalid key: {}", key))
            );
```

src/keys/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+use super::commands::{self, CommandFn};
+
+mod keypress;
+pub use self::keypress::KeyPress;
+
+mod event;
+pub use self::event::KeyEvent;
+
+lazy_static! {
+    static ref BINDINGS: RwLock<HashMap<KeyPress, KeyEvent>> =
```
Timidger: This isn't used anywhere else, will it be implemented later or is it no longer necessary?

src/lua/init_path.rs
```diff
@@ -0,0 +1,51 @@
...
+pub const INIT_FILE_FALLBACK_PATH: &'static str = "/etc/way-cooler/";
+
+#[inline]
+fn read_file<P: AsRef<Path>>(path: P) -> IOResult<File> {
+    OpenOptions::new().read(true).open(path)
+}
+
+/// Parses environment variables
+pub fn get_config() -> Result<File, &'static str> {
+    if let Ok(path_env) = env::var("WAY_COOLER_CONFIG") {
```
Timidger: It would perhaps be cleaner to implement this with a for loop iterating over a tuple or hard-coded array of paths. It would *maybe* be slightly less efficient by reading those variables and paths at the start, but it's doubtful to make any kind of difference. And you can improve that by making it iterate over a lazy iterator and it'd probably compile down to similarly efficient code.

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
+//! Rust code which is called from lua in the init file
+#![deny(unused_code)]
```
Timidger: I think you meant `#![deny(dead_code)]`

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+use commands;
+use keys::{self, KeyPress, KeyEvent};
+use convert::json::{json_to_lua, lua_to_json};
+use convert::serialize::ToTable;
+
+type OkayResult = Result<(), &'static str>;
+type ValueResult = Result<AnyLuaValue, &'static str>;
+
+/// We've `include!`d the code which initializes from the Lua side.
+const LUA_INIT_CODE: &'static str = include_str!("../../lib/lua/lua_init.lua");
```
Timidger: Though it would probably be optimized out, should this really be a `const`? It'll only be used in one place in this entire module and it's not even exported.

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+use hlua::any::AnyLuaValue::*;
+use rustc_serialize::json::Json;
+
+use registry::{self, RegistryField, RegistryError, AccessFlags};
+use commands;
+use keys::{self, KeyPress, KeyEvent};
+use convert::json::{json_to_lua, lua_to_json};
+use convert::serialize::ToTable;
+
+type OkayResult = Result<(), &'static str>;
```
Timidger: You're inconsistent with your usage of this Result, and it's generally not really worth it. It might save on a few keypresses, but reading the code it requires a lookup for a very simple type.

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+type ValueResult = Result<AnyLuaValue, &'static str>;
```
Timidger: This is also inconsistently used, but has a better use case. I suggest either getting rid of it or using it everywhere.

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+/// Run a command
+fn ipc_run(command: String) -> Result<(), &'static str> {
+    trace!("Called ipc_run with {}", command);
+    match commands::get(&command) {
+        Some(com) => {
+            com();
+            Ok(())
+        },
+        None => Err("Command does not exist")
+    }
```
Timidger: `commands::get(&command).map(|com| com())
    .ok_or("Command does not exist")

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+}
+
+/// ipc 'set' handler
+fn ipc_set(key: String, value: AnyLuaValue) -> Result<(), &'static str> {
+    trace!("Called ipc_set with {}", key);
+    let json = try!(lua_to_json(value).map_err(
+        |_| "Unable to convert value to JSON!"));
+    match registry::set_json(key.clone(), json.clone()) {
```
Timidger: Consider:

```
registry::set_json(key.clone(), json.clone())
    .map(|data| data.call(json.clone()))
    .or_else(|err| {
        match err {
            RegistryError::InvalidOperation => {
                Err("That value can not be set!")
            },
            RegistryError::KeyNotFound => {
                registry::insert_json(key, AccessFlags::READ(), json.clone());
                Ok(())
            }
        }
    })
```

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+fn init_workspaces(options: AnyLuaValue) -> OkayResult {
+    error!("Attempting to call `init_workspaces`, this is not implemented");
+    Ok(())
+}
+
+/// Registers a command keybinding.
+fn register_command_key(mods: String, command: String, _repeat: bool) -> Result<(), String> {
+    trace!("Registering command key: {} => {}", mods, command);
+    if let Ok(press) = keypress_from_string(&mods) {
+        if let Some(command) = commands::get(&command) {
```
Timidger:
```
    let command = try!(commands::get(&command)
                       .ok_or(format!("Command '{}' for keybinding '{}' not found", command, press)));
    keys::register(press, KeyEvent::Command(command));
```

src/lua/rust_interop.rs
```diff
@@ -0,0 +1,174 @@
...
+    else {
+        Err(format!("Invalid keypress {}, {}", mods, command))
+    }
+}
+
+/// Rust half of registering a Lua key: store the KeyPress in the keys table
+/// and send Lua back the index for __key_map.
+fn register_lua_key(mods: String, repeat: bool) -> Result<String, String> {
+    trace!("Registering lua key: {}, {}", mods, repeat);
+    if let Ok(press) = keypress_from_string(&mods) {
```
Timidger:
```    
keypress_from_string(&mods)
    .map(|press| {
        keys::register(press.clone(), KeyEvent::Lua);
        press.get_lua_index_string()
    }).map_err(|_| format!("Invalid keys '{}'", mods))
```

src/lua/thread.rs
```diff
@@ -98,15 +99,27 @@ pub fn init() {
...
-    lua.execute_from_reader::<(), File>(
-        File::open("lib/lua/init.lua")
-            .expect("Lua thread unable to find init file")
-    ).expect("Lua thread: unable to execute init file");
     trace!("Loading way-cooler libraries...");
-    funcs::register_libraries(&mut lua);
+    rust_interop::register_libraries(&mut lua);
+
+    if !cfg!(test) {
+        if let Ok(init_file) = init_path::get_config() {
```
Timidger: Wouldn't a better solution be a 
```
#[cfg!(test)]
fn get_config() {
    <insert sane default here, be it a blank one or something to test>
}

src/keys/keypress.rs
SnirkImmington: Can't derive `Copy` because `KeySym` doesn't have it. I'll switch us off of `KeySym` later.

PR #43, opened 2016-08-07
Added better tree error info propogated up

Timidger:
Instead of just returning a simple error string, commands to the Tree will not return an error that has information about what actually went wrong so it can be dealt with at a higher level.

### DIFF
--- a/src/layout/layout_tree/tree.rs
+++ b/src/layout/layout_tree/tree.rs
@@ -3,6 +3,8 @@
 
 use petgraph::graph::NodeIndex;
 
+use uuid::Uuid;
+
 use rustwlc::{WlcView, WlcOutput, Geometry, Point};
 
 use super::super::LayoutTree;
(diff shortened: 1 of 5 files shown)


src/layout/layout_tree/tree.rs
```diff
@@ -17,6 +19,21 @@ pub enum Direction {
     Left
 }
 
+#[derive(Clone, Debug)]
+pub enum TreeError {
+    /// A Node can nt be found in the tree with this Node Handle.
```
SnirkImmington: You're always telling me to spellcheck so I figured I'd point this out.

Timidger: Yes, thank you I will be sure to fix this. 

Part of the big deal I make however is that there are sometimes spell checking issues on your PRs that I don't find but rather my computer finds (e.g: the little underline telling me there's a misspelling). We all make spelling mistakes, and part of this PR process should be to mitigate that, but I was concerned after that happened a few times that you didn't have spell checking enabled. 

This is one of the those cases where the computer isn't smart enough to tell me I made a mistake (at least on my editor), and I appreciate this process.

Timidger:
Is this good to merge in?

PR #45, opened 2016-08-09
Announcement cleanup

Timidger:
Cleanup of a few things, some code, some user facing stuff.

* Configuration file environment variable now should point to the variable, not the path that has the file.
* Pruned the travis build script.
* Better logging / logging prune.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,81 +1,95 @@
-# way-cooler
+# Way Cooler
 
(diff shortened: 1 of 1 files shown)


Timidger:
There's nothing else I need to clean up, is there anything else you wanted to do @SnirkImmington?

PR #74, opened 2016-08-16
Made it more clear we link a video

Timidger:


SnirkImmington:
You took a screenshot halfway through the video. Maybe it's a better frame than the one we had before but it looks hideous.

Timidger:
Yeah, that was a concern I had about it. I'll can attempt to edit it later so we can keep the frame but make it look like it's at the beginning of the video. Should I do that?

SnirkImmington:
You could remove the alt image altogether, give me a lovely "Firefox has prevented the outdated plugin 'Adobe Flash' from running" box in the readme instead 😛

Timidger:
Though to be honest I don't think my change looks that hideous, I think it makes it clearer that it's a video. If it was literally an image that would be hideous but I think this is an ok compromise.

SnirkImmington:
Okay, I suppose it's not _that_ hideous.

PR #88, opened 2016-08-18
Add section on how we want our PRs to look

Timidger:


SnirkImmington:
I think we need to refine Contributing.md - remove rambling and narrow down to hard code checks (i.e. clippy, guidelines, etc.) and proper PR format. It should ideally fit into a couple of screenfulls or so.

Timidger:
I think the best way to do that would be to have a separate file for "ways to contribute" and just link it from there.

SnirkImmington:
Feel free to add that to this PR.

PR #93, opened 2016-08-23
Added ability to move containers (Fixes #71)

### LINKED ISSUE #71: Add ability to move views between containers
Right now the only modification you can do to the placement of views is move them between workspaces. This is not entirely helpful for when you want to move them around with their other siblings or move them into sub containers.

There are some edge cases that need to be considered, but most of them should be pretty easy. We are not looking for complete feature parity with i3, but just what makes sense in this context. If it would make more sense to move it into a separate keybinding, it can be changed. For now though it will be very similar, if not identical, to how i3 works.

This will not add support for moving containers, as that requires having some visual way to view containers (#72) and some way to select parent/child containers as the active container (#73)

Timidger:
* A new keybinding to move the active view: `Alt+Shift+<Arrow-Key>` to move in a desired direction.
* Split up the code a bit more, all of the movement code is in its own module.
* Tiny cleanups here and there where we can have better error propagation, more tracing, or easier to read code.

### DIFF
--- a/src/layout/container.rs
+++ b/src/layout/container.rs
@@ -203,7 +203,7 @@ impl Container {
     /// Determines if the container is focused or not
     pub fn is_focused(&self) -> bool {
         match *self {
-            Container::Root { .. } => false,
+            Container::Root(_)  => false,
             Container::Output { ref focused, .. } |
             Container::Workspace { ref focused, .. } |
             Container::Container { ref focused, .. } |
@@ -214,7 +214,7 @@ impl Container {
     /// Sets the focused value of the container
     pub fn set_focused(&mut self, new_focused: bool) {
         match *self {
-            Container::Root { .. } => {},
+            Container::Root(_) => {},
             Container::Output { ref mut focused, .. } |
             Container::Workspace { ref mut focused, .. } |
             Container::Container { ref mut focused, .. } |
--- a/src/layout/container.rs
+++ b/src/layout/container.rs
@@ -62,7 +62,7 @@ pub enum Layout {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Container {
     /// Root node of the container
-    Root,
+    Root(Uuid),
     /// Output
     Output {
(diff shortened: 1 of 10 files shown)


src/layout/graph_tree.rs
```diff
@@ -12,6 +12,14 @@ use rustwlc::WlcView;
 
 use layout::{Container, ContainerType};
 
+#[derive(Clone, Copy, Debug)]
+pub enum GraphError {
+    /// These nodes were not siblings.
+    NotSiblingsError(NodeIndex, NodeIndex),
```
SnirkImmington: Nit: The other error is just named `NotParent` (which is easier because you're gonna need to write the `GraphError::` anyway)

SnirkImmington:
If you replace all of the `Container::Root { ... }` with `Container::Root(_)`, won't you just have to switch them back if you ever needed to add another field to `Root` (like a path or something)?

Timidger:
Before they were just a label, so I didn't we didn't have the {} syntax before. Also we shouldn't have to add another field, but if we do it's a simple fix

src/layout/graph_tree.rs
Timidger: Good point, I'll change that

src/layout/container.rs
```diff
@@ -199,7 +203,7 @@ impl Container {
     /// Determines if the container is focused or not
     pub fn is_focused(&self) -> bool {
         match *self {
-            Container::Root { .. } => false,
```
SnirkImmington: This diff shows an old `Container::Root { .. }`

Timidger: Ah that was something that slipped by and probably didn't fix for consistency. 

Regardless, I'd rather keep the Root to only be a tuple so as to mark that it shouldn't really be extended.

SnirkImmington: Fair enough.

SnirkImmington:
This one didn't have merge conflicts so I'll merge it first.

PR #95, opened 2016-08-24
Keysym fix - update to latest rustwlc - fixes #57

### LINKED ISSUE #57: Some commands (like window close) command doesn't work
I noticed some commands don't seem to be working now (such as close window, normally alt+q) does this have to do with changes made in xkb @SnirkImmington? This also changes depending on whether running with `cargo run` or `cargo build`. While debugging #52 I came across this issue, so it may have to do with that as well.

SnirkImmington:
Finally getting this fix out. We had improper bindings to wlc before, they are now fixed. Also #57 will be fixed.

I never got around to updating dummy-rustwlc, and I'd really rather write a `syntex_syntax` thing to do it for us (i.e. rebuild in Travis every time). So all of our builds will fail after this is merged in until I do that. However, this was an annoying bug, and it shows that we really need systems in place to enable faster deploy times. I'll start looking into having Travis run GitHub releases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = "0.3.1"
 dependencies = [
  "bitflags 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


Timidger:
It looks like there are just three errors, can we just not fix and update a new version of dummy-rustwlc so we don't have a broken build until you can come up with a better solution?

PR #97, opened 2016-08-25
Coveralls hotfix

Timidger:
This fixes coveralls, by building it with a patch that hasn't been merged into the upstream yet. There are problems with testing the coveralls upload from Travis program against nightly using `dylib`, but since that's so niche I thought it prudent to maintain a fork until that is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,10 @@ notifications:
 # load travis-cargo
 before_script:
   - |
-      pip install 'travis-cargo<0.2' --user &&
-      export PATH=$HOME/.local/bin:$PATH
+      git clone git@github.com:Timidger/travis-cargo.git
+      (cd travi-cargo;
+      pip install -e . --user &&
+      export PATH=$HOME/.local/bin:$PATH)
 script:
   - cargo test --verbose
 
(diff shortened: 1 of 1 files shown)


Timidger:
Changes were "unknown" because there's a huge gap in its knowledge, and its failing because its below the threshold we set it to be valid for (50%). I suggest you force merge this in, and we update the tests in other branches before we merge anything else in.

SnirkImmington:
```yaml
# upload the documentation from the build with stable (automatically only actually
# runs on the master branch, not individual PRs)
- travis-cargo --only stable doc-upload
```
Has this even been doing anything this whole time? It's using the script to upload our docs - but where? Our docs don't even change much most of the time, and last I checked they were very sparse because we're building a binary. We need to
* Make sure the docs this is generating are complete enough (i.e. the cargo flags)
* Only run this on tagged releases (we need to set up travis to run github releases anyway)
* Find somewhere to put the docs - I suggest moving our documentation to _project sites_ on GitHub sites - subtly different from what we have now.

Timidger:
I didn't change it at all, looks like we copied it straight from the project home page. It probably wouldn't change anything to remove it. It automatically uploads it to the github pages, doesn't look like we can configure the end-point

SnirkImmington:
Actually, this is what it does:
```
error: uploading docs for package with no library
```

Timidger:
No worries. You probably can't tell what I did because I only merged in the fix, I didn't actually need to write any code, it already exists just hadn't been merged in yet from someone else's fork

PR #99, opened 2016-08-26
Fixed pop-up view closing bug

Timidger:
Sorry for the awful history, pushed to master first so had to do a double revert in order to not destroy the history.


SnirkImmington:
Hey, at least it's not me having git problems.

Issue #73, opened 2016-08-16
Active container should be able to select non-view containers


Timidger:
Right now the active container will only point to view containers, which is sufficient for basic use of the tiling aspects, but we need a way for the user to select multiple containers. While easy to implement in the back end, it is blocked by #72 since we need a way to give visual feedback to the user on which windows are currently selected.

Timidger:
Tiling system is going away and will be done in Lua entirely.

Issue #63, opened 2016-08-16
Can't open new window when there is no active container


Timidger:
To fix: if no active container, use the path to locate the place to put the window (i.e: treat the end of the path as the active container)

Timidger:
Is this bug still present on the latest master (c34f7835f5597927725971d2234810c3fa304f44)?

Timidger:
Is this still an issue future me?

valryne:
There is more which does not work without a focused container:
• launch dmenu
• open terminal (probably what you witnessed)
• all move focus keybindings
• all move active container keybindings
• all split containers keybindings
• mouse click (on window) --left/middel/right-- without moving the mouse prior to click
• mouse click (on window) --middle-- with moving the mouse prior to click
• mouse click (on window title) --middle--  with and without moving the mouse prior to click
• mouse click (on window border "resize pointer") --middle-- with and without moving the mouse prior
• all switch to workspace keybindings (for empty workspaces and current)

What works/gives back focus
• switch to non-empty workspace
• switch to empty workspace (does not actually go there) then switch to current workspace →return focus in current workspace
• click on window title --left/right-- (works even without moving the mouse first)
• click on window border (resize pointer) --left/right-- with and without moving the mouse prior
• click on window without moving --left/right-- if window was active before (like when exiting dmenu)
• click on window after moving the mouse --left/right-- right-click closes window and returns focus to upper right window on screen
• quit way-cooler

If there is at leas one floating window, then there is the following exception to the above:
works: click on window without moving the mouse --left/right-- if window is the floating one

----
I cannot tell, if it all boils down to the same issue, or if some should be listed as separate instances.
----
Tried this with release version 4.0 and current master (synced yesterday) — 'cargo build' with executable in /way-cooler/target/debug and 'cargo install' with executable in ~/.cargo/bin.

Timidger:
You are correct, those all boil down to the same issue. I am looking to address these in 0.4.1, mostly by making there always be some active container. 

If you'd like to test it further (as more eyes are always better) feel free to test it out by using the attached modified PKGBUILD to use the branch that has these fixes. If cargo gives you a weird error in graphmap (like it did with me) install the latest version of cargo and remove it from "optdepends" in the PKGBUILD, there might be a cargo versioning issue I'll need to address later.

valryne:
Works.
With only tiled containers, when I quit dmenu, the last active container becomes active again.
And so far, I could not recreate the situation where there is no active container at all.

[...]

PR #103, opened 2016-09-04
Added Tree paths, Fixes #79

### LINKED ISSUE #79: Add Tree Paths
Right now there are several niceties that can't be implemented because the way Way Cooler remembers the active container is "dumb". It's essentially just an index into the `Node`, and as soon as it is removed or switches it completely forgets it.

If we promote the edge weights to be more than simple `u32`s and instead complex structs that remember which of their children were accessed last we can have a far more sophisticated concept of what `Node` was used last. This will allow Way Cooler to switch between containers and go back the last view that was used in that Container. It will also allow us to many more actions even if we don't have an active container but only a general path to one.

Timidger:
Adds a more sophisticated system to model the connections between nodes in the tree. Each edge is considered a "path" and it could be active or inactive. 

This system is designed to be flexible enough to accommodate various features. Most notably multiple active containers, "smart" memory of which was the last active container, and tags will be able to utilize this while leaving the "containers" concept alone (and thus simple).

This unblocks a few issues, namely #73 and #63.


SnirkImmington:
Let's see if the test changes improve coverage here...

PR #108, opened 2016-09-19
Tree cleanup

Timidger:
* Added error propagation (instead of bare panics). This will allow us to do later fancy thing, e.g like having a way to recover from strange errors with a soft reset.
* Moved more things out of tree.rs
  + Added workspace to actions
* More concise error names.

### DIFF
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -50,7 +50,7 @@ impl LayoutTree {
                                 match self.tree[new_active_ix].get_type() {
                                     ContainerType::Container => {
                                         let path_ix = self.tree.follow_path(new_active_ix);
-                                        // If the pat wasn't complete, find the first view and focus on that
+                                        // If the path wasn't complete, find the first view and focus on that
                                         let node_ix = (self.tree.descendant_of_type(path_ix, ContainerType::View)).unwrap();
                                         let parent_ix = try!(self.tree.parent_of(node_ix).map_err(|err| TreeError::PetGraphError(err)));
                                         match self.tree[node_ix].get_type() {
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -2,7 +2,7 @@ use super::super::commands::CommandResult;
 use petgraph::graph::NodeIndex;
 use rustwlc::WlcView;
 
-#[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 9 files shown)


Timidger:
Please merge in after #107

src/layout/actions/focus.rs
```diff
@@ -137,4 +167,16 @@ impl LayoutTree {
             _ => panic!("Active container not view or container!")
         });
     }
+
+    /// Normalizes the geometry of a view to be the same size as it's siblings,
```
SnirkImmington: *its

src/layout/actions/layout.rs
```diff
@@ -224,6 +254,36 @@ impl LayoutTree {
         self.validate();
     }
 
+    /// Gets the active container and toggles it based on the following rules:
+    /// * If horizontal, make it vertical
+    /// * else, make it horizontal
+    /// This method does *NOT* update the actual views geometry, that needs to be
+    /// done separately by the caller
+    pub fn toggle_active_horizontal(&mut self) {
```
SnirkImmington: Maybe call this something besides `horizontal`? Like `toggle_active_standard`?

Timidger: good point. Also I should look into why the view geometry isn't actually updatedd by this method...

SnirkImmington: `switch_active_container_layout`?

Timidger: `switch_active_container_layout` is confusing, because it doesn't switch it, it'll only toggle between horizontal/vertical. It would be confusing within the context of later layout types

SnirkImmington: `swap_active_container_standard_layout`
`swap_active_container_horizontal_vertical`
`switch_active_container_layout_swap_vertical_with_default_horizontal`
I understand _why_ this needs to exist, but I dunno how else to name it either.

Timidger: hmmmm `toggle_cardinal_tiling`? Or something to the effect to mean switching between "up and down" (vertical) and "left and right" (horizontal)? I don't use "cardinal" anywhere else though, so it could be confusing, but that's the best name I can come up with.

SnirkImmington: Yeah or `toggle_standard_tiling`/`toggle_standard_tiling_modes`

PR #114, opened 2016-10-05
(Not finished) petgraph 0.3 + StableGraph

bluss:
Since I did this, I wanted to share to not have the work duplicated.

There's a failing test left in here, it's expecting an index to have shifted. I don't know what more code might be depending on indices to shift.

The code has a good handle on the issues of using `Graph`, so it's not really a problem that needs fixing (but the shorter .edge() calls in the first commit are nice anyway.)

Since you mentioned a map-based approach you could use something similar to petgraph::GraphMap. Just like StableGraph, those are not the main things of the library and are not as well tested or yet developed as `Graph`.

The roadmap includes rewriting `GraphMap` once more, so that it still uses `N` (Uuids in your hypothetical case) for node keys, but graph indices internally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ log = "0.3"
 env_logger = "0.3"
 hlua = "0.1"
 bitflags = "0.6"
-petgraph = "0.2"
+petgraph = { version = "0.3.1" }
 rustc-serialize = "0.3"
 json_macro = "0.1"
 nix = "0.6"
--- a/src/layout/core/graph_tree.rs
+++ b/src/layout/core/graph_tree.rs
@@ -7,8 +7,10 @@ use std::ops::{Deref, DerefMut};
 use std::fmt::{Debug, Formatter};
 use std::fmt::Result as FmtResult;
 
-use petgraph::EdgeDirection;
-use petgraph::graph::{Graph, NodeIndex, EdgeIndex};
+use petgraph::Direction;
+use petgraph::stable_graph::StableGraph;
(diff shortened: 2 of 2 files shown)


Timidger:
Looks good, I'll take a look tonight and fix the failing test. The code using 'edges' does look much simpler :). 

`GraphMap` sounds interesting, especially if there comes a point where we can just use `Uuid`s everywhere instead of `NodeIndex`. Looking at the docs for `GraphMap`, it looks like we would need to make the `Container`s hashable though, and we could not compare on a `Uuid` directly, because the backing value for the graph is a container,. The first solution I can think of is a kludgey new variant on the enum, something like `Lookup(Uuid)` that would wrap a `Uuid` and allow us to lookup a container, which is less than ideal but something we could consider.

bluss:
Ok, I'm sorry, then my impression of the graph node weights needed were way oversimplified.

Timidger:
No worries, the current system isn't that big of a deal. Right now I throw `NodeIndex` all over the place, but that's more a failing of mine than of the PetGraph API. So long as I thread everything through the `Uuid` hashmap to `NodeIndex`, and only deal with the indices in `graph_tree.rs`, then I can abstract the majority of the code enough that it doesn't need to worry about the implementation details of PetGraph (or that it even exists).

bluss:
Most of these changes are needed for updating to petgraph 0.4, if there would be interest for that (Graph::edges() changed to match StableGraph::edges() w.r.t iterator element type).

PR #120, opened 2016-10-12
Hotfix init file loading code

SnirkImmington:
* Error messages replace `$HOME` with `~` (nicer replacements later)
* Fix using the wrong path for `$XDG_CONFIG_DIR`
* Fix not using `$WAY_COOLER_INIT_FILE` (breaking change)
* Warnings when looking for the init file are improved

### DIFF
--- a/src/lua/init_path.rs
+++ b/src/lua/init_path.rs
@@ -29,33 +29,38 @@ pub fn get_config() -> (bool, Result<File, &'static str>) {
 fn get_config_file() -> Result<File, &'static str> {
     let home_var = env::var("HOME").expect("HOME environment variable not defined!");
     let home = home_var.as_str();
+
     if let Ok(path_env) = env::var("WAY_COOLER_INIT_FILE") {
+        info!("Found $WAY_COOLER_INIT_FILE to be defined, will look for the init file there.")
         if let Ok(file) = read_file(Path::new(&path_env)) {
             info!("Reading init file from $WAY_COOLER_INIT_FILE: {}",
                   path_env.as_str().replace(home, "~"));
             return Ok(file)
         }
-        warn!("Looking for init file: $WAY_COOLER_INIT_FILE={}, was not a valid path to a config file.",
+        warn!("Did not find an init file at $WAY_COOLER_INIT_FILE! It points to {}",
               path_env.as_str().replace(home, "~"));
     }
 
     if let Ok(xdg) = env::var("XDG_CONFIG_HOME") {
         let init_file_path = Path::new(&xdg).join("way-cooler").join(INIT_FILE);
+        info!("Found $XDG_CONFIG_DIR to be defined, will look for the init file at $XDG_CONFIG_DIR/way-cooler/init.lua");
         if let Ok(file) = read_file(&init_file_path) {
-            info!("Reading init file from $XDG_CONFIG_HOME");
+            info!("Reading init file from $XDG_CONFIG_HOME/way-cooler/init.lua");
             return Ok(file)
         }
         else {
-            warn!("Looking for init file: nothing at $XDG_CONFIG_HOME, no file {}.",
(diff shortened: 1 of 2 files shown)


src/lua/init_path.rs
```diff
@@ -27,27 +27,34 @@ pub fn get_config() -> (bool, Result<File, &'static str>) {
...
+        warn!("Looking for init file: $WAY_COOLER_INIT_FILE={}, was not a valid path to a config file.",
+              path_env.as_str().replace(home, "~"));
     }
 
     if let Ok(xdg) = env::var("XDG_CONFIG_HOME") {
-        if let Ok(file) = read_file(Path::new(&xdg).join(INIT_FILE)) {
-            info!("Reading init file from $XDG_CONFIG_HOME: {}", xdg);
+        let init_file_path = Path::new(&xdg).join("way-cooler").join(INIT_FILE);
+        if let Ok(file) = read_file(&init_file_path) {
+            info!("Reading init file from $XDG_CONFIG_HOME");
```
Timidger: This isn't very helpful to us, we should be printing what `XDG_CONFIG_HOME` points to (while also saying we got the value from that variable)

src/lua/init_path.rs
```diff
@@ -27,27 +27,34 @@ pub fn get_config() -> (bool, Result<File, &'static str>) {
...
             return Ok(file)
         }
-        warn!("$XDG_CONFIG_HOME={}, was not a valid path to a config file. Defaulting ...", xdg);
+        else {
+            warn!("Looking for init file: nothing at $XDG_CONFIG_HOME, no file {}.",
```
Timidger: Same here, we should be printing out what exact path was being looked at.

Also this warning is a bit awkward here: "nothing at $XDG_CONFIG_HOME, no file {}", why not just: "No init file at $XDG_CONFIG_HOME: {path_name}"

src/lua/init_path.rs
```diff
@@ -27,27 +27,34 @@ pub fn get_config() -> (bool, Result<File, &'static str>) {
...
+                  &init_file_path.to_string_lossy().replace(home, "~"));
+        }
     }
-    let dot_config = Path::new(&env::var("HOME")
-                               .expect("HOME environment variable not defined!"))
-        .join(".config").join("way-cooler").join(INIT_FILE);
-    trace!("Looking at {:?}", dot_config);
+    let dot_config = Path::new(home).join(".config").join("way-cooler").join(INIT_FILE);
+
+    trace!("Looking for init file at {:?}", dot_config.to_string_lossy().replace(home, "~"));
```
Timidger: Why do we say we are looking for the init file here, when we don't print this out anywhere else?

Also if you're going to keep this you should move the home replacement into a variable, because you do it twice (once here and again below in the `info!`)

src/lua/init_path.rs
```diff
@@ -27,27 +27,34 @@ pub fn get_config() -> (bool, Result<File, &'static str>) {
...
     if let Ok(file) = read_file(&dot_config) {
-        info!("Reading init file from {:?}", &dot_config);
+        info!("Reading init file from {:?}", &dot_config.to_string_lossy().replace(home, "~"));
```
Timidger: Why are you using `to_string_lossy` here? Why not `as_str` like you did everywhere else?

SnirkImmington: This variable is a `Path`, the others were `OsStr` or whatever string type you get from environment variables.

Timidger: Right, my mistake. What I meant was `to_str` (which is implemented for `Path`)

SnirkImmington:
Improved the messages somewhat, and added a warning at the end describing where the init file can be placed. Also mentioned that next version will not require this check.

Timidger:
Will merge in when docker passes

PR #121, opened 2016-10-12
Added proprietary driver warnings

Timidger:
* Added proprietary driver warnings
* Added checksums to Cargo lock file (I believe it's a new feature from the most recent Cargo)


src/main.rs
```diff
@@ -93,6 +97,34 @@ fn log_format(record: &log::LogRecord) -> String {
...
+        // If DISPLAY is present, we are running embedded
+        return
+    }
+    match File::open(Path::new(MODULE_PATH)) {
+        Ok(file) => {
+            let reader = BufReader::new(&file);
+            for line in reader.lines() {
+                if let Ok(line) = line {
+                    if line.contains("nvidia") {
+                        error!("Warning: Proprietary nvidia drivers do NOT support Wayland. Use nouveau.");
```
SnirkImmington: I think we should say something like `try using noveau` or `consider using noveau` or `noveau, open source nvidia drivers, do support Wayland`. My desktop won't boot with noveau drivers last I tried, it's not always an option to switch to. Perhaps something like `Proprietary nvidia graphics drivers are installed, but they are not compatible with Wayland. Consider using noveau drivers for Wayland.`

Timidger: Updated to be more friendly

PR #127, opened 2016-10-22
Update dependencies

SnirkImmington:
Update direct dependencies `bitflags = 0.7` and `petgraph = 0.3` in anticipation of 0.3.3.
`petgraph` is actually at `0.4` right now, but since that introduces backwards-breaking changes we can skip it for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 name = "way-cooler"
 version = "0.3.2"
 dependencies = [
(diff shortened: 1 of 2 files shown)


Timidger:
Did you ensure pet graph updates did not affect the tree?

Timidger:
Ok I tested it, the tree seems to work mostly the same.

Issue #115, opened 2016-10-07
Nothing but background shows


AdrianSchurz:
Greetings,
been itching to go wayland for so long now and when I saw something awesomewm-inspired my heart skipped a beat. I am burning to play with this thing.

For the moment, though, if I try running it on both of my up to date arch boxes, only the (purple) background shows. No other functionalities/hotkeys appear to work besides my custom cursors being used and Alt+Shift+Esc actually exiting.

This, and no discernibly different, behaviour shows whether I install from aur or via
`cargo install way-cooler,`
and whether run from within an x-session or on a separate tty.

Since it showed some similarity to #52 I tried with wlc versions 0.0.6 through 0.0.3 as well, no difference.

I would have tried "removing dlopen from Cargo.toml". which I interpreted as, roughly, cloning the repo, 
taking things like `["client", "dlopen"]` to `["client"]` inside Cargo.toml and then doing
`cargo install --path <pathtorepo>`, but the ensuing compilation fails with

`Compiling way-cooler v0.3.2 (file:///home/ulmeyda/repositories/way-cooler)
error:`
`no method named "has_parent" found for type "&mut layout::core::graph_tree::InnerTree" in the current scope
   --> src/layout/core/graph_tree.rs:186:43`
`186 |         if cfg!(debug_assertions) && self.has_parent(child_ix) {`

which, understandably, is what currently happens when installing from the git version aur package, as well.

Would appreciate some leads to resolve it. Am still hyped.

Many kisses

Timidger:
Salutations.

Yes, I accidentally messed up the release build for the master branch :(. That's fixed now, and should be merged into master shortly. 

Good on you for doing your research. I suggest some next steps in order for us to help you would be to give us the output from Way Cooler after you start it up and try to do things. Right now we don't log to a file by default, so you'll have to run it with something like `way-cooler 2> dump.txt`, and send us the ensuing file. 

Also just to be sure, do you have `dmenu` and weston/weston-terminal installed on your machine? We use weston-terminal as the default terminal emulator in the config file, since it's one of the only wayland backed terminal emulators, and thus a good litmus test for basic functionality. Unfortunately, it's not included by default and it not appearing is a very soft warning in the log output only. 

If you have both of those, something else must be afoot, and hopefully we can figure it out by pouring over your log files :).

AdrianSchurz:
Ólà,

thanks for the swift reply. Aur and cargo packages both compile.
[...]

PR #128, opened 2016-10-22
Travis deploy

SnirkImmington:
This PR adds deploy automation and more code checks in `.travis.yml`.

**Overall**
- Improvements in speed
- Clarity of the `travis.yml`
- Add `ci.py` with some actions which can be run locally

**Per-branch**
- Code will be checked for being able to compile without `--test` (prevent #115 issues where master branch didn't compile)

**Release branches**
- Travis should check if it is being run under a branch named `release-{version}`
- Travis should check that version numbers in the readme badges, `Cargo.toml`, and `Cargo.lock` match `{version}`
- ~~Also runs `cargo check --release` just in case~~

**Deploy**
- Travis is hooked into GitHub releases when a tag is pushed
- Should rebuild way-cooler and attach a `way-cooler_linux_x86_64` and `init.lua` to a GitHub release
- Should deploy to crates.io as well

Obviously, we won't know if the release works until we try it. This current script was made with many, many attempts and re-runs. The `release-0.3.3` branch may contain more fixes. If the upcoming 0.3.3 deploy does not work we will do so ourselves.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ notifications:
     on_success: :change
     on_failure: :change
   email: false
-# load travis-cargo
+# install travis-cargo
 before_script:
   - |
       git clone https://github.com/Timidger/travis-cargo
@@ -30,7 +30,11 @@ script:
   - cargo test --verbose
 
 after_success:
-  # measure code coverage and upload to coveralls.io (the verify
-  # argument mitigates kcov crashes due to malformed debuginfo, at the
-  # cost of some speed <https://github.com/huonw/travis-cargo/issues/12>)
+  # Test and upload coverage to coveralls
+  # Justification of excluded files:
+  # main.rs: just initialization of logging, sigkill handler, and initializing lazy static objects. (Arg parsing should be moved into a different module and tested)
+  # mod.rs: although at the moment we have too much code in various mod.rs's these are not really testable either way
+  # tests.rs: should not be running on our tests themselves
+  # wayland_client_api.rs: we really don't have a good way of testing any of this. At the moment it's simple enough that if it didn't work it would be obvious in user testing.
+  # callbacks.rs: we don't have any integration tests for callbacks right now. They'd need to deal with the lazy statics that the other components of the program live in.
   - travis-cargo coveralls --no-sudo --verify --exclude-pattern="main.rs,mod.rs,tests.rs,wayland_client_api.rs,callbacks.rs"
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ install:
   - pip install --user -e ./travis-cargo
   # install dependencies for ci.py
   - pip install --user docopt
-  - cargo install cargo-check
 
 script:
   # check integrity of Lua files in /lib/lua/
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ script:
   # run tests
   - cargo test --verbose
   # run tag checks
-  - ci.py travis-check
+  - python $TRAVIS_BUILD_DIR/ci.py travis-check
 
 after_success:
   # Test and upload coverage to coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ before_deploy:
   - ci.py prepare-deploy
 deploy:
   provider: releases
-  skip_cleanup: false # We can skip cleanup as we recompile with --release
+  skip_cleanup: false # We do not need to skip cleanup as we recompile with --release
   on:
     tags: true
   api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-language: rust
+language: 
+  - rust
+  - lua
 rust:
   - stable
 cache: cargo
@@ -25,9 +27,7 @@ before_script:
       pip install -e . --user &&
       export PATH=$HOME/.local/bin:$PATH)
 script:
-  - echo "Checking Lua libs integrity"
-  - find -X lib/lua -type f -name "*.lua" | xargs luac -p
-  - echo "Checking init.lua integrity"
+  - find lib/lua -type f -name "*.lua" | xargs luac -p
   - luac -p config/init.lua
   - cargo check
   - cargo test --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: rust
 rust:
   - stable
+ sudo: false
 
 cache:
   - cargo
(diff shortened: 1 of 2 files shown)


SnirkImmington:
Even in opening a PR for the Travis stuff, we find a bug in the PR checker...

.travis.yml
```diff
@@ -1,34 +1,77 @@
...
+  # tests.rs: should not be running on our tests themselves
+  # wayland_client_api.rs: we really don't have a good way of testing any of this. At the moment it's simple enough that if it didn't work it would be obvious in user testing.
+  # callbacks.rs: we don't have any integration tests for callbacks right now. They'd need to deal with the lazy statics that the other components of the program live in.
   - travis-cargo coveralls --no-sudo --verify --exclude-pattern="main.rs,mod.rs,tests.rs,wayland_client_api.rs,callbacks.rs"
+
+before_deploy:
+  - ci.py prepare-deploy
+deploy:
+  provider: releases
+  skip_cleanup: false # We can skip cleanup as we recompile with --release
```
Timidger: Wait... is this skipping clean up or not? The comment seems to indicate we can, because we recompile with --release, but `skip_cleanup` is set to `false`. Or is this correct and the option is just reading confusingly to me.

.travis.yml
```diff
@@ -1,34 +1,77 @@
...
+  on:
+    tags: true
+  api_key:
+    secure: "i+7X7qZgq5gwWbi/KBoL03rBTJam2ayJUxuTUGnQNwOEUc1KaDintZG/j6bp2DqFfodBx0oq/nvZi+mIAzbpQHlaDCav+GC0S3o2k10CAPlcg844IZVB22J/WbUjFfkpAWz0Mbs6U7cjCeAPAdYTbcJy5haduKm9FKOat87bBrNIdPZno3j7RDd1ne3+gsbEp0W9788jqX99vV7CWtV0hT5w1yT+COg8YdvCowOLh3mS/y43LZlmyqNFqJpF/tMnBWI9qDlhyCCzmsVb+xq9tjX4XQAgZJI6Y0uP8JoityqG1OpCAfgSpS99+0iy+lSGpALDfKAk04lIRMWublnMSXSqcHRupxF6mvlzer9V4XV1emgDn1rtnIkQiRryuA180mTVpHjGg+Ockl/wjOo4FJGKlDmQn7oT6EMDfY62md2tYbsB2MmJNlFXiQCinNFmjpTJuUbHwXmF4FArq4B8AmoFwY1I7I8x6eHKqLEZOxgm9EgtHC9+dwbRSY7t6X7AiHYv7LsDIjEO70oG7j/nuoZgBZZvMjOWlm7EKUjrcxXVLUZK9f9P/6Pc7r+OQr4jaRRclCtZPsPGPu/EjNfeQ0+PpZdByzT9k1y4loAqZPtP0ANBp7flKsPNWXaMSNsb/Hu4zolDDyeU5eL3VHyrDk1wi9nmFTOnrCl8W++7Fi4="
```
Timidger: What is this key for? Is it a github oauth token?

ci.py
```diff
@@ -0,0 +1,100 @@
...
+  ci.py (-h | --help)
+  ci.py --version
+
+Options:
+  -h --help    Show this menu
+  -v           Be verbose, print actions taken
+  --version    Show version information
+"""
+
+failed = False
```
Timidger: This flag is never used

ci.py
```diff
@@ -0,0 +1,100 @@
...
+def check_release_branch(version):
+    all_clear = True
+    for (file_name, file_regex) in FILE_REGEX_MAP.items():
+        print("Checking " + file_name)
+        if not check_file_version(file_name, file_regex, version):
+            all_clear = False
+    return all_clear
+
+if __name__ == "__main__":
+    print("Running ci.py...")
```
Timidger: Small thing, but you could also put `sys.argv[0]` instead of hardcoding the name of the file, though I doubt it would ever be an issue.

ci.py
```diff
@@ -0,0 +1,100 @@
...
+        print("Running travis-check...")
+        travis_pr_branch = os.environ["TRAVIS_PULL_REQUEST_BRANCH"]
+        if travis_pr_branch == "":
+            print("Not running in a PR.")
+            sys.exit(0)
+        print("PR " + travis_pr_branch + " detected, checking for versions.")
+        version_match = re.match(BRANCH_REGEX, travis_pr_branch)
+        if not version_match:
+            print("Not in a release branch PR.")
+            sys.exit(0)
```
Timidger: Should we be exiting cleanly if `TRAVIS_PULL_REQUEST_BRANCH` was non-empty and didn't match the branch regex? Or are there cases where it'll be set but it should just be ignored.

ci.py
```diff
@@ -0,0 +1,100 @@
...
+    elif args["bump"]:
+        sys.stderr.write("Not supported yet")
+        sys.exit(1)
+
+    elif args["prepare-deploy"]:
+        print("Not compiling for multiple targets yet :(")
+        print("cargo build --release --verbose")
+        retcode = subprocess.call(["cargo", "build", "--verbose", "--release"])
+        if retcode != 0:
+            sys.stderr.write("Cargo build exited with %s\n", retcode)
```
Timidger: You're using `%s` formatting on an integer, which technically isn't an issue here since these will always be whole numbers, but you should really just be using `{}` and `.format`

ci.py
```diff
@@ -0,0 +1,100 @@
...
+            sys.exit(2)
+
+        print("Moving way-cooler => way-cooler_linux_x86_64")
+        build_dir = os.environ["TRAVIS_BUILD_DIR"]
+        os.rename(build_dir + "/target/release/way-cooler", build_dir + "/way-cooler_linux_x86_64")
```
Timidger: Not a comment on your code, but wow `os.rename` is a dumb name. That should be `os.move`, it doesn't even make sense from a platform agnostic standpoint...

ci.py
```diff
@@ -0,0 +1,100 @@
...
+
+    else:
+        sys.stderr.write("Invalid arguments!\n")
```
Timidger: Print usage here.

.travis.yml
SnirkImmington: I had originally set it to `skip_cleanup: true` accidentally, _I_ was confusing myself too.

.travis.yml
SnirkImmington: Yes. Will add a comment.

ci.py
SnirkImmington: I think travis sets `TRAVIS_PULL_REQUEST_BRANCH` to `""` in regular builds, added the check for it not being defined.

PR #132, opened 2016-10-29
Resizing windows (Fixes #68)

### LINKED ISSUE #68: Resize windows in the layout
This will be slightly more complicated than it is in the simple example for rust-wlc. You need to shrink the neighbors as much as you grow the window, in order to not expose the background at all. For floating windows they will be freely resizable.

Timidger:
# Features
- Added resizing
  - To resize a container, hold down `ctrl` and then click and hold the secondary mouse button ("right click") on the window you'd like to resize.
    - It is not yet configurable, this will require changes to the configuration file and might affect other components, so this has been put off.
    - It is exactly the same as the "move window" keybinding, but with the right button instead of the left.
  - The mouse pointer follows in the direction of movement.
    - E.G: If you are resizing in the top-left direction, it jumps to the top left corner of the window you are resizing.
  - Can resize floating windows 
    - You can do so freely, with no restrictions except for the size of the screen (and thus the limits of your mouse pointer)
  - Can resize tiled windows
    - This automatically adjusts the size of the neighboring containers, ensuring that maximum screen coverage is maintained
      - This is achieved by doing an opposite action for every resize. Eg if we make window A bigger by 5 pixels we make sure neighboring window B smaller by 5 pixels.
    - You grab by "edge" and adjust by moving the mouse.
      - If the edge shares a neighbor with a window that is conceptually higher up in the tree (e.g it is an sibling of an ancestor), then the resize affects _both_.
- Fixed Size 0 bug
  - If a window reaches size `0` (in either height or width), it is automatically resized so that it does not disappear. This is most apparent when resizing with nested containers or when you have a lot (> 30) windows on the screen at the same time.
- Fixed debug-enabled panic when floating a window after switching to it from another floating window.
# Tree Internals
- Added a new field to `GraphTree`, `view_map`
  - Maintains mapping from `WlcView` to `NodeIndex` to ensure lookups are computationally efficient.
  - Necessary because callbacks return a `WlcView` and there's currently no way to get the underlying `NodeIndex`, `Container`, or `Uuid` from that except by searching the entire tree, which is very expensive.
- Added `MIN_SIZE` constant
  - Currently only used by resize code, the layout code uses different logic to determine "minimum size" when correcting for the Size 0 bug.
  - Hardcoded to be 80x40
- Fixed a few {over,under}flow bugs that cropped up.


src/callbacks.rs
```diff
@@ -163,36 +177,87 @@ pub extern fn view_request_geometry(_view: WlcView, _geometry: &Geometry) {
...
-                        },
-                        err => error!("Could not set active container {:?}", err)
-                    }
-                });
+    if state == ButtonState::Pressed {
+        if button == 0x110 && !view.is_root() {
+            if let Ok(mut tree) = try_lock_tree() {
+                tree.set_active_view(view)
+                    .unwrap_or_else(|err| {
+                        error!("Could not set active container {:?}", err);
```
Timidger: Ok, I'll leave the logging in but add a check so it doesn't print for the background

src/callbacks.rs
```diff
@@ -163,36 +177,87 @@ pub extern fn view_request_geometry(_view: WlcView, _geometry: &Geometry) {
...
+                        edges: ResizeEdge::empty()
+                    };
+                    set_performing_action(Some(action));
+                }
+            }
+        } else if button == 0x111 && !view.is_root() {
+            if let Ok(mut tree) = try_lock_tree() {
+                tree.set_active_view(view)
+                    .unwrap_or_else(|err| {
+                        error!("Could not set active container {:?}", err);
```
Timidger: There's could be a lot of things that could go wrong with setting an active view (eg, there might be an issue setting the active path. Right now the active path isn't allowed to go into floating containers, and this could explain WHY the user facing issue (eg the inability to focus on something) happened)

src/callbacks.rs
```diff
@@ -204,25 +269,45 @@ pub extern fn pointer_scroll(_view: WlcView, _time: u32,
...
-            Err(TreeError::PerformingAction(_)) => {},
-            Err(TreeError::Movement(MovementError::NotFloating(_))) => {},
-            Err(err) => error!("Error: {:#?}", err)
+    match maybe_action {
+        None => result = false,
+        Some(action) => {
+            if action.edges.bits() != 0 {
+                if let Ok(mut tree) = try_lock_tree() {
+                    // TODO Change to id of _view
+                    // Need to implement a map of view to uuid first though...
```
Timidger: Haha woops, implemented the map but didn't update this piece of code...that's important!

src/layout/actions/layout.rs
```diff
@@ -170,9 +176,16 @@ impl LayoutTree {
                         if scale > 0.1 {
                             scale = geometry.size.h as f32 / scale;
                             let new_size_f = |child_size: Size, sub_geometry: Geometry| {
+                                let height = if child_size.h > 0 {
+                                    child_size.h as f32
+                                } else {
+                                    // If the width would become zero, just make it the average size of the container.
```
Timidger: That should be height, but otherwise I want to keep the comment here

src/layout/actions/resize.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+    new_geo
+}
+
+/// If the geometry is at the minimum size (in either the x or y plane)
+/// and the pointer is trying to make it even smaller in that direction,
+/// the it returns true (to indicate you should abandon all operations).
+///
+/// Otherwise returns false
+fn is_making_too_small(geo: Geometry, edge: ResizeEdge, cur_point: Point,
```
Timidger: that's a better name

src/layout/commands.rs
```diff
@@ -368,4 +370,26 @@ impl Tree {
...
+    pub fn resize_container(&mut self, id: Uuid, edge: ResizeEdge, pointer: Point) -> CommandResult {
+        if let Ok(mut lock) = try_lock_action() {
+            if let Some(ref mut action) = *lock {
+                if try!(self.0.lookup(id)).floating() {
+                    self.0.resize_floating(id, edge, pointer, action)
+                } else {
+                    self.0.resize_tiled(id, edge, pointer, action)
+                }
+            } else {
+                // TODO errors
```
Timidger: Well the point is that I return SOME kind of error, I elided this because I wasn't sure how to expand it. One of the issues with returning a generic error with CommandResult, is if I want to expand the Error enum (as I would here, by adding an error for the action already being performed), I would have to change all the enums...it seems to be one of the downsides of Result style error handling, it can be difficult to expand because not all the cases are handled everywhere...wild carding helps, but that isn't always what you want obviously.

src/callbacks.rs
```diff
@@ -117,8 +117,22 @@ pub extern fn view_request_move(view: WlcView, _dest: &Point) {
...
-pub extern fn view_request_resize(_view: WlcView, _edge: ResizeEdge, _location: &Point) {
-
+pub extern fn view_request_resize(view: WlcView, edge: ResizeEdge, point: &Point) {
+    if let Ok(mut tree) = try_lock_tree() {
+        match try_lock_action() {
+            Ok(guard) => {
+                if guard.is_some() {
+                    if let Some(id) = tree.lookup_view(view) {
+                        if let Err(err) = tree.resize_container(id, edge, *point) {
+                            error!("Context: Trying to resize a container for the user \
```
Timidger: Ha, I was trying out that new error reporting that I read in that blog post...I totally forgot about it, so I think I'll just remove it here. It's confusing if it's only in one place ;).

I wouldn't say our logs are a user interface, the point of this error was to help the debugger (eg us) understand what the user was doing without having to mentally recreate the scene.

jhailman: > the point of this error was to help the debugger (eg us) understand what the user was doing without having to mentally recreate the scene. 

I'm not sure that warrants too much extra information in the log - we already have the line in the source code, so we know it's in the callbacks, and it's the tree failing to resize a container. The only way we're _really_ going to know what the user was doing without having to mentally recreate the scene is to `trace!`-log every single user interaction, and I think our layout code is stable enough that we shouldn't have to do that in production code. We should either try immediately logging the state of the tree + information (values of variables, not a description of the code) about the action being called, or only enable `trace!` logs for debug builds and try to recreate errors on our machines in response to bug reports.

Timidger: :heavy_check_mark:

src/callbacks.rs
Timidger: Scratch that, just removing it since a run-time check would be too obnoxious 

:heavy_check_mark:

src/callbacks.rs
Timidger: :haircut_man:

src/layout/actions/layout.rs
Timidger: :heavy_check_mark:

src/layout/commands.rs
Timidger: Ugh, it was annoying, but I integrated action errors in.

I'm thinking I'll need to do a pass where I remove all error handling (and maybe some of the more complex logic) from the commands wrapper, because that's starting to get a bit annoying.

:heavy_check_mark:

Timidger:
Hm, yes that bug you are experiencing is one I had fixed in an earlier iteration of this PR, but I forgot to fix it on this attempt.

I had dubbed it "window smashing", if you look at commit d5458fc,  you'll see I attempted to correct some of the more overt examples of this bug 3 weeks ago, but I did not fix the extreme versions as I had hoped it would be fazed out as I tried different methods of resizing.

I forgot about it until now, and I will look into fixing it next. It _should_ be easy, but I'll keep you updated if I run into any issues.

PR #134, opened 2016-11-19
Added background program

Timidger:
* Added `way-cooler-bg` as a separate program
  + Can still change the background by changing the same field in `init.lua`.
  + Updated `init.lua` to call this program after everything else is done, this should probably be moved to a much more sophisticated "post-init" callback that will be invoked by Way Cooler after wlc's init method has been called.
* `way-cooler-bg` now supports the following image formats:
  + PNG
  + JPEG
  + GIF
  + BMP
  + ICO
* Re-organize the project structure to accommodate `way-cooler-bg`.
  + Possibly this structure should change, although I think it best if `way-cooler-bg` exists within the same project as `way-cooler`.
* Added default background image (subject to change)


way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn as_u32(&self) -> u32 {
+        unsafe { transmute(self.0)}
+    }
+}
+
+fn main() {
+    let args: Vec<_> = env::args().collect();
+    if args.len() < 3 {
+        println!("Please supply either a file path or a color (written in hex)");
+        println!("Please also supply a path to a cursor image");
```
SnirkImmington: Can we not allow the user to set the background without setting the cursor?

way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+    }.expect("could not generate image");
+
+    background_surface.commit();
+    background_surface.set_buffer_scale(1);
+    let mut cursor_surface = compositor.create_surface();
+    let _cursor_buffer = self::cursor_surface(cursor_path.as_str(), &mut cursor_surface, &env);
+    main_background_loop(background_surface, cursor_surface, evt_iter, &env);
+}
+
+fn weird_math(num: u8, third_num: u32) -> u8 {
```
SnirkImmington: ...

way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+            pixel[0] = tmp;
+        }
+    }
+    let vec = image.into_vec();
+    let mut tmp = tempfile::NamedTempFile::new().expect("Unable to create a tempfile.");
+    tmp.set_len(size as u64).expect("Could not truncate length of file");
+    tmp.write_all(&*vec).unwrap();
+
+
+    // TODO I want this shit outta here
```
SnirkImmington: ...

way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+    let shm = env.shm.as_ref().map(|o| &o.0).unwrap();
+
+    // Create the surface we are going to write into
+
+    let pool = shm.create_pool(tmp.as_raw_fd(), size as i32);
+    let background_buffer = pool.create_buffer(0, width as i32, height as i32, stride as i32, WlShmFormat::Argb8888);
+    // Tell Way Cooler not to put this in the tree, treat as background
+    // TODO Make this less hacky by actually giving way cooler access to this thing...
```
SnirkImmington: Do you know how the programs should talk to each other? We could use DBus, if you want the user calling the Lua `exec` on the background we could give the background privileges and expose "get display resolution" and "set this view ID as background for this display"

way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+        let width = image.width();
+        let height = image.height();
+        let stride = width * 4;
+        let size = stride * height;
+        // TODO Split this into its own function
+        {
+            let pixels = image.enumerate_pixels_mut();
+            for (_x, _y, pixel) in pixels {
+                let alpha = pixel[3] as u32;
+                pixel[0] = weird_math(pixel[0], alpha);
```
SnirkImmington: Ah, here's what `weird_math` is for. If you split this into its own function (like your comment says) you might be able to give it a better name.

way-cooler-bg/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+                _ => { /* unhandled events */ }
+            }
+        }
+        event_iter.dispatch().expect("Connection with the compositor was lost.");
+    }
+}
+
+
+#[test]
+fn test_weird_math() {
```
SnirkImmington: :+1: I'm glad you test the weird math, wouldn't want it to be _too_ weird.

way-cooler/README.md
```diff
@@ -0,0 +1,132 @@
...
+  * Allows extensions of the window manager to exist as separate programs talking over the IPC
+- A Lua environment designed to make extending Way Cooler simple and easy
+  * Lua is the configuration format, allowing the user to enhance their window manager in any way they want.
+  * Utilities library included to aid communicating with Way Cooler
+- X programs supported through XWayland
+
+## Planned Features
+
+- i3 tabbed/stacked tiling
+- Floating windows
```
SnirkImmington: You have the honor of removing this from the "planned features" list. I dunno what improvements you wanted to make to floating windows still but we definitely have them.

way-cooler/README.md
```diff
@@ -0,0 +1,132 @@
...
+- Wayland
+  * Including the server and client libraries
+- wlc
+  * Installation instructions can be found on [their github page](https://github.com/Cloudef/wlc)
+- Weston (optional)
+  * The init file defaults to using `weston-terminal` as the default terminal emulator
+- Cargo
+  * The package manager / build system used by Rust
+
+Finally, to install Way Cooler simply run the following cargo command:
```
SnirkImmington: Do you want to mention compiling `way-cooler-bg` here?

way-cooler/config/init.lua
```diff
@@ -25,8 +25,8 @@ config.init_workspaces(workspace_settings) -- Not implemented yet
 -- Background
 --
 --
--- A background can either be a 6 digit hex value or an image path (not yet supported)
-way_cooler.background = 0x5E4055
+-- A background can either be a 6 digit hex value or an image path
+way_cooler.background = "../way-cooler-bg/assets/background.jpg"
```
SnirkImmington: Where did you get the background image? What's the license on it?

way-cooler/config/init.lua
```diff
@@ -114,3 +114,9 @@ end
 
 -- !! Do not place any code after this comment.
 -- !! way-cooler and plugins may insert auto-generated code.
+local status = os.execute("which way-cooler-bg 2>/dev/null")
```
SnirkImmington: Did you want to put this under the `!! Do not place any code after this comment` comment? Do you want a `way-cooler.exec(path)` Lua method for the init file? 
This is something in the init file we'll have to change a lot - we may be running the background as an IPC client with IPC privileges in the future, and we won't want to be `print`ing things once we have graphical error messages. Obviously that is very far in the future. 
Do you want to move this to a wrapped Lua method, like `way-cooler.set_background(image_or_color)` which in Rust handles finding and executing `way-cooler-bg` and logging errors? It might help us change the way background is set internally without breaking backwards compatibility in the init file.

way-cooler/src/callbacks.rs
```diff
@@ -117,8 +117,21 @@ pub extern fn view_request_move(view: WlcView, _dest: &Point) {
     }
 }
 
-pub extern fn view_request_resize(_view: WlcView, _edge: ResizeEdge, _location: &Point) {
-
+pub extern fn view_request_resize(view: WlcView, edge: ResizeEdge, point: &Point) {
```
SnirkImmington: It looks like this PR is behind resizing and the `git mv`s may have screwed it up (might also be annoying on future PRs). Not sure what to do here.

SnirkImmington:
`ci.py` should stay in the crate root with `.travis.yml` - we'll want to use it with other `way-cooler` clients.

Timidger:
Hmm..I accidentally left the resize commits in here...I rebased and updated the version on my branch, but it's not updating here (possibly because it's across repos?)....Since it's the same commits (and resize was just merged in), I'll just merge in master and everything will be fine, annoying that it's duplicated in the history however

way-cooler-bg/src/main.rs
Timidger: We _could_, but then we run into the same issue as before where the background program will fail if it can't find an appropriate cursor to use from the default theme. A better way to think of this cursor is a "falback" option if a cursor can't be found (like on Ubuntu). If you want I can change it so you don't have to specify a fallback option, but since we _should_ already have one (check the `assets/` folder) I decided to just keep it required. 

Also, this program should _never_ be called directly by the user, it should be called by way cooler / lua, so ultimately this decision will not affect the user thankfully.

way-cooler-bg/src/main.rs
Timidger: Yes, probably a bad name. I wasn't sure what that conversion process was called (it's the same one eg sway uses), but all I know is it goes from RGBA ->Some slightly different values for RGBA that are somehow correct :tm:

way-cooler-bg/src/main.rs
Timidger: Ah, thought I got that comment. I have, indeed, already gotten that shit outt here

way-cooler-bg/src/main.rs
Timidger: Another comment that is in the wrong place. This was talking about how we tell way cooler that this is supposed to be the background surface. Right now we are being _super_ hacky and setting the class name for the view to be "Background" which Way Cooler treats differently.

The _correct_ way (and the way eg Sway does it) would be to use a well defined wayland protocol (see [how sway does it](https://github.com/SirCmpwn/sway/blob/master/protocols/desktop-shell.xml#L10)) where we pass the surface we want to Way Cooler (and then it would display it).

Using dbus would also be fine, and probably the way we ultimately go with. Either way, to make this secure, you are correct and we should have Lua spawn this program with the appropriate credentials. 

I don't knew enough about how Lua / HLua works to set that up in a better way than I already have (which is admittedly pretty bad/hacky). If you can think of a better way right now, without dbus, to spawn the background feel free to make changes. We can add dbus "privileges" for the background later

way-cooler-bg/src/main.rs
Timidger: Ah, I was going to split this into it's own method, and I can't remember why exactly I didn't...I think it was weird type issues because this is from Piston' `Image` library, and it has weird types with a bunch of generics that were not happy with me. I'll take another look at it.

way-cooler/README.md
Timidger: Hm, yes I think we should change a significant portion of our README now, especially as we need to move to a different distribution model. It's totally possible for the user to compile `way-cooler-bg` and then use it in their init file, but it would be much more work, and at this point having to compile two separate programs (wlc and way-cooler-bg) just to make way-cooler work 100% is a little much. 

We can keep the manual steps, but I think first and foremost we should recommend people use our binary releases (which will have this packaged up as nicely as possible). Of course, that will be replaced by rpm/apt packages which are much easier to maintain/update.

way-cooler/config/init.lua
Timidger: Ha, you got me. I tried to find one in the public domain that looked nice (using [this](https://search.creativecommons.org/)), but couldn't find one I liked. Feel free to find a better one to use, as I'm not sure what this is licensed under. Eventually, we should contract someone (Donbot?) to make one for us.

way-cooler/config/init.lua
Timidger: Yes, that's what I wanted to do, but didn't have enough Lua/hlua knowledge to know how to go about it (in particular, the thing I was planning was for a Lua function to be registered to be called after wlc init, which could then spawn the program, and provide a nice dumping ground for other "startup" programs). 

If you wouldn't mind, it would be much faster if you made that change.

way-cooler/src/callbacks.rs
Timidger: I'll fix it, sorry

Timidger:
Yes, I shouldn't have moved `ci.py`.

There's been a lot of changes, as my blog post write up said. I was planning on all these changes to be 0.4, unless you have a good reason to hold off on such a big number change. If you want, we can do a 0.3.4 release that doesn't have dbus (or the split up background program), and then also release 0.4 with dbus + background program.

way-cooler/README.md
Timidger: README needs a general update, I'll push that change off until then.

way-cooler/src/callbacks.rs
Timidger: Fixed

way-cooler-bg/src/main.rs
Timidger: fixed

way-cooler-bg/src/main.rs
Timidger: fixed

way-cooler-bg/src/main.rs
Timidger: fixed comment

Timidger:
Moved this code/feature to a separate program, we are keeping the directory structure. Please direct all further comments/issues to [this repo](https://github.com/Immington-Industries/way-cooler-bg)

PR #155, opened 2016-12-04
Added GDM entry file for way-cooler

### LINKED ISSUE #154: Add way-cooler to gdm
I installed the AUR package and way-cooler is not in gdm's window manager list.

starfys:
Resolves #154

### DIFF
--- a/way-cooler.desktop
+++ b/way-cooler.desktop
@@ -3,5 +3,5 @@ Encoding=UTF-8
 Name=Way Cooler
 Comment=Customizable Wayland compositor
 TryExec=way-cooler
-Exec=way-cooler
+Exec=dbus-launch --exit-with-session way-cooler
 Type=Application
--- a/way-cooler.desktop
+++ b/way-cooler.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Encoding=UTF-8
+Name=Way Cooler
+Comment=Customizable Wayland compositor
+TryExec=way-cooler
+Exec=way-cooler
+Type=Application


starfys:
Note that for Arch/AUR packaging, this file should be stored as /usr/share/xsessions/way-cooler.desktop

Timidger:
Haven't had a chance to test this yet, but I believe the command to be executed should be `dbus-launch --exit-with-session way-cooler`. This will properly launch the dbus server for Way Cooler when ran headless (Most X WM's already start this, which is why this is unnecessary when testing within X)

starfys:
Added the line for DBUS. TryExec stays the same because it's used to check if way-cooler is installed. I installed GDM and tested the session using the new conf file. way-cooler starts, but 2 mouse pointers are shown. One of them is from GDM and the other is for way-cooler.  I logged into a Weston session from GDM and got the same mouse bug there.

starfys:
This issue seems related to SirCmpwn/sway#759

starfys:
After changing the location of way-cooler.desktop from ```/usr/share/xsessions/way-cooler.desktop``` to ```/usr/share/wayland-sessions/way-cooler.desktop```, there was no change. However, if I Ctrl+Alt+F2 to a TTY and then Ctrl+Alt+F7 back to way-cooler, the hardware cursor from GDM disappears.

Timidger:
Hm, yes I'm getting the same issue on my testing. From the linked discussion on sway, it looks this is a gdm bug? I know that this bug appeared when using light-dm on Ubuntu, I haven't had time to test gdm (borked Ubuntu trying to install gdm, somehow -_-). Hopefully there's some magic we can do to tell gdm to not do that. Is gdm running Way Cooler in an embedded X environment? Is that why two pointers are showing?

starfys:
It's not running it within an X session. At least on Arch, GDM uses Wayland as its backend by default.

Timidger:
After moving `way-cooler.desktop` to `/usr/share/wayland-sessions/` it started working flawlessly for me on my fedora machine using lightdm. I'll retest with gdm and see if I can reproduce the weird cursor issue.

PR #167, opened 2016-12-10
In-place restart

Timidger:
Added ability to restart Way Cooler in place (default keybinding is `<mod>+shift+r`). Reloads the configuration file, including background program arguments, keybindings, and lua function hooks.

* Added `way_cooler_restart` command
  + Not exposed via D-Bus
* Added `way_cooler_restart` lua function hook
  + When Way Cooler is restarted in place, all the code in `way_cooler_restart` is executed
* Added `way_cooler_terminate` lua function hook
  + When Way Cooler is terminated, all the code in `way_cooler_terminate` is executed

### DIFF
--- a/config/init.lua
+++ b/config/init.lua
@@ -105,6 +105,10 @@ for _, key in pairs(keys) do
     config.register_key(key)
 end
 
+function cleanup_background()
+  os.execute("pkill way-cooler-bg")
+end
+
 
 -- Execute some code after Way Cooler is finished initializing
 function way_cooler_init()
@@ -118,8 +122,13 @@ end
 
 --- Execute some code when Way Cooler restarts
 function way_cooler_restart()
-    os.execute("pkill way-cooler-bg")
+  cleanup_background()
 end
+
+function way_cooler_terminate()
+  cleanup_background()
+end
+
 
 -- To use plugins such as bars, or to start other programs on startup,
 -- call util.exec.spawn_once, which will not spawn copies after a config reload.
--- a/config/init.lua
+++ b/config/init.lua
@@ -116,6 +116,11 @@ function way_cooler_init()
   end
 end
 
+--- Execute some code when Way Cooler restarts
+function way_cooler_restart()
+    os.execute("pkill way-cooler-bg")
+end
+
 -- To use plugins such as bars, or to start other programs on startup,
 -- call util.exec.spawn_once, which will not spawn copies after a config reload.
 
--- a/config/init.lua
+++ b/config/init.lua
@@ -85,7 +85,8 @@ local keys = {
   key({ mod }, "e", "horizontal_vertical_switch"),
   key({ mod }, "q", "close_window"),
   key({ mod, "Shift"}, "space", "toggle_float_active"),
-  key({ mod }, "space", "toggle_float_focus")
+  key({ mod }, "space", "toggle_float_focus"),
+  key({ mod, "Shift" }, "r", "way_cooler_restart")
 
   -- Quitting way-cooler is hardcoded to Alt+Shift+Esc.
   -- This my be modifiable in the future
--- a/config/init.lua
+++ b/config/init.lua
@@ -85,7 +85,7 @@ local keys = {
   key({ mod }, "e", "horizontal_vertical_switch"),
   key({ mod, "Shift" }, "q", "close_window"),
   key({ mod, "Shift" }, "space", "toggle_float_active"),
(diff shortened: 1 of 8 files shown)


Timidger:
@SnirkImmington have you finished the changes you wanted to make to this PR?

src/lua/thread.rs
```diff
@@ -155,30 +154,33 @@ fn main_loop(receiver: Receiver<LuaMessage>, lua: &mut Lua) {
...
+                    return
+                }
             }
         }
     }
 }
 
 /// Handle each LuaQuery option sent to the thread
-fn handle_message(request: LuaMessage, lua: &mut Lua) {
+fn handle_message(request: LuaMessage, lua: &mut Lua) -> bool {
```
Timidger: Should this maybe be a `Result`? The bool here makes it a little implicit that `false` here means we are supposed to restart, and isn`t an error from handling the message

Timidger:
I'll test this later tonight

lib/lua/lua_init.lua
```diff
@@ -65,3 +65,59 @@ config.register_key = function(key)
...
+    __rust.restart_callback = callback
+end
+-- Register a function to execute on terminate
+way_cooler_table.on_terminate = function(callback)
+    assert(callback, "missing callback")
+    assert(type(callback) == 'function', "callback: expected function")
+    __rust.on_terminate = callback
+end
+-- This could technically be called by clients if they want, it should be more hidden.
+way_cooler_table.handle_termination = function()
```
SnirkImmington: Clients can't kill the wm with this function, they can only call the Lua callback that does cleanup - the callback itself should be safe to call, at most it would just be annoying for any Lua code the user set up, and would probably cause the Lua thread to start erroring if other functions were called.

For example, if a Lua init file were to open a file, register a keybinding to read/write to the file, and in the `terminate`/`restart` close the file, then maliciously calling `way_cooler.handle_restart()` would close their open file and each keypress would probably error. 

The biggest concern we have with Lua right now is the various `dmenu-lua-do-things` commands - those still allow any program to register any Lua code to handle any keybinding. I don't think having this exposed is a good idea at all, but we should wait for a later PR because even if we did sandbox this a little we have bigger concerns.

config/init.lua
```diff
@@ -104,6 +105,10 @@ for _, key in pairs(keys) do
     config.register_key(key)
 end
 
+function cleanup_background()
+  os.execute("pkill way-cooler-bg")
+end
+
 
 -- Execute some code after Way Cooler is finished initializing
 function way_cooler_init()
```
SnirkImmington: I still don't understand why this is here. Is this supposed to happen post-wlc-startup? I think it's definitely useful to have something like that in place because there are definitely race conditions as we load the init file and wlc concurrently, so if users want to i.e. open a program on startup they might have to wait until after the compositor environment exists or the program would crash. But in order to do this we have to send the Lua thread some kind of `COMPOSITOR_READY` message in the wlc callback. In that case, we should have a `way_cooler.on_compositor_is_ready` callback that the Lua session calls explicitly when it has received a message to do so. That might be better off in a different PR, especially if we want `compositor_terminating` to match.

Timidger: Lua starts up [after the compositor has initialized](https://github.com/Immington-Industries/way-cooler/blob/master/src/callbacks.rs#L329). Than init function [then calls](https://github.com/Immington-Industries/way-cooler/blob/master/src/lua/thread.rs#L126) this function (`way_cooler_init`). There's no need to send a `COMPOSITOR_READY` message, as the compositor is already ready.

lib/lua/lua_init.lua
Timidger: Ok, this needs to be locked down but at least it's not as bad as I expected. Make sure you keep a list of what needs to be locked down as we move forward, so we don't miss anything.

config/init.lua
SnirkImmington: Okay, I totally missed that. Those changes were added in another PR, I'm gonna leave that alone. If we want to look at fixing this later I'll open another PR.

Timidger: It was probably added in the same PR as when I split up the background program, because that was the point we needed `way_cooler_init`.

Timidger:
Seems to still work, which is good. One issue though this error appears when restarting:

```
Lua termination callback returned an error: ExecutionError("[string \"chunk\":82: attempt to index global \'__rust\'  (a nile value)")
```

It _seems_ to work fine despite that error, but it should still be silenced

PR #168, opened 2016-12-10
Add static linking to wlc, Fixes #157

### LINKED ISSUE #157: Statically link with wlc
Right now the biggest barrier to trying out way cooler is having to install wlc, which can be a huge pain on non Arch/Fedora distros (which already have it in their repos).

As well, there have been stability issues in the past due to users having wrong versions of wlc. wlc is very unstable, and so having it statically linked would remove the possibility of a version mismatch.

Timidger:
Updates Way Cooler to use the latest version of rust-wlc (0.5.3), which includes static linking to wlc as a feature.

The feature is not enabled by default in Way Cooler, and should only be enabled for distribution purposes.


Timidger:
Haha don't know how that increased coverage but whatever 👍

PR #170, opened 2016-12-12
Updated way-cooler-bg to not specify cursor

Timidger:
Latest master of way-cooler-bg no longer requires the cursor, it is compiled in (adds about a megabyte to the binary).

This is a backwards compatible change, but I might as well change the binary call in the config file to not include the old (unnecessary) default argument.

### DIFF
--- a/config/init.lua
+++ b/config/init.lua
@@ -111,7 +111,7 @@ function way_cooler_init()
   if not status then
     print "Could not find way-cooler-bg! Please install it"
   else
-    os.execute("way-cooler-bg " ..  way_cooler.background .. " ../way-cooler-bg/assets/arrow.png &")
+    os.execute("way-cooler-bg " ..  way_cooler.background)
   end
 end
 
--- a/config/init.lua
+++ b/config/init.lua
@@ -111,7 +111,7 @@ function way_cooler_init()
   if not status then
     print "Could not find way-cooler-bg! Please install it"
   else
-    os.execute("way-cooler-bg " ..  way_cooler.background)
+    os.execute("way-cooler-bg " ..  way_cooler.background .. " &")
   end
 end
 


valryne:
I had problems with this change.

I Installed the current way-cooler-bg and specified an image in my init.lua (the quotes were necessary)
`way_cooler.background = "~/Pictures/My_Picture.jpg"`

Now, with your change, way-cooler freezes – no reaction to mouse or keyboard, such that I have to restart the computer.

When I add an ampersand to the relevant line, it correctly displays the background and hands back keyboard and mouse control.
`os.execute("way-cooler-bg " .. way_cooler.background .. "&")`

Now there is the problem, that the mouse cursor lives behind the background image and is only visible when above an opened container, or – when the background image is smaller than the screen – at the parts of the screen that are not covered by the image.

Timidger:
Ahh, whoops I forgot to put it in the background! Nice check, I should have tested the default config file before pushing this.

As for the mouse cursor living behind the background image -- that's strange, it sounds like the cursor image isn't being saved properly...Are you sure you're using the latest master of way-cooler-bg? I can test it later today and double check I didn't do anything stupid on way-cooler-bg's side.

valryne:
I just removed all traces of way-cooler-bg, git cloned it again, looked at main.rs to verify that latest changes are applied and ran cargo install – issue with cursor remains.

Timidger:
Ok, will investigate tonight

valryne:
Sorry – I'm a moron:
There was a hard link (should have been a soft link) to way-cooler-bg in my path that interfered.
Corrected that and now it works.

Timidger:
Haha oh good, glad to hear it's resolved 👍

Timidger:
I really should version way-cooler-bg...I'll throw it on crates.io later and actually have it behind a version so these breaking changes won't be as rough.

PR #182, opened 2016-12-25
Added distribution scripts

Timidger:
* Added `distribute.sh`, sets up a gzip for distributing way cooler. It is very platform specific, because it requires the machine to have a static version of the wlc library, and way-cooler-bg on the machine as well. It zips up:
  - `way-cooler`
  - `way-cooler-bg` 
  - `install.sh`
* Added `install.sh`, script to install Way Cooler on user's machine.
  - Gives error text if ran without root, but gives an option for recourse if that's what they _really_ want to do.

### DIFF
--- a/distribute.sh
+++ b/distribute.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+set -euo pipefail
+IFS=$'\n\t'
+mkdir dist
+echo "Building Way Cooler..."
+cargo build --release --features static-wlc
+cp target/release/way-cooler dist
+echo "Building Way Cooler Background..."
+(cd ../way-cooler-bg;
+ cargo build --release
+ cp target/release/way-cooler-bg ../way-cooler/dist)
+echo "Bundling install script..."
+cp install.sh dist
+echo "Zipping folder"
+mv dist way-cooler
+tar -cvzf way-cooler.gz way-cooler
+echo "Cleaning up..."
+rm -r way-cooler
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+set -eo pipefail
+IFS=$'\n\t'
+
+function assert_file_exists() {
+    if ! [ -f "$1" ]; then
+        echo -e "\e[31mCould not find file \"$1\", halting...\e[0m"
+        exit 1
+    fi
+}
+
+assert_file_exists "way-cooler"
+assert_file_exists "way-cooler-bg"
+
+if ! [[ $(id -u) = 0 ]] && ! [[ $# == 1 ]]; then
+    echo -e "\e[31m"
+    echo "The install script should be ran as root!"
+    echo -e "\e[0m \e[93m"
+    echo "If you *really* want to install Way Cooler as a user, please pass an install path to this script, like so:"
+    echo -e "\e[0m"
+    echo "./install.sh ~/bin"
+    echo -e "\e[31m"
+    echo "It it highly discouraged to install Way Cooler as your regular user, as it makes it much less secure!"
+    echo -e "\e[0m"
+    exit 1
+fi
+
+install_path=$1;
+: ${install_path:='/usr/bin'}
+[ -d $install_path ] || mkdir $install_path
+
+cp way-cooler $install_path
+cp way-cooler-bg $install_path
+
+chown $USER $install_path/way-cooler
+chgrp $USER $install_path/way-cooler
+chown $USER $install_path/way-cooler-bg
+chgrp $USER $install_path/way-cooler-bg
+
+if ! [[ $(pidof systemd) ]] && [[ $(id -u) = 0 ]]; then
+    echo "systemd is not installed on this machine, activating the setuid bit on $install_path/way-cooler"
+    chmod u+s $install_path/way-cooler
+fi
+
+echo -e "\e[32mWay Cooler has been installed on your system\e[0m"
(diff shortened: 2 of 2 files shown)


install.sh
```diff
@@ -0,0 +1,45 @@
...
+        exit 1
+    fi
+}
+
+assert_file_exists "way-cooler"
+assert_file_exists "way-cooler-bg"
+
+if ! [[ $(id -u) = 0 ]] && ! [[ $# == 1 ]]; then
+    echo -e "\e[31m"
+    echo "The install script should be ran as root!"
```
SnirkImmington: nit: `run`

install.sh
```diff
@@ -0,0 +1,45 @@
...
+    echo -e "\e[0m"
+    echo "./install.sh ~/bin"
+    echo -e "\e[31m"
+    echo "It it highly discouraged to install Way Cooler as your regular user, as it makes it much less secure!"
+    echo -e "\e[0m"
+    exit 1
+fi
+
+install_path=$1;
+: ${install_path:='/usr/bin'}
```
SnirkImmington: This is great syntax right here

distribute.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+set -euo pipefail
+IFS=$'\n\t'
+mkdir dist
+echo "Building Way Cooler..."
```
SnirkImmington: Should we integrate this into `ci.py` and post the zip files to GitHub releases? :+1: for separate file, so users can run without touching our Travis stuff.

- Alternatively we could extend the script to `git clone` the repo first, to get one of those `wget | sh` scripts...
- Should we specify "you'll need `cargo` and `rustc` to build way-cooler"

install.sh
Timidger: The confusing system of English strikes again! (or rather, my poor memorization of grammar rules)

install.sh
Timidger: We have the best syntax. The very best.

distribute.sh
Timidger: This script should not be ran by users (except for those that want to emulate/test our build system for distribution). This should only be ran by us, I was planning for it to be integrated into the build system kicked off by `ci.py`, however I didn't want to go through the trouble of setting up travis to build this all correctly (it took me long enough to get working on my local machine...). This is something I will correct in the future, when the pain of building the releases manually on my machine becomes too much.

As for your first bullet point, again here's the general layout of how I envisioned the distribution system:

0) `distribute.sh`, ran only by us
1) `curl https://way-cooler.org/install-script.sh | sh`, not built yet and _optional_. This script will be downloaded to the user's machine, it will do another curl for the most up-to-date version packaged by `distribute.sh`, and then run the second stage.
2) `install.sh`, ran by users, either manually or through stage 1.

SnirkImmington: I was thinking it would be convenient to set the build system up to be triggerable for testing purposes, i.e. it would create GitHub releases just to confirm things work. We could either have it mark them (I was thinking we could call it "prereleases" but that term is overloaded) or just delete them after they're created. This would be a lot easier if the github releases part of the code wasn't just magically done by Travis.

Timidger: I'm not sure I follow, what part of this workflow isn't testable? I was expecting the distribute system to be tested, even on non-release branches. All that needs to be done to test locally (or on the travis server) is to setup the environment to compile wlc statically and then you can test it works with `distribute.sh`. If you want to test installing (which would probably be better done on the travis side, to not mess with personal configuration), then you can do that with `install.sh`. 

If you want to test actually uploading these as github releases, I think you are taking the testing a step too far (we're testing our stuff works, not github's ;P). If you _want_ you can extend these scripts to update the github releases, but right now that seems out of the scopes of these two scripts (and would be bettter suited to putting in a third script)

SnirkImmington: > If you want to test actually uploading these as github releases, I think you are taking the testing a step too far (we're testing our stuff works, not github's ;P). If you want you can extend these scripts to update the github releases, but right now that seems out of the scopes of these two scripts (and would be bettter suited to putting in a third script)

This was the original idea of `ci.py` - provide cross-compiled binary releases automatically. Now it seems that providing GitHub distributions isn't _as_ important if the desired installation is piping a build-from-source script into `sh`.
One thing to note is that this is a "build from source" setup, and the idea was `ci.py` could handle cross-compilation to several different binary targets. Given that we don't have a lot of users, we have a large overlap with the Rust community right now, and we're still in alpha we're probably fine with this distribution, but the goal was to cross-compile and statically link everything via Travis so we could automatically provide binary downloads and distance ourselves from hard Rust/cargo dependencies.

Timidger: Not sure how I can make myself more clear, let me try again...

The user **is not building Way Cooler from source**
`distribute.sh` **is to be ran by us, we are building the binaries. It's automatically statically linked to wlc, the user doesn't even have to know this exists.**
We are offering **two** ways to install Way Cooler (excluding building from source):
1) The preferred way, they download tarred versions of Way Cooler. This includes:
  a) `way-cooler`
  b) `way-cooler-bg`
  c) `install.sh` <-- Which simply puts these binaries in their correct places
2) The least preferred way, but offered because I want to simplify the process for people to try out Way Cooler as much as humanly possible (and it will be noted as such) will be a `curl https://way-cooler.org/way-cooler-install.sh | sh`. This will run `way-cooler-install.sh`, which I **haven't written yet. It isn't one of these scripts**. 

We will not be offering the second option in the next release, only the first (and preferred) option.

PR #186, opened 2016-12-27
IPC debug enhancement

Timidger:
Small change to the `Debug` command on the ipc. Now returns `Uuid`s of all view/containers. This is pre-security considerations, and the `Debug` command was named as so as not to be relied upon. Later, this will be replaced with more specific commands to get the state of the tree (with varying levels of information) but for now I'm opening up the data to be displayed to make making client applications easier.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -52,15 +52,17 @@ impl ToJson for LayoutTree {
                     inner_map.insert(format!("Workspace {}", name), Json::Array(children));
                     return Json::Object(inner_map);
                 }
-                &Container::Container { ref layout, .. } => {
+                &Container::Container { ref layout, id, .. } => {
                     let mut inner_map = BTreeMap::new();
                     let children = tree.tree.children_of(node_ix).iter()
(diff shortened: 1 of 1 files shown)


SnirkImmington:
There are an alarming number of changes in this PR throughout the history, and the leaf nodes of the JSON are still just formatted strings, right? It's still got things like
```
format!("Container w/ layout {:?} and id {:?}", layout, id)
```
I might be missing something, but shouldn't that actually be a JSON object?
```
{
    "type": "container",
    "layout": "horizontal",
    "id": "<id>",
    "children": [ ... ]
}
```
Obviously this whole API still isn't done but if you wanted to enhance it that might be a better way to go. I can open a new PR with those proposed changes.

Timidger:
There we go, fixed the history.

This is just for my benifit, I realize the JSON is awful and totally unparsable. `Debug` was always meant to just be that -- debugging. There's no other way to get UUIDs of containers (which basically all other commands need) so i just wrote this so I can get the UUIDs of containers easily (again, this command will go away later and be replaced by others)

PR #192, opened 2017-01-01
let kill-way-cooler command be rebindable

### LINKED ISSUE #172: Let the "Kill Way Cooler" command be reboundable

thefarwind:
#### Summary of Changes

This was resolved by adding a set to track the command names (as strings). when keys::init is called, it checks to see whether the "quit" command has been bound, and if so, does not bind it.

Part of this involved changing when keys::init was called, so that it would occur after lua::init was called. That way it could be determined whether alt-shift-esc needed to be bound or not. 

#### Testing

The changes were tested as follows. No software tests were added, since the meat of this functionality relied on how the software behaved on start up.

1. Ran way-cooler with default init file.
2. Verified way-cooler closed with alt-shift-esc.
3. Ran way-cooler again.
4. Entered way-cooler restart command (alt-shift-r)
5. Verified way-cooler closed with alt-shift-esc.
6. Added 'alt-shift-d' quit command to init file and started way-cooler.
7. Verified 'alt-shift-esc' did not kill way-cooler.
8. Verified 'alt-shift-d' killed way-cooler.
9. Ran way-cooler again.
10. Entered way-cooler restart command (alt-shift-r)
11. Verified 'alt-shift-esc' did not kill way-cooler.
12. Verified 'alt-shift-d' killed way-cooler.

#### Alternative Solutions and Modifications

Alternative solutions would be to have the set be for the CommandFn, which could then be compared by `(&*cmd0 as *const _) == (&*cmd1 as *const _)`, or instead of having a set, loop through the KeyEvents to check for this specific message.

Currently the set_bound is separate from the register function, which means when binding a command the programmer needs to remember to call both functions. It would make sense to create a separate register function which took a function name as a string, looked up the command with the string, and then added the string to the bound_commands set.

#### Commit message:

added bound_commands set to keep track of command strings which have key bindings. Added set and is function to add items to to the bound commands set. Added commands set by the lua script to the bound commands. modified keys::init to check whether the quit command is bound, and if so, not bind alt-shift-esc to quit. Moved keys::init to be called after the completion of lua::init calls.

this should resolve #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ Way Cooler was started by @Timidger and @SnirkImmington, but these fine people h
 - @starfys created way-cooler desktop file
 - @toogley fixed a link
 - @paulmenzel fixed a typo
+- @thefarwind made kill way-cooler command rebindable
 
 And of course, thanks to the Rust community and the developers of [wlc].
 
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -327,6 +327,7 @@ pub extern fn compositor_ready() {
     info!("Preparing compositor!");
     info!("Initializing Lua...");
     lua::init();
+    keys::init();
 }
 
 pub extern fn compositor_terminating() {
--- a/src/commands/defaults.rs
+++ b/src/commands/defaults.rs
@@ -23,7 +23,7 @@ pub fn register_defaults() {
         coms.insert(name.to_string(), val);
     };
 
-    register("quit", Arc::new(quit));
+    register("way_cooler_quit", Arc::new(way_cooler_quit));
     register("launch_terminal", Arc::new(launch_terminal));
     register("launch_dmenu", Arc::new(launch_dmenu));
     register("print_pointer", Arc::new(print_pointer));
@@ -138,7 +138,7 @@ fn print_pointer() {
         .expect("Error telling Lua to get pointer coords");
 }
 
-fn quit() {
+fn way_cooler_quit() {
     info!("Closing way cooler!!");
     ::rustwlc::terminate();
 }
--- a/src/keys/mod.rs
+++ b/src/keys/mod.rs
@@ -46,19 +46,18 @@ pub fn keymod_from_names(keys: &[&str]) -> Result<KeyMod, String> {
 pub fn init() {
     use rustwlc::xkb::keysyms;
     use commands;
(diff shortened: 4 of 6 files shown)


Timidger:
Oh also I forgot -- feel free to add yourself to the [list of contributors](https://github.com/Immington-Industries/way-cooler#contributors) in this PR as well.

thefarwind:
I've modified the name to way_cooler_quit, and redid the tests above to check that it works. I've also added myself to the list of contributors!

I haven't yet added in the change to a bool to determine whether the command has been bound. I'll wait for @SnirkImmington to weigh in before adding the change into the pull request.

SnirkImmington:
I actually kind of like the `HashSet` here, having the functionality to check for unbound keybindings could be helpful. `way_cooler_quit` is probably the only command we want to ensure is bound, but would there be benefits to giving the Lua client some kind of `way_cooler.check_keybinding(cmd)`? I can see it being used to fill in unbound keys or commands, for example if we gave client programs the ability to register their own keybindings. This is maybe not something that needs to be done in a window manager...
Would we in the future want to display a warning on other keybindings not being registered?
If there aren't any ideas, I say just make it a `bool` - we probably don't need to keep track of any other keybindings.

thefarwind:
On another note, I just ran into an issue with this fix using the hashset to track keybindings -- if the quit keybinding is overwritten after being put in the init file, the alt-shift-esc keybinding will not be made.

Fixing this would make special casing the quit command lighter than a general solution like the hashset provides.

SnirkImmington:
I'm gonna go with :+1: for special-casing the quit command - if we have use for "check if command is registered" later we can look into it.

thefarwind:
Okay, that's what I get for trying to be clever. The HashSet has been ripped out, and the is_quit_bound code now checks by cycling through the values in the hashmap to determine whether the quit function has been bound.

Timidger:
Ok cool, looks good to me. 

Also a side note -- generally it's considered bad practice (at least imho) to rewrite public git history in a PR. It makes it difficult to see the progression of changes that happened (in case there needs to be a rollback, for example) and also makes it more difficult to test (instead of just pulling your changes, I need to fetch and reset onto your new base). 

Thanks again for your contribution.

thefarwind:
Okay, thanks for the advice. I'll keep that in mind for next time.

PR #209, opened 2017-01-07
Added disable-debug feature

### LINKED ISSUE #202: Enable all debug checks for < 1.x versions
Since we aren't 1.0, it is expected to break often. The debug checks that are currently used in the layout code should be enabled for release builds as well.

The simplest way to do that is with a compilation flag / feature, eg `enable_debug_checks` or something to that effect. Then when 1.0 hits we can turn that off when compiling release distributions.

Timidger:
* Added `disable-debug` feature. 
  + Disables debug checks in Way Cooler. Not recommended, though it will improve performance.
  + When we reach 1.0, this feature will be replaced with a `enable-debug`, which will do the opposite (enable debug checks with present)

Fixes #202

### DIFF
--- a/src/layout/actions/workspace.rs
+++ b/src/layout/actions/workspace.rs
@@ -3,6 +3,7 @@ use uuid::Uuid;
 use rustwlc::{Geometry, Point};
 use super::super::LayoutTree;
 use super::super::core::container::{Container, ContainerType};
+use ::debug_enabled;
 
 // TODO This module needs to be updated like the other modules...
 // Need to add some errors for this (such as when trying to move a non-container/view,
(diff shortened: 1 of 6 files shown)


src/layout/actions/workspace.rs
```diff
@@ -182,7 +182,7 @@ impl LayoutTree {
 
             // Get the root container of the next workspace
             let next_work_children = self.tree.children_of(next_work_ix);
-            if cfg!(debug_assertions) {
+            if cfg!(debug_assertions) || !cfg!(disable_debug) {
```
SnirkImmington: Do you want to turn this into a macro/inlined function? i.e. `debug_enabled!()`

Timidger: Good call, I was dotting it all over the place and the logic can be difficult to read if there are a bunch of `cfg!`s clogging it up.

Timidger: Also would mean we only have to change it in one place once we switch the flag.

Issue #160, opened 2016-12-05
Add fullscreen command


Timidger:
New command `fullscreen` should make a view/container be the width and height of the screen. It should cover all other views (except ones that steal focus, and thus need attention in order to not lock the user out), and not be resizeable/movable. The keybinding will be `<mod>+f`, and it will toggle fullscreen (so if used while an app is fullscreen, it will make it normal sized again).

PR #213, opened 2017-01-07
Merged in petgraph+stablegraph

### LINKED ISSUE #207: Update Petgraph
A patch was sent to us by @bluss [here](https://github.com/Immington-Industries/way-cooler/pull/114), but ultimately it was not merged in. This fix is critical to fixing an edge case where Way Cooler can panic if these specific criteria are met:

* There's a fullscreen view in a workspace
* A non-active-container is removed (either by user or IPC)

To fix this, the `NodeIndex` invalidation needs to no longer be a thing. 

I will regrab his patch, update it to fix the many conflicts that have probably spawned since then, and also fix the tests (which was the reason it wasn't merged in to begin with)

Timidger:
Fixes #207.

Needs a lot of manual testing to ensure nothing broke. Did a few simple tests and it seems fine but I'll be running this exclusively over the next couple of days to see if there is any breakage. 

Unblocks #160.

Note that somehow the commits from @bluss got dropped (he basically wrote all of it, I just fixed tests and updated it for 0.4). He has never the less been added to the contributors list for being so kind as to help update the project.


Timidger:
Looks like this PR breaks a big invarient: That the active container always points to a valid node (breaks eg fullscreen very badly).

Going to fix this and then it should be good to go (barring more show stoppers)

bluss:
Great. Do let me know if there are any unexpected issues with StableGraph, or some API from Graph that it is missing for you.

Timidger:
Will do, thanks again for your patch.

PR #216, opened 2017-01-10
Replaced old, bad video with new, good image.

Timidger:
The video was misleading, and out of date. It has been placed with an up-to-date image of what Way Cooler looks like today.


SnirkImmington:
I think we'll want to revisit this with lemonbar.

PR #230, opened 2017-01-13
Added debug symbols to release build

Timidger:


SnirkImmington:
Just use `rustc -C debuginfo=2` when compiling. No need to change `Cargo.toml`.

Timidger:
I'd prefer to not have to remember that everytime I update my version to the latest master.

PR #231, opened 2017-01-13
Added gap support

### LINKED ISSUE #206: Add gap support

Timidger:
* Can control gap width with `way_cooler.gap_width` (default 0)
  + Negative numbers are considered 0
* Gaps are separate from borders (which are currently unimplemented). It's possible to have both borders and gaps.

Fixes #206

### DIFF
--- a/config/init.lua
+++ b/config/init.lua
@@ -54,9 +54,7 @@ local key = config.key -- Alias key so it's faster to type
 
 way_cooler.terminal = "weston-terminal" -- Use the terminal of your choice
 
--- The width of gaps between windows in pixels
--- Setting it to anything <= 0 puts no gaps between windows.
-way_cooler.gap_size = 0
+way_cooler.gap_size = 0 -- The width of gaps between windows in pixels
 
 local keys = {
   -- Open dmenu
--- a/config/init.lua
+++ b/config/init.lua
@@ -54,6 +54,10 @@ local key = config.key -- Alias key so it's faster to type
 
 way_cooler.terminal = "weston-terminal" -- Use the terminal of your choice
 
+-- The width of gaps between windows in pixels
+-- Setting it to anything <= 0 puts no gaps between windows.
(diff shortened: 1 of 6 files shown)


config/init.lua
```diff
@@ -54,6 +54,10 @@ local key = config.key -- Alias key so it's faster to type
 
 way_cooler.terminal = "weston-terminal" -- Use the terminal of your choice
 
+-- The width of gaps between windows in pixels
+-- Setting it to anything <= 0 puts no gaps between windows.
```
SnirkImmington: I'd remove this comment from the sample config, it's nice to know from an API point of view that we ignore bad input but unless the user was hoping to have all their windows overlap I don't think they care.

src/layout/actions/layout.rs
```diff
@@ -617,6 +623,68 @@ impl LayoutTree {
...
+        };
+        let gap = registry::get_data("gap_size")
+            .map(RegistryGetData::resolve).and_then(|(_, data)| {
+                Ok(data.as_f64().map(|num| {
+                    if num <= 0.0 {
+                        0u32
+                    } else {
+                        num as u32
+                    }
+                }).unwrap_or(0u32))
```
SnirkImmington: You mention that if the specified gap is less than zero it will be ignored but you never mentioned that if it isn't a number it's also ignored.

config/init.lua
Timidger: Good point, I'll remove the second line.

src/layout/actions/layout.rs
Timidger: Yeah, I wasn't being clear there. I think if we need to document that behaviour any more it should probably live in a different place than the configuration file (some sort of proper docs on a site, for example). So I won't add that information, I'll do it later in a big doc sweep once we have more features to document (and a better place than the configuration file to document them)

PR #232, opened 2017-01-14
Floating windows choose where they display

### LINKED ISSUE #218: Popup windows should be centered on the screen
Currently, we just float the windows and don't do anything with their positioning. However, by default this puts them in the top left corner (0, 0). Instead, they should be in the exact center of the screen.

* Grab the resolution of the screen
* Find the mid point
* Subtract by width and height of the view
* ???
* Profit

Timidger:
* Floating windows can decide where they display (and Way Cooler will honor the request).
* Internal: Tree commands to get a container now returns a `Result<&/&mut Container, TreeError` for better error checking.

Doesn't totally fix #218, as we don't do any centering. Still considering whether that's worth implementing.


src/callbacks.rs
```diff
@@ -360,7 +367,7 @@ pub extern fn pointer_motion(view: WlcView, _time: u32, point: &Point) -> bool {
                 if let Ok(mut tree) = try_lock_tree() {
                     // TODO Change to id of _view
                     // Need to implement a map of view to uuid first though...
```
SnirkImmington: Nit: comments are out of date

src/layout/commands.rs
```diff
@@ -514,5 +514,18 @@ impl Tree {
             self.0.layout_active_of(ContainerType::Output);
             Ok(())
         })
+   }
+
+    /// Updates the geometry of the view from an external request
+    /// (such a request can come from the view itself)
+    ///
+    /// This is only applied to the view if it is floating.
```
SnirkImmington: Nit: another old comment

PR #234, opened 2017-01-19
Background now uses actual output geometry

Timidger:
Background will now ignore the top bar and correctly stretch to the entire length of the screen.

### DIFF
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -31,13 +31,19 @@ impl LayoutTree {
                 }
             }
             ContainerType::Output => {
-                let geometry = self.tree[node_ix].get_geometry()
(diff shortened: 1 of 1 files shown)


SnirkImmington:
Will the top bar cut off the top of the background?

Timidger:
Making this configurable is the way to go, I'll look into it later. Shouldn't be hard to add to this PR.

Timidger:
Hmm, as I think about this, I still think this should be an option but an option that the background can choose. 

The current background program is very very basic, but eventually it should be possible for it to dictate where it should be drawn (and be able to query to see the geometry of the bar, and it can decide what to do with that information).

For now, I'll close this PR so I can be reminded to make this change with how the background is drawn.

PR #239, opened 2017-01-22
D-Bus commands now block locking the layout tree

Timidger:
Shaking your mouse violently should no longer make your top bar suddenly stop displaying the workspace listing ;)


Timidger:
Will be fixed when borders is merged in (it has the correct method)

PR #249, opened 2017-02-16
Cleaned up init/configuration file

### LINKED ISSUE #174: Remove old Registry

### LINKED ISSUE #246: Fix bad user config error in log
This was an issues in debugging #245. Eventually we should display a notification to the user, but at the bare minimum the log file should output the correct error about the user config.

Timidger:
# Configuration
* Changed the format of the init file significantly
  + See the new default for layout
* Added `spawn_*` functions, to aid in registering programs to run at startup and every restart.
* An error is now logged if Way Cooler has to default to the pre-compiled configuration file
# Registry
Redid internal structure of internal registry
* Each value must be part of some category
* Clients to the registry can only access certain categories
  + They have permissions to read and write at a category level
  + Has a "nil" client that can read and write anything.
    * By default, internal Way Cooler use and the Lua thread uses this client.

Fixes #174 
Fixes #246


SageEx:
Hi again, nice to see this.
My initial hack to bind keybindings to programs is failing.

I defined
```lua
function launch_program(prog_name)
    os.execute(prog_name .. " &")
end
```

and used to use it like
```lua
  key({ mod }, "a", function () launch_program("st -e ranger") end),
```

This does not work anymore.

Also, it would be nice to provide a default implementation of such a ``launch_program_once`` function which the user can call directly.

lib/lua/utils.lua
```diff
@@ -136,4 +136,39 @@ function util.table.get_random(tab)
...
+function util.program.spawn_startup_programs()
+  for index, program in ipairs(util.program.programs) do
+    os.execute(program.bin .. " " .. program.args .. " &")
+  end
+end
+
+-- Stops the startup programs. Does not remove them from the global list.
+function util.program.terminate_startup_programs()
+  for index, program in ipairs(util.program.programs) do
+    -- TODO Store PIDs and kill those instead
```
SageEx: Storing the pIDs and killing them later is not a good idea, since the program may have been restarted by the user themselves.

Timidger: Ah, good point I didn't consider that problem...hmm, any suggestions on how to tackle it properly then?

Timidger:
> Hi again, nice to see this.
> My initial hack to bind keybindings to programs is failing.
> 
> I defined
> 
> function launch_program(prog_name)
>     os.execute(prog_name .. " &")
> end
> 
> and used to use it like
> 
>   key({ mod }, "a", function () launch_program("st -e ranger") end),
> 
> This does not work anymore.

I forgot to test Lua functions, ahh. I'll fix this ASAP.

> Also, it would be nice to provide a default implementation of such a launch_program_once function which the user can call directly.

I'll add that as `utils.program.spawn_once`. I'll hold off on general documentation of these `utils` until I'm sure this is the API I want, but until then I'll update the README so that users are generally aware that these exist.

Timidger:
Hmm, from my testing lua functions still work...why don't you send me your updated config and I can see where the error is?

Timidger:
There we go, that issue you reported should be fixed.

I have also added `spawn_once` to the `util`s. It has the same arguments as the other `spawn_*` functions.

lib/lua/utils.lua
SageEx: I thought about it a bit.
Everything specifically related to way-cooler (like way-cooler-bg and way-cooler-bar) should be *pkilled*.

Then close each window in all the workspaces, this helps since the cmd-line programs won't be killed.

This should be enough hopefully, rest of the cleanup can be done in the shutdown function.

SageEx:
Thanks, this fixed the function call problem.

How to bind the new ``spawn_once`` function to keybindings ?

Timidger:
Just wrap it in a function, like you did before.

E.g: ` function () spawn_once("st -e ranger") end`.

If you want to make it more concise, try having a helper that returns a closure:


```
function() spawn_wrapper(bin, args)
    return function() utils.program.spawn_once(bin, args) end
end

...
key({mod}, "a", spawn_wrapper("st -e ranger"))

```

That should work.

SageEx:
Aha ! got it working, thanks for the closure way.

Note: The error was calling ``utils.program.spawn_once`` while it should be ``util.program.spawn_once``

Timidger:
Hah, whoops. Maybe it should be `utils`...eh, I've made enough breaking changes for one PR ;).

I'm going to try to fix kconv so that I can see coverage changes with this PR, and then I'll merge this into master.

lib/lua/utils.lua
Timidger: Currently the way this is implemented, any program launched from `spawn_at_startup` is killed in this way (`pkill` currently).

When Way Cooler dies, everything else is killed normally (no `pkill` nonsense). For now, I'm going to keep it in its current form.

The main reason I wanted to have a more fine-grained way to fix this was testing Way Cooler while running Way Cooler would kill my host background/bar program as well as the instance I was testing :(. This is a very small issue though, so I should probably just live with it ;)

PR #262, opened 2017-02-27
Improved last active container algorithm

### LINKED ISSUE #204: Active path should be tracked with numbers, not a flag
If we change the active path to a `u32`, instead of a simple `bool` flag, then it's possible for Way Cooler to remember the active path between siblings in the nodes of the tree.

In practical terms, this means that previous paths will be remembered within containers (e.g like it does in i3, but not Sway) and that it would be possible to remember the previous used workspace (e.g like the command in awesome)

Timidger:
The last active container algorithm should now behave exactly as i3 does (e.g, unlike Sway it remembers what to focus on within in a container's group of siblings).

Still needs to be tested, but preliminary testing seems to imply it works.

Fixes #204

### DIFF
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -146,6 +146,9 @@ impl LayoutTree {
     /// tree until either a view is found or the workspace is (in which case
     /// it set the active container to the root container of the workspace)
     ///
+    /// If there are siblings, the chosen node is the one with the lowest
+    /// active count.
+    ///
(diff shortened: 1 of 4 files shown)


Timidger:
The active count numbers, while guaranteed to be unique, are not restricted by how many siblings there are. This means _very_ long lived layouts might eventually cause a panic when the `u32` overflows. This will be fixed before this is merged in, as well as tests to verify that this feature works.

PR #279, opened 2017-03-26
Added lock screen support

Timidger:
This PR adds support for lock screen client programs :tada:

Lock screen programs, and their respective keybinding to bind them, are specified in the Lua configuration file, under `programs`: 
```  
lock_screen = "/path/to/lock-screen-program"

-- lock screen
key({ mod, "Shift" }, "l", "lock_screen")

```

When the lock screen is active, all other commands and inputs are disabled. You cannot switch workspaces, spawn other programs, or otherwise do anything else.

D-Bus commands are **NOT** checked by this, as right now this is not a problem. This may change in the future, and appropriate errors will be returned (e.g "Way Cooler is locked"). Since the API for the D-Bus layout commands have not stabilized yet, this feature has not yet been implemented until they are done.

Currently, there are no lock screens implemented, but the only requirement to make one is:
* A program that spawns a GUI of some kind
* Can be killed by the user without the use of a Way Cooler keybinding (otherwise it can't be unlocked).

Note that since *all* window creation is disabled, the lock screen cannot make a subsurface. This means e.g right click menus would not be possible yet in the lock screen. This may change in the future, but only if it imposes a significant burden for lock screen programs.


Timidger:
I'm going to go ahead and block this PR until

1) #287 is finished
2) A lock screen program is made so that this feature is actually helpful AND to ensure that it is complete.

Timidger:
Updated to use the new modes setting, just need to wait on [this Rust PR](https://github.com/rust-lang/rust/issues/38903) to merge and be pushed to a stable release before I can be sure that the lockscreen can't lock the user out of Way Cooler inadvertently.

Timidger:
Though I have problems with the current client implementation, I'm going to merge this into master since the Way Cooler side of things works well enough.

There will be a patch later that fixes the previously mentioned nightly guarded [try_wait](https://doc.rust-lang.org/std/process/struct.Child.html#method.try_wait) method on `std::process::Child`.

Problems with the client, just so readers don't have to go looking for it:
1) The most secure part of the application (interfacing with PAM) is written in C, because there is no good enough Rust library that suits this purpose.
2) The graphical display is not very helpful to the user.

PR #283, opened 2017-03-30
Fix panic when a container's origin.x/y is negative

hedning:
Hi, I've been playing around with way-cooler a bit which have been really straightforward and enjoyable. Haven't used rust before, but it seems great :)

When trying to get a big scrollable container working I found that if the container have negative coordinates way-cooler will crash due to casting a negative i32 to an u32. Inverting the cast fixes the issue.


Timidger:
Ah, nice catch. I was assuming containers would always be in the view port because I was tiling them...in retrospect I should have been more careful about that! Thanks for the patch!

PR #284, opened 2017-04-01
Possible layout feature: Scrollable tiling (don't merge)

hedning:
Hi, I've made a working mockup of what could be called scrollable tiling. Here's a concept picture:

scrollable-container

It's just a wide container which scrolls the active window into view.

My brother (@olejorgenb) and I have been using a [prototype in Notion](https://github.com/olejorgenb/notion/tree/paper-wm) for a while now, and would like to find a home for the concept in Wayland.

We often found ourselves using several workspaces for a single project since a workspace isn't all that big. A workspace grid where each row is a named group is a natural solution to this problem, but grids are hard to navigate and forces you to artificial screen sized groupings. Merging the row into a scrollable strip is a much cleaner solution with less limitations. This way you'll always have space for a new window without having to worry about how it affects the rest of the layout. We think it gives a lot of the freedom with floating windows, but with the structure of tiling.

To test out it just pull and cargo build. I've hardcoded the Scroll layout as default, so just spawn a bunch of terminals (hardcoded to 500px wide at the moment) and use `meta+left/right` to move around as usual.

So yeah, wondering if you're interested in having support for something like this in Way-cooler?

(Edit: there's a few rendering bugs and a bunch of failed tests, but should give you an idea of the concept)

### DIFF
--- a/src/callbacks.rs
+++ b/src/callbacks.rs
@@ -138,7 +138,7 @@ pub extern fn view_created(view: WlcView) -> bool {
     }
     if let Ok(mut tree) = lock_tree() {
         let result = tree.add_view(view).and_then(|_| {
-            view.set_state(VIEW_MAXIMIZED, true);
+            // view.set_state(VIEW_MAXIMIZED, true);
             match tree.set_active_view(view) {
                 // If blocked by fullscreen, we don't focus on purpose
                 Err(TreeError::Focus(FocusError::BlockedByFullscreen(_, _))) => Ok(()),
--- a/src/ipc/layout.rs
+++ b/src/ipc/layout.rs
@@ -54,6 +54,7 @@ dbus_interface! {
         // TODO Tree commands need to have these defined on the Tree,
         // for now this is _ok_, but we are swallowing an potential Tree lock error here.
         match axis {
+            Layout::Scroll => layout_cmd::split_horizontal(),
             Layout::Horizontal => layout_cmd::split_horizontal(),
             Layout::Vertical => layout_cmd::split_vertical()
         }
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -85,6 +85,8 @@ impl LayoutTree {
         match self.tree[parent_ix] {
             Container::Container { layout, .. } => {
                 match (layout, direction) {
+                    (Layout::Scroll, Direction::Left) |
+                    (Layout::Scroll, Direction::Right) |
                     (Layout::Horizontal, Direction::Left) |
                     (Layout::Horizontal, Direction::Right) |
                     (Layout::Vertical, Direction::Up) |
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -124,6 +124,59 @@ impl LayoutTree {
                     _ => unreachable!()
                 };
                 match layout {
+                    Layout::Scroll => {
+                        let children = self.tree.grounded_children(node_ix);
+                        let children_len = children.len();
+                        let width = LayoutTree::calculate_scale(children.iter().map(|child_ix| {
+                            let c_geometry = self.tree[*child_ix].get_geometry()
+                                .expect("Child had no geometry");
+                            c_geometry.size.w as f32
+                        }).collect(), geometry.size.w as f32);
+
+                        if width > 0.1 {
+
+                            let default_width = 500 as u32;
+                            // Updated geometry with enough width to hold all children
+                            let new_geometry = Geometry {
+                                origin: geometry.origin.clone(),
+                                size: Size {
+                                    w: default_width*children_len as u32,
+                                    h: geometry.size.h.clone()
+                                }
+                            };
+                            let new_size_f = |child_size: Size, sub_geometry: Geometry| {
+                                Size {
+                                    w: default_width,
+                                    h: sub_geometry.size.h
+                                }
+                            };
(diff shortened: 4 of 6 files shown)


Timidger:
Ooo, this is really cool. That picture really helps to understand the feature too :). Never thought about tiling windows like this, but this is certainly an interesting way to do it. Normally when I want windows like that, I use the tabbed/stacked layouts if I'm using i3, but of course I haven't implemented that in Way Cooler yet :P. This is quite a bit different though, since you're not restricted by the size of the screen.

I think this would be a great addition to Way Cooler, there's just a few questions that need to be resolved in the actual implementation.

1) Will the individual windows be re-sizable / have different sizes? Right now they are locked at their size, though I assume that was to make it easier to proto-type.
2) How will it fit with the rest of the tiling options? The other can be nested (e.g a horizontal container can be in a vertical one), and I'm not sure how that could be achieved with this. Since it's made to scroll outside of the visible view, you could do the same thing but have it hidden by the other containers around it (e.g they would be displayed on top of the "offscreen" ones). Or, we could restrict it to the root container only and still allow the others to be nested inside of it. That would be the easiest implementation, and probably the best since most of the benefits of this layout is that it's not necessarily restricted in size.
3) Another interesting feature on top of this would to be able to scroll through them using the scroll wheel on the mouse when hovering over their title bars. That should be fairly easy to add--the title bars aren't real views, so when a scroll event is detected on the root view (WlcView(0), there should be helper methods in rust-wlc to detect this) and the focused container is in a `Scroll` container, we could scroll along it just like when going left or right. If you are interested in that we can try to implement that after making this patch work and merging it into master.

From a look at the tests it looks like this just broke because you changed the default layout option--which hints I should make those tests more resilient to changes in the default setup :P. We should add some tests once we switch it back to the default to ensure this setup works as well.

In terms of configuration, the only thing that needs to be added looks to be a keybinding to activate this tiling option (e.g like for `Horizontal` and `Vertical`), that can be added at the end when it's closer to being done.

hedning:
Glad you like it, I'll try to get stuff cleaned up and root out the rendering bugs :)

1. Yeah, was just to prototype it quickly. The widths were huge when they got to layout_helper. 
The initial width should probably be the requested width of the window, perhaps restricted by some configurable maximum. It seems like layout_helper doesn't have access to this at the moment (normalize_views sets the width I think). Actions for maximizing/unmaximazing horizontal width should be implemented. "META+-/+" could decrease/increase the size.
2. Yeah, it doesn't make much sense having a horizontal Scroll inside a horizontal Scroll. Though it could work with a vertical Scroll, but I agree that it's probably best to just restrict it to root for the time being. There's a bug where half the screen goes grey if there's a vertical split hidden to the left (think it has something to do with the borders). A binding that does vertical tiling if the container is horizontal/scroll and horizontal if it's vertical might be nice.
3. Scrolling with wheel/touch would be awesome.

Some other stuff we have planned to get working eventually is a minimap of the workspace, smooth animated scrolling and expose, but yeah, that will probably take a while :)

Some implementation questions: 
1. I've added a new layout for this, but another option is to have a something like `container.scroll: bool` which modifies the horizontal/vertical layout, but it might be better to just have it more decoupled?
2. `scroll_into_view` is called from `set_active_node` which causes two re-layouts to be run when adding a new window, is there a better place to put this?

hedning:
Here's some pictures of the rendering bug with nested layouts:
2017-04-02-003345_889x733_scrot
To the left is this:
2017-04-02-003315_1017x727_scrot

Timidger:
1) I think it is best to keep it decoupled into a separate layout type, especially if it's restricted to the root container.
2) Hmm, you're right about the inefficiency there when adding a new view. That's something that's been bothering me for a bit, since `set_active_node` is really helpful in ensuring that everything is set properly for the active node (e.g the borders, the `active_container` index, etc.) but it does make it inefficient when there are other methods that guarantee parts of those invariants.

Instead, is there a way to see if the view is already visible in `scroll_into_view`? That would mean we could keep it in the same place but then it'll never do unnecessary work.

hedning:
I'll continue with the separate layout then :)

Hmm, yeah,  scroll_into_view only issues the re-layout if some part of the active container is outside the output resolution, and the layout on the workspace after a view is added is in general necessary. So not much one can do. Think it was a case of trying to pre-optimize on my part.

hedning:
So, sorta fixed the rendering bug. Seems like it's `VIEW_MAXIMIZED` that's causing problems, commenting it out makes things work. I'm guessing it's a wlc bug. But yeah, not really a full fix.

Timidger:
How's this coming along? What else is needed to make it merge-able/is it ready for a full review?

hedning:
Haven't worked on it much lately. We've started looking at the possibility of building a wayland compositor using Clutter and gjs (clutter have a lot of stuff built in, and it should be somewhat easy to integrate javascript), so time have gone there.

It basically works now, but not with `MAXIMIZED` views, which isn't ideal. What's left to do:
- Restrict Scroll layout to root container
- Some config option/keybinding to activate (switching from Scroll to another layout wouldn't work well though)
- Iron out some focus bugs

I'll probably see how the Clutter thing goes before getting it into a reviewable state (though it's functional enough to try out now).

Timidger:
I'm going to go ahead and close this, since there hasn't been much activity. If anyone would like to update this and resubmit in a PR, I'll be happy to review adding it to Way Cooler.

PR #293, opened 2017-04-17
Fixed link in README.md

nycatelos:
Made the link to the i3wm website in the description link properly.

Before it wouldn't link to the website, but attempt to link to a nonexistent page at https://github.com/way-cooler/way-cooler/blob/master/i3wm.org


Timidger:
Thanks!

PR #297, opened 2017-04-22
Multi head output

### LINKED ISSUE #64: Core Feature: multi-head output
This is a very sought after feature, and will need to be broken down into different tasks in order to achieve it.
- [x] Change workspace initialization to include initializing all outputs, similar to how i3 places a new workspace on each output, starting at 1, then 2, etc...
- [x] Add a new invariant to the tree: Each output needs to have at least one workspace. Add it to validation checks and add tests on all methods that might invalidate this (such as workspace pruning movement/pruning methods).
- [ ] ~Rewrite the container movement / focus movement methods to account for moving into another output at the edge of a workspace.~ Since they are individual buffers, we do not know where the output exists at. This feature has been punted. An explicit command to move to a workspace/output is used instead, like in i3
- [ ] Ensure we use the correct resolution setting functions from wlc.
- [x] Gracefully handle adding and removing outputs to the tree.

### LINKED ISSUE #235: Is multi-seat / multi-pointer / multi-focus supported or planned?
I'm looking for a tiling window manager for Wayland that supports multi-seat / multi-pointer / multi-focus features ([example](https://www.youtube.com/watch?v=WO2L_ihO_rI)). Are these features supported or planned for Way Cooler?

Mainly I am looking support for a concept of multi-focus where different sets of pointers and keyboards would have their own focus (perhaps visually color coded on the windows). Wayland itself would allow the features.

Timidger:
The long awaited feature that really should have been there at the beginning is here. Yes, Way Cooler can now be used with multiple screens! :tada:

When an output is connected, a new workspace is automatically created for it. To switch to that output, you switch to that workspace. To move a container to that output, you send it to a workspace on that output. To create more workspaces on that output, you switch to a workspace already on that output and then switch to the desired workspace.

There are a few limitations with the current design however:
* The new workspace created by the output takes on the name of the numerical value of the `WlcOutput` backing the output. This means:
  + It's possible for there to be a workspace collision, which will crash the wm :(
  + It assumes that the workspace is accessible, which is not the case if the workspace name is not bound (e.g if the number >= 10, or if the user chose to name their workspaces something other than "1", "2", "3"...)
* Backgrounds, X11 bars, and screen scrapping are not supported on multi monitors (yet)

These issues will be fixed in a later version, they are not fixed here because to fix this properly will require exposing some sort of configuration option to the user (e.g what workspaces to have), and also some sort of fallback option to do if all workspaces are taken (which will probably entail stealing one of them from the other outputs).

Any feedback on this feature is appreciated, including how to structure the configuration options for outputs/workspaces going forward.

Fixes #64

### DIFF
--- a/config/init.lua
+++ b/config/init.lua
@@ -105,6 +105,7 @@ local keys = {
   key({ mod, "Shift" }, "space", "toggle_float_active"),
   key({ mod }, "space", "toggle_float_focus"),
   key({ mod, "Shift" }, "r", "way_cooler_restart"),
+  key({ mod, "Shift" }, "s", "toggle_active_output"),
 
   -- Quitting way-cooler is hardcoded to Alt+Shift+Esc.
   -- If rebound, then this keybinding is cleared.
--- a/config/init.lua
+++ b/config/init.lua
@@ -105,7 +105,6 @@ local keys = {
   key({ mod, "Shift" }, "space", "toggle_float_active"),
   key({ mod }, "space", "toggle_float_focus"),
   key({ mod, "Shift" }, "r", "way_cooler_restart"),
-  key({ mod, "Shift" }, "s", "toggle_active_output"),
 
   -- Quitting way-cooler is hardcoded to Alt+Shift+Esc.
   -- If rebound, then this keybinding is cleared.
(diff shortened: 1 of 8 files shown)


hellracer:
May I ask if we have 3 to 4 monitor setup does still work how about xrandr support?

Timidger:
I haven't actually tested 3 + monitors but it should work.

xrandr will never work with any Wayland compositor

PR #305, opened 2017-06-03
Move lua init to before compositor init

kiran-kp:
- moved lua init to before compositor init
- defer lua on_init() trigger to after compositor init

### DIFF
--- a/src/lua/thread.rs
+++ b/src/lua/thread.rs
@@ -161,7 +161,7 @@ pub fn init() {
     // Immediately update all the values that the init file set
     send(LuaQuery::UpdateRegistryFromCache)
         .expect("Could not update registry from cache");
-    
+
     // Re-tile the layout tree, to make any changes appear immediantly.
     if let Ok(mut tree) = lock_tree() {
         tree.layout_active_of(ContainerType::Root)
--- a/src/lua/thread.rs
+++ b/src/lua/thread.rs
@@ -152,10 +152,23 @@ pub fn init() {
         info!("Skipping config search");
     }
 
+    // Only ready after loading libs
+    *RUNNING.write().expect(ERR_LOCK_RUNNING) = true;
+    debug!("Entering main loop...");
+    let _lua_handle = thread::Builder::new()
+        .name("Lua thread".to_string())
+        .spawn(move || { main_loop(receiver, &mut lua) });
+    // Immediately update all the values that the init file set
+    send(LuaQuery::UpdateRegistryFromCache)
+        .expect("Could not update registry from cache");
+}
+
+pub fn on_compositor_ready() {
     // Call the special init hook function that we read from the init file
-    if let Err(error) = lua.execute::<()>(INIT_LUA_FUNC) {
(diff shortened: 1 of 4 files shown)


src/lua/thread.rs
```diff
@@ -152,10 +152,24 @@ pub fn init() {
...
+pub fn on_compositor_ready() {
+    info!("Running lua on_init()");
     // Call the special init hook function that we read from the init file
-    if let Err(error) = lua.execute::<()>(INIT_LUA_FUNC) {
-        error!("Lua init callback returned an error: {:?}", error);
-    }
+    send(LuaQuery::Execute(INIT_LUA_FUNC.to_owned()))
+        .err()
+        .map(|error| { error!("Lua init callback returned an error: {:?}", error); error });
+
```
Timidger: No trailing whitespace please

src/lua/thread.rs
```diff
@@ -152,10 +152,24 @@ pub fn init() {
...
     // Call the special init hook function that we read from the init file
-    if let Err(error) = lua.execute::<()>(INIT_LUA_FUNC) {
-        error!("Lua init callback returned an error: {:?}", error);
-    }
+    send(LuaQuery::Execute(INIT_LUA_FUNC.to_owned()))
+        .err()
+        .map(|error| { error!("Lua init callback returned an error: {:?}", error); error });
+    
     // Re-tile the layout tree, to make any changes appear immediantly.
     if let Ok(mut tree) = lock_tree() {
```
Timidger: This re-tile needs to be in `init`, not `on_compositor_ready`. If you restart the Lua thread, this method is never called (which is the correct logic). However, that means if a value that affects the layout is updated (e.g, you change the gap size or the colour of the borders for example) then it won't take effect immediately.

So please move this into `lua::thread::init`, right after you send the request to update the registry from the cache.

Timidger:
Interesting, it looks like this change also fixed a small bug I had missed -- the `on_startup` function was being called whenever Way Cooler was restarted. This should fix, which is nice.

src/lua/thread.rs
kiran-kp: Hmm, I assumed the layout wouldn't be initialized in init() since the compositor isn't ready yet. I'll test this.

src/lua/thread.rs
kiran-kp: Could you clarify which space you mean? I'm not seeing any trailing spaces in my editor. I did muck around with editor settings recently so maybe I enabled some weird formatting.

src/lua/thread.rs
kiran-kp: Updated. Re-tile seems to work fine in init.

src/lua/thread.rs
Timidger: When I pulled it down in my editor, it showed there whitespace on that blank line (165). It's not a big deal, I can always strip it if I notice any after I merge. I really should set up an auto-formatter for the project soon.

Timidger:
Looks good, thanks again!

PR #331, opened 2017-07-09
Multi head output fixes

Timidger:
Fixes a few issues/regressions with multi-head output:
* Borders will no longer be improperly sized when moving a container to a different output
* Containers will now be visible in their new output once they have been moved
* Backgrounds will not resize to try to fit on a background that they don't fit on


Timidger:
One more bug I want to fix: when moving between outputs, the borders aren't updated correctly for the moved container

PR #332, opened 2017-07-10
Remove clones on Size and Geometry, fixes #224

### LINKED ISSUE #224: Remove all the ".clones" on Size and Geometry types
Old versions of rustwlc use to require `Geometry` and `Size` to be cloned. Newer versions now implement the `Copy` trait on these values. A simple grep for `geometry`/`geo` and `size`/`resolution` should find most of these occurrences (or a grep for `.clone`, though that would be a lot to sift through).

yzhs:
This should remove all .clone() calls for Size and Geometry.


Timidger:
Sweet! I'll review this later today

PR #345, opened 2017-07-28
Added unbinding keys and allowing key passthrough.

### LINKED ISSUE #342: Ungrabbing for key bindings
I've written a small extension with which you can easily add modal window managing, inspired by modal text editing à la vi.
It works something like this:
```
normal = Mode.new("normal")
move = Mode.new("move")

normal:add({
  key({ mod }, "m", move:mode()) --Mode.mode() returns a function
})

move:add({
  key({ }, "h", "move_active_left"),
  ...
  key({ }, "return", normal:mode()) --"escape" is also bound to return to normal mode by default
})
```

For this I need to add the binding of the "h"-key when entering move-mode, and remove it when leaving.
Currently I set it to a noop-function, which obviously also blocks the "h"-key from working normally, since it is grabbed by way-cooler.

I see two solutions to this.

The simpler solution would allow unregistering key bindings.
On a mode-switch I also need(/want) to remove all key bindings of the current mode, which means that every mode-switch comes with a slew of unregistering and registering keys. I don't know if that's a good thing to do, performance-wise, since I expect some kind of associative data-structure is used for this.

This brings me to my second solution: Allowing the lua-callbacks to signal whether the pressed key should be swallowed or passed through to the active window. This could be achieved by allowing the callbacks to return a boolean, indicating what behavior is desired.
With this I would also not need to have a slew of unregistering and registering keys on mode-switches. I could simply assign a different callback to already registered keys, so the mapping data-structure does not need to add and remove a bunch of keys, just change some values.

Timidger:
This should fix #342.

In order to register a key for passthrough, note that you need **two** `true` statements after declaring a key. E.g:
```lua
-- enables passthrough to the mod+shift+h keybinding
key({ mod, "Shift" }, "h", function () print("Hello world!") end, true, true)
```

This is because there is a currently unused variable there (loop) and I did not want to break backwards compatibility in case I use that. In the future this might change, but because this is being added in a minor version bump I will not make this a breaking change (even if it wouldn't actually break anything).

@z33ky Can you please test this to ensure that it works for your use case? I'll take the next couple of days to run it on my own system (to ensure no regressions) and to review the code.

This PR also fixes a bug where the adding a key would mutate the original keybinding table that was passed in, meaning subsequent calls to it would be invalid.

### DIFF
--- a/lib/lua/lua_init.lua
+++ b/lib/lua/lua_init.lua
@@ -30,15 +30,20 @@ commands.init_workspaces = function(settings)
 end
 
 -- Create a new keybinding to register with Rust
-commands.key = function(mods, key, action, loop)
+commands.key = function(mods, key, action, loop, passthrough)
     assert(type(mods) == 'table', "modifiers: expected table")
     assert(type(key) == 'string', "key: expected string")
     if loop == nil then loop = true end
+    if passthrough == nil then passthrough = false end
     if type(action) ~= 'string' and type(action) ~= 'function' then
         error("action: expected string or function", 2)
     end
     return {
-        mods = mods, key = key, action = action, loop = loop
+      mods = mods,
+      key = key,
+      action = action,
+      loop = loop,
+      passthrough = passthrough,
     }
 end
 
@@ -46,15 +51,15 @@ commands.on_init = function() end
 commands.on_restart = function() end
 commands.on_terminate = function() end
 
-local use_key = ", use the `key` or `way_cooler.key` method to create a keybinding"
+local use_key = ", use the `way_cooler.key` method to create a keybinding"
 -- Converts a list of modifiers to a string
 local function keymods_to_string(mods, key)
     table.insert(mods, key)
     return table.concat(mods, ',')
 end
 -- Save the action at the __key_map and tell Rust to register the Lua key
-local function register_lua_key(index, action, loop)
-    local map_ix = rust.register_lua_key(index, loop)
+local function register_lua_key(index, action, loop, passthrough)
+    local map_ix = rust.register_lua_key(index, loop, passthrough)
     __key_map[map_ix] = action
 end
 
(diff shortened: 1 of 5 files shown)


z33ky:
It works, thanks!
I've pushed my extension [here](/z33ky/wc-apreggio) if you want to see how I use this feature. It's very much WIP, but it works.

lib/lua/lua_init.lua
```diff
@@ -91,6 +91,17 @@ commands.register_key = function(key)
     end
 end
 
+-- unregisters a keybinding
+commands.unregister_key = function(key)
```
z33ky: Using `commands.key` for this requires specifying an action: `unregister_key(key({ mod }, "x", ?))`.
Having a way that does not require specifying one would be more ergonomic. Maybe the check for the type of action should be moved to `commands.register_key`.

src/lua/rust_interop.rs
```diff
@@ -105,7 +106,29 @@ fn register_lua_key(mods: String, _repeat: bool, passthrough: bool)
...
-        }).map_err(|_| format!("Invalid keys '{}'", mods))}
+        }).map_err(|_| format!("Invalid keys '{}'", mods))
+}
+
+/// Rust half of unregistering a Lua key. This pops it from the key table, if
+/// it exists.
+///
+/// If a key wasn't registered, a proper error string is raised.
+fn unregister_lua_key(mods: String) -> Result<String, String> {
+    error!("Got {:#?}", mods);
```
z33ky: Is this a left over debug artifact?

Timidger:
@z33ky I tried to make the unregister key method use the same layout as the input to  register key, though in retrospect that's probably not what is best. 

I'll modify it so it takes in the mods and the key (eg the first two arguments to ` key`)

Timidger:
@z33ky There we go, that should be fixed. If there's any other issues please bring them up now. I hope to cut the release before the weekend is over.

z33ky:
LGTM

PR #350, opened 2017-07-29
Scrape active screen

### LINKED ISSUE #349: Screen scraping doesn't scrape active output
Eventually I want to stitch them together, but for now `wc-grab` should scrape the active output (instead of the default first one).

Timidger:
Adds a new screen endpoint to get the UUID of the active output.

Also ensures only scrapes the pixels of the active output.

Fixes #349


Timidger:
Since this isn't a breaking change, this will be included in 0.6.1. However, it **is** a breaking change on `wc-grab`'s side, so that will bump the minor version rather than the patch version.

PR #352, opened 2017-08-04
Add config directory to Lua's package.path

z33ky:
This way, `init.lua` can `require()` additional scripts in the same directory.
A consequence of this is that the `WAY_COOLER_INIT_FILE` environment variable can also influence the `package.path`.

I have a version checking that accessing `package.path` succeeds [here](1bf3521a520b73905a4b2d5b8adc53f42ae407a4), whereas this PR just uses `unwrap()`. Since `package.path` should be set by the Lua interpreter I think we can just assume it exists though.

### DIFF
--- a/src/lua/init_path.rs
+++ b/src/lua/init_path.rs
@@ -3,7 +3,7 @@
 
 use std::fs::{OpenOptions, File};
 use std::env;
-use std::path::Path;
+use std::path::{Path,PathBuf};
 
 use std::io::Result as IOResult;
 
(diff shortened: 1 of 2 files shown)


src/lua/init_path.rs
```diff
@@ -18,58 +18,64 @@ fn read_file<P: AsRef<Path>>(path: P) -> IOResult<File> {
...
     if let Ok(path_env) = env::var("WAY_COOLER_INIT_FILE") {
         info!("Found $WAY_COOLER_INIT_FILE to be defined, will look for the init file there.");
-        if let Ok(file) = read_file(Path::new(&path_env)) {
+        let path = Path::new(&path_env);
+        if let Ok(file) = read_file(&path) {
             info!("Reading init file from $WAY_COOLER_INIT_FILE: {}",
                   path_env.as_str().replace(home, "~"));
-            return Ok(file)
+            // If the parent doesn't exist it's just in the current directory.
+            let dir = path.parent().map_or(PathBuf::new(), Path::to_path_buf);
```
Timidger: This changes how the `WAY_COOLER_INIT_FILE` env works, which I'm fine with but if you could note it in your main comment that would be nice. When I cut releases I usually just look at the root comment, and it would be nice to not have to dig through the code to document this there.

src/lua/thread.rs
```diff
@@ -130,7 +130,13 @@ pub fn init() {
     let (use_config, maybe_init_file) = init_path::get_config();
     if use_config {
         match maybe_init_file {
-            Ok(init_file) => {
+            Ok((init_dir, init_file)) => {
+                if init_dir.components().next().is_some() {
+                    // Add the config directory to the package path.
+                    let mut package: hlua::LuaTable<_> = lua.get("package").unwrap();
+                    let paths: String = package.get("path").unwrap();
```
Timidger: While you're right that these can't ever happen, given a valid Lua environment, just in case can you change these to `expects` that say something to the effect of `"package not defined in lua"` and `"package.path not defined in lua"`. Makes it slightly easier to figure out the error when inevitably that user who uses a weird lua sends me a bug report ;)

src/lua/init_path.rs
z33ky: Sure. By main comment, do you mean the commit message?

src/lua/thread.rs
z33ky: The commit I wanted to link is this 1bf3521a520b73905a4b2d5b8adc53f42ae407a4, but I find that overly verbose. `expect` seems like the best choice.

Timidger:
By main comment I meant the description / root comment in the github issue thread. Eg in the description when you opened the thread. It just makes, it easier for me later.

z33ky:
Alright, changed the `unwrap()`s to `expect`s and added a note about `WAY_COOLER_INIT_FILE` in the main comment of the PR.

Timidger:
Looks good, thanks again! :)

PR #353, opened 2017-08-04
Change `CommandFn` to `fn()`

z33ky:
It currently uses the `Fn()` trait and thus requires `Arc`s for it. Since there are no closures used as `CommandFn`, but just plain functions, this could just use plain `fn()`.
If the use of closures should be allowed though, then `fn()` will not cut it.

### DIFF
--- a/src/commands/defaults.rs
+++ b/src/commands/defaults.rs
@@ -1,6 +1,5 @@
 //! Register commands on the registry.
 
-use std::sync::Arc;
 use std::process::{Command, Stdio};
 use std::thread;
 use std::io::prelude::*;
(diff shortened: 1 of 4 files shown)


Timidger:
That was written a while ago, and I think the use case was so that we could dynamically add functions at runtime (e.g closures or something), but thinking about it I don't think that feature will be necessary. Let me think about it for a bit just to be sure, but from a glance at the code it looks good assuming I decide I don't need that flexibility.

Thanks for this, it's probably small but anything to make this faster / reduce allocations is great :)

Timidger:
This shouldn't cause any problems further down the line, I'll merge it. Thanks again! :)

PR #362, opened 2017-08-07
Add info about openSUSE

jubalh:


Timidger:
Hey, is there a reason you added Way Cooler to the list of X11 window managers for the package list instead of the wayland one I found [here](https://build.opensuse.org/project/show/X11:Wayland)?

jubalh:
Its not a list.
I added it to the devel project for window managers. Historically its a sub project of X11.

The repo you show is the devel project for wayland related stuff. Its also a subproject of X11.

It could also go there, but since its a window manager I put it here. It doesnt really matter.

This page is not very detailed but maybe you would like to read it: https://en.opensuse.org/openSUSE:Build_Service_Concept_Devel_Project

Timidger:
Ok good just wanted to make sure. I confess I have never used OpenSUSE, but the fact that Sway was in the list I linked confused me so I wanted to make sure everything was all good. Thanks for clearing that up.

jubalh:
Yes basically its the choice of the packager to find a more or less suitable devel project. Sometimes there are multiple options which would fit.

PR #363, opened 2017-08-07
Add noop command

z33ky:
For [apreggio](/z33ky/wc-apreggio) I currently use a Lua noop function (`function() end`) to register keys before they are used to avoid necessitating creating new hashmap entries in runtime. With this commit I provide an equivalent command "noop" in Rust.

Using a Lua function has no issues, but of course using a Rust function instead avoids a round-trip to Lua just for doing nothing.
Wanting to "pre-register" keys may no be a commonly requested feature, but this change is also very small and has almost no impact on binary size, maintenance efforts, ect.

### DIFF
--- a/src/commands/defaults.rs
+++ b/src/commands/defaults.rs
@@ -29,6 +29,8 @@ pub fn register_defaults() {
     register("way_cooler_restart", way_cooler_restart);
     register("dmenu_lua_dofile", dmenu_lua_dofile);
 
+    register("noop", noop);
+
     /// Generate switch_workspace methods and register them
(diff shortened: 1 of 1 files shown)


Timidger:
I see no reason why this shouldn't be added.

Can you put a doc comment (e.g triple `/`, like `///`) on the `noop` function describing why it exists though (e.g to not require a roundtrip to Lua to do nothing)

z33ky:
I amended the commit with the comment.

Timidger:
Looks great, thanks!

PR #383, opened 2017-08-29
Added Awesome interface: keygrabber

Timidger:
This is the beginning of my attempt to emulate the Awesome API :tada:.

This is a simple one, it allows the user to set a callback to be triggered on every key press. I tried to make it as identical to the Awesome API as possible, but I haven't had a chance to test that it's 100% compatible (the key names _should_ be correct, but they my not be).

Solves most of #384 (still need to implement indexing, which requires signals which will be added in a later PR)

**If the API is in anyway inconsistent, please report the bug and the faulty input**

### DIFF
--- a/src/awesome/keygrabber.rs
+++ b/src/awesome/keygrabber.rs
@@ -55,7 +55,12 @@ fn call_keygrabber(lua: &Lua,
 
 fn run(lua: &Lua, function: rlua::Function) -> rlua::Result<()> {
     let keygrabber_table = lua.globals().get::<_, Table>(AWESOME_TABLE)?;
-    keygrabber_table.set(SECRET_CALLBACK, function)
+    match keygrabber_table.get::<_, Value>(SECRET_CALLBACK)? {
+        Value::Function(_) =>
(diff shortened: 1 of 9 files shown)


src/awesome/keygrabber.rs
```diff
@@ -0,0 +1,78 @@
...
+/// defined with the input.
+pub fn keygrabber_handle(mods: KeyboardModifiers, sym: Keysym, state: KeyState)
+                         -> rlua::Result<()> {
+    let lua = LUA.lock().expect("LUA was poisoned");
+    let lua_state = if state == KeyState::Pressed {
+        "press"
+    } else {
+        "release"
+    }.into();
+    let lua_sym = sym.get_name().expect("Key symbol did not have a defined name");
```
Timidger: Should return an error here

src/awesome/keygrabber.rs
```diff
@@ -0,0 +1,78 @@
...
+    let lua_mods = ::lua::mods_to_lua(&lua.0, mods.mods)
+        .expect("Could not convert mods to lua representation");
```
Timidger: Return error here as well

PR #402, opened 2017-09-19
Awesome Class structure + Awesome Button

### LINKED ISSUE #386: AwesomeWM: button API
## Functions
* [x] `add_signal`
* [x] `connect_signal`
* [x] `disconnect_signal`
* [x] `emit_signal`
* [x] `instances`
* [x] `set_index_miss_handler`
* [x] `set_newindex_miss_handler`
* [x] `connect_signal`
* [x] `disconnect_signal`
* [x] `emit_signal`

## Meta Functions
* [x] `__index`
* [x] `__newindex`
* [x] `__tostring`

Timidger:
This adds 90% of the [awesome button class](https://awesomewm.org/doc/api/classes/button.html#) (but doesn't integrate the functionality anywhere particularly). Fixes #386 

The button is intentionally unfinished, as I will have to implement other functionality until it's usable and I didn't want to implement so much that I couldn't easily make changes to this if the class structure changes in my implementation.

The main brunt of this PR is implementing the mechanisms that will let me build the interesting classes that will follow. E.g drawable and drawin, that will let me implement such lovely drawables as awesome notification and status bars :).

This PR is BLOCKED until [this rlua issue](https://github.com/chucklefish/rlua/issues/41) is released in a version. Right now this PR depends upon the git url which I don't want polluting master.


Timidger:
No longer blocked. I want to reduce the warnings before I get this in master though, because that can be a very slippery slope.

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+    button: xcb_button_t,
+    modifiers: KeyMod
+}
+
+#[derive(Clone, Debug)]
+pub struct Button<'lua>(Table<'lua>);
+
+impl <'lua> Button<'lua> {
+    pub fn button(&self) -> rlua::Result<Value<'lua>> {
+        let button = self.0.get::<_, ButtonState>("data")?;
```
Timidger: I don't like accessing the table directly, even in functions like this. Add a auto-implemented function for `Objectable` that takes `&self` and returns the `Table<T>` that exists in the `"data"` field. No boiler plate necessary, and it makes it less error prone.

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+        Ok(Value::Integer(button.button as _))
+    }
+
+    pub fn set_button(&self, new_val: xcb_button_t) -> rlua::Result<()> {
+        let mut button = self.0.get::<_, ButtonState>("data")?;
+        button.button = new_val;
+        self.0.set("data", button)?;
```
Timidger: Double check with a test, but this should be unnecessary

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+    pub fn modifiers(&self) -> rlua::Result<KeyMod> {
+        let button = self.0.get::<_, ButtonState>("data")?;
+        Ok(button.modifiers)
+    }
+
+    pub fn set_modifiers(&self, mods: Table<'lua>) -> rlua::Result<()> {
+        use ::lua::mods_to_rust;
+        let mut button = self.0.get::<_, ButtonState>("data")?;
+        button.modifiers = mods_to_rust(mods)?;
+        self.0.set("data", button)?;
```
Timidger: This should also be unnecessary

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+}
+
+impl <'lua> ToLua<'lua> for Button<'lua> {
+    fn to_lua(self, lua: &'lua Lua) -> rlua::Result<Value<'lua>> {
+        self.0.to_lua(lua)
+    }
+}
+
+impl <'lua> Objectable<'lua, Button<'lua>, ButtonState> for Button<'lua> {
```
Timidger: Can we implement these functions with a basic macro?

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+    fn _wrap(table: Table<'lua>) -> Button {
+        Button(table)
+    }
+
+    fn get_table(self) -> Table<'lua> {
+        self.0
+    }
+}
+
+impl Display for ButtonState {
```
Timidger: Move these impls before the `Button` impls

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+            button: xcb_button_t::default(),
+            modifiers: KeyMod::empty()
+        }
+    }
+}
+
+impl UserData for ButtonState {}
+
+/// Makes a new button stored in a table beside its signals
+pub fn allocator(lua: &Lua) -> rlua::Result<Object> {
```
Timidger: Why not make this a static method on `Button`? E.g `Button::allocate`?

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+    let meta = lua.create_table();
+    let class = class::button_class(lua)?;
+    Ok(Button::new(lua, class)?
+       .add_to_meta(meta)?
+       .build())
+}
+
+pub fn new<'lua>(lua: &'lua Lua, _table: Table<'lua>)
+                 -> rlua::Result<Object<'lua>> {
+    allocator(lua)
```
Timidger: Double check why this exists, and if it should exist behind a trait.

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+
+pub fn init(lua: &Lua) -> rlua::Result<Class> {
+    Class::new(lua, Some(Rc::new(allocator)), None, None)?
```
Timidger: Rename new to be `builder`

src/awesome/button.rs
```diff
@@ -0,0 +1,343 @@
...
+                                Some(lua.create_function(set_button))))?
+        .property(Property::new("modifiers".into(),
+                                Some(lua.create_function(set_modifiers)),
+                                Some(lua.create_function(get_modifiers)),
+                                Some(lua.create_function(set_modifiers))))?
+        .save_class("button")?
+        .build()
+}
+
+fn set_button<'lua>(lua: &'lua Lua, (table, val): (Table, Value))
```
Timidger: These still seem messy, due to the explicit conversion. I can't pass in an `Into<Button>`?

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use std::default::Default;
+use std::rc::Rc;
+use rlua::{self, Lua, ToLua, Table, UserData, AnyUserData, Value, Function};
+use super::object::Object;
+use super::property::Property;
+
+pub type Allocator = Rc<Fn(&Lua) -> rlua::Result<Object>>;
+pub type Collector = Rc<Fn(Object)>;
+pub type PropF<'lua> = rlua::Function<'lua>;
```
Timidger: Why not Rc?

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+pub type Checker = Rc<Fn(Object) -> bool>;
+
+#[derive(Debug)]
+pub struct Class<'lua> {
+    table: Table<'lua>
+}
+
+#[derive(Clone)]
+pub struct ClassState {
```
Timidger: Need comment explaining where the missing fields are compared to the C version (they are stored in Lua). Add a getter for them

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+        table.set_metatable(Some(meta));
+        Ok(ClassBuilder{
+            lua: lua,
+            class: Class { table }
+        })
+    }
+
+    #[allow(dead_code)]
+    pub fn properties(&self) -> rlua::Result<Table<'lua>> {
+        self.table.get("properties")
```
Timidger: Any particular reason we store this in the super table and not with the data (like with the checker)?

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+impl <'lua> From<Table<'lua>> for Class<'lua> {
+    fn from(table: Table<'lua>) -> Self {
+        Class { table }
+    }
+}
+
+fn set_index_miss_handler<'lua>(_: &'lua Lua, (obj, func): (Table, Function))
```
Timidger: Is this for the class? Make it a method on the class, since we don't use lua (or does it need this sig? If so, still make it a method please)

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+                                -> rlua::Result<()> {
+    let meta = obj.get_metatable()
+        .expect("Object had no metatable");
+    meta.set("__index_miss_handler", func)?;
```
Timidger: again, why not store with data like with properties?

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+fn set_newindex_miss_handler<'lua>(_: &'lua Lua, (obj, func): (Table, Function))
+                                   -> rlua::Result<()> {
+    let meta = obj.get_metatable()
+        .expect("Object had no metatable");
+    meta.set("__newindex_miss_handler", func)?;
+    Ok(())
+}
+
+pub fn button_class(lua: &Lua) -> rlua::Result<Class> {
```
Timidger: Make this a macro.

src/awesome/mod.rs
```diff
@@ -3,13 +3,20 @@ use rlua::{self, Lua};
 pub mod keygrabber;
 pub mod mousegrabber;
 pub mod awful;
+pub mod button;
```
Timidger: Does the whole module really need to be public

src/awesome/object.rs
```diff
@@ -0,0 +1,254 @@
...
+impl <'lua> ToLua<'lua> for Object<'lua> {
+    fn to_lua(self, _: &'lua Lua) -> rlua::Result<Value<'lua>> {
+        Ok(Value::Table(self.table))
+    }
+}
+
+impl <'lua> Object<'lua> {
+    /// Coerces a concrete object into a generic object.
+    ///
+    /// This requires a check to ensure data integrity, and it's often useless.
```
Timidger: replace useless with expensive

src/awesome/object.rs
```diff
@@ -0,0 +1,254 @@
...
+    {
+        let table = obj.get_table();
+        let has_data = table.contains_key("data")
+            .expect("Could not get data field of object table");
+        assert_eq!(has_data, true);
+        Object { table }
+    }
+
+    pub fn signals(&self) -> rlua::Table<'lua> {
+        self.table.get::<_, Table>("signals")
```
Timidger: Why not return result

src/awesome/object.rs
```diff
@@ -0,0 +1,254 @@
...
+            .expect("Object table did not have signals defined!")
+    }
+
+    pub fn table(self) -> rlua::Table<'lua> {
+        self.table
+    }
+}
+
+
+// TODO Abstract these two functions, they have similar functionality
```
Timidger: Do that

src/awesome/property.rs
```diff
@@ -0,0 +1,63 @@
+use rlua::{self, Lua, Value, ToLua, FromLua};
+use super::class::PropF;
+
+pub struct Property<'lua> {
```
Timidger: Add documentation explaining this isn't userdata and that it's just a Rust representation of a table from Lua.

src/awesome/button.rs
Timidger: Nope, `UserData` is cloned out, no way to get around the forced copy. May revisit if it becomes annoying. Adding an update per `Drop` is stupid and not really possible. So I'll just accept it

src/awesome/button.rs
Timidger: Done, but it might be better on Objectable...I won't move it yet, but time will tell...

src/awesome/button.rs
Timidger: Put in a closure instead

src/awesome/button.rs
Timidger: Nah it's fine, that's all Lua handling code. The raw setting of the state is behind `Button` which is the important part.

src/awesome/class.rs
Timidger: Because it's only used in one place, in the properties. Which is just a temporary type that isn't stored in lua directly except as a table which is what the properties struct deserializes to (with `ToLua`).

Moved to properties though.

src/awesome/class.rs
```diff
@@ -0,0 +1,163 @@
...
+    table: Table<'lua>
+}
+
+#[derive(Clone)]
+pub struct ClassState {
+    name: String,
+    // TODO Needed? If we store this like the object surely it's not needed...
+    /// The global Lua key that corresponds to this class' signal list.
+    /// Stored this way so that it's not tied to Lua lifetime.
+    signals_global: Option<String>,
```
Timidger: Move to meta table directly and eschue the weird stringly type? That would make it actually garbage collector counted too!

src/awesome/class.rs
Timidger: So we can't access it, because meta table is looked up on index from Lua!

src/awesome/class.rs
Timidger: Good point, I should do that so it's not accessed incorrectly. But whatever

src/awesome/object.rs
Timidger: Nigh, it's just useless

Timidger:
Ok done. Again, doesn't add much to user facing code except `button` which is useless. Just glad this pivotal change is in so I can work on making it make a difference :)

Issue #371, opened 2017-08-16
Resizing is not as performant as 5.x


Timidger:
An issue I've noticed when comparing Way Cooler 0.6.x to 0.5.2 (which I still run on my other laptop) is that it's much laggier to resize containers. This is probably a result of full-scale tiling being done on each geometry change, which needs to be investigated and fixed. The problem is much worse the more windows you have on the screen.

Arnaz87:
A posible cause of this problem is [this section of code](https://github.com/way-cooler/way-cooler/blob/master/src/layout/core/container.rs#L620-L634). The test always fails because [relocate_buffers](https://github.com/way-cooler/way-cooler/blob/master/src/layout/core/borders/borders.rs#L96) always modifies the geometry (to account for the borders and title) before saving it.

Timidger:
I will investigate @Arnaz87 report more closely tonight.

Timidger:
After testing it by changing the geometry comparison so that it occurs with the changes applied, I did not notice a big change in performance. Although, I tried to test 0.5.2 and I noticed it seem laggier than I remember.

Since, long term, I expect this tiling system to be done away with in favour of the tiling system of awesome in Lua I'm going to go ahead and close this as WONTFIX.

If someone else finds a solution and supplies a patch, I'd be more than happy to merge it. However, it's not worth me looking into at this time.

PR #404, opened 2017-10-02
Make border buffer allocator more explicit with vec! macro

alexander-smoktal:
Hi,

I've profiled app a bit and found, that using vec::repeat(0).take(n) is not an optimal way to allocate memory slice. 
It's a huge bottleneck.
The change replaces it with vec! macro.

The change may be related to https://github.com/way-cooler/way-cooler/issues/371

Attached profiling results. CPU spent on allocations dropped from 85% to less than 3%.

way-cooler-prof-fixes
way-cooler-prof-master

### DIFF
--- a/src/layout/core/borders/borders.rs
+++ b/src/layout/core/borders/borders.rs
@@ -1,4 +1,3 @@
-use std::iter;
 use std::cmp::{Eq, PartialEq};
 use rustwlc::{Geometry, Size, WlcOutput};
 use rustwlc::render::{calculate_stride};
@@ -53,7 +52,7 @@ impl Renderable for Borders {
         geometry.size.h += title_size;
         let Size { w, h } = geometry.size;
         let stride = calculate_stride(w) as i32;
-        let data: Vec<u8> = iter::repeat(0).take(h as usize * stride as usize).collect();
+        let data: Vec<u8> = vec![0; h as usize * stride as usize];
         let buffer = data.into_boxed_slice();
         let surface = ImageSurface::create_for_data(buffer,
                                                     drop_data,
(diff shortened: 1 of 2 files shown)


Timidger:
Hi there,

Thanks for the profiling! However, did you make sure you had optimizations enabled when you profiled this? It looks like the path is in `target/debug/way-cooler`?

I'll probably merge this either way, since it's clearer, but you might want to re-profile

alexander-smoktal:
It's much better in release with iterators, but anyway the change will help with debugging.

Timidger:
Ok, I'll test it when I get home.

But I found the exact same code in 5.2 [here](https://github.com/way-cooler/way-cooler/blob/v0.5.2/src/layout/core/borders/borders.rs#L57), so I'm not sure if this is the whole picture.

alexander-smoktal:
With release build it has the same performance, as with macro

Timidger:
Had a feeling, those iterators are zero cost.

I like the macro more, so I'll merge it. Do you mind updating the title though? Makes it easier for me to write release notes / see the history of the project at a more macro level than individual commits.

Timidger:
Also, if you want to improve the performance I think you should look at the number of times the tiling code  is called. I was a bit lazy when implementing tabbed / stacked tiling, and so it does a lot of unnecessary work when it could just update smaller parts of the layout tree.

If you want to work on that specifically I can send you to certain places in the code where I'm being inefficient and you can try to tackle it if you want

Timidger:
Actually, here's a link to a bottle neck I've noticed anyways, in case anyone wants to tackle this / makes it easier for me to do it later:

[here's the inefficient code now](https://github.com/way-cooler/way-cooler/blob/master/src/layout/commands.rs#L622)
[here's what it was in 0.5.2](https://github.com/way-cooler/way-cooler/blob/v0.5.2/src/layout/commands.rs#L477)

The reason that tile is there is so that we properly update tiled/stacked containers when we move focus. E.g this entails updating the borders and making sure the correct view is focused. To see what needs to be done, you can comment out the `layout_active_of` code in the master version and play around with nested tabbed/stacked containers to see what breaks.

There's probably other code like that for resize that I can't find right now. Fixing this by making some helper functions that updates tabbed/stacked containers in the tree would make it easier to fix the resize issue once I find the place it's calling tiling a lot.

PR #405, opened 2017-10-04
Fix tabbed/stacked container title

### LINKED ISSUE #381: Tabbed / Stacked counter doesn't update if focusing descendant of one of the children
Reproduce:
1) Make tabbed/stacked container
2) Make two windows
3) In one of the windows, make a sub container.
4) Notice that when switching to that sub container, the tabbed/stacked counter in the ancestor container doesn't update

alexander-smoktal:
Fixed #381 issue, when tabbed/stacked containers with subcontainers didn't update title counter of an active window.
Added optimization to not include children duplicates in a border drawing function.

I've added the HashSet because children vector contains parent node and its children. In this case we add another children instances because of `self.tree.grounded_children(child_ix)` for the parent container.

### DIFF
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -1,4 +1,6 @@
 use std::cmp;
+use std::collections::HashSet;
+use std::iter::FromIterator;
 
 use petgraph::graph::NodeIndex;
 use rustwlc::{WlcView, Geometry, Point, Size, ResizeEdge};
@@ -880,15 +882,24 @@ impl LayoutTree {
     /// Draws the borders recursively, down from the top to the bottom.
     fn draw_borders_rec(&mut self, mut children: Vec<NodeIndex>)
                         -> CommandResult {
+        let mut updated_nodes: HashSet<NodeIndex> = HashSet::from_iter(children.iter().cloned());
+
         while children.len() > 0 {
             let child_ix = children.pop().unwrap();
-            children.extend(self.tree.grounded_children(child_ix));
+
+            for child in self.tree.grounded_children(child_ix) {
+                if !updated_nodes.contains(&child) {
+                    updated_nodes.insert(child);
+                    children.push(child);
+                }
+            }
+
             let parent_ix = self.tree.parent_of(child_ix)
                 .expect("Node had no parent");
             let children = self.tree.children_of(parent_ix);
             let index = children.iter().position(|&node_ix| node_ix == child_ix)
                 .map(|num| (num + 1).to_string());
-            if Some(child_ix) != self.active_container {
+            if !self.tree.on_path(child_ix) && Some(child_ix) != self.active_container {
                 self.set_borders(child_ix, borders::Mode::Inactive)?;
             } else {
                 match self.tree[parent_ix] {


Timidger:
Thanks, I'll test this out later tonight

alexander-smoktal:
I suspect there is a better way to avoid children duplication. But it requires deeper changes in the codebase.

Timidger:
Hmm, I thought of another way to code this algorithm, not sure if it's sufficiently more efficient / less clear though. I'll let you be the judge.

Use `drain` to clear the vector and convert it to a `HashSet`, like so:

```rust
let mut children_set = children.drain(..).collect::<HashSet<NodeIndex>>();
```

In order to make that more efficient you can pre-allocate the `HashSet` using `with_capacity` with the length of the `children` Vec.

`children_set` extend with the grounded children (no need for a loop with a branch)

```rust
children_set.extend(self.tree.grounded_children(child_ix));
```

Finally, convert back to `Vec` (with optional `with_capacity` optimization like before)

```rust
children = temp.into_iter().collect::<Vec<NodeIndex>>();
```

Full code:

```rust
// Up here you could init children_set using with_capacity and then keep a mutable reference to it that you can update so as to have the amortized cost of allocating the HashSet like you do in your implementation
while children.len() > 0 {
    let child_ix = children.pop().unwrap();
    let mut children_set = children.drain(..).collect::<HashSet<NodeIndex>>();
    children_set.extend(self.tree.grounded_children(child_ix));
    children = children_set.into_iter().collect::<Vec<NodeIndex>>();

    ...
```

Honestly, either is fine with me. If you think yours is better (and I wouldn't blame you if you did) then we'll go with that

Timidger:
Also implementation looks good, works on my machine at least ;)

src/layout/actions/layout.rs
```diff
@@ -880,15 +882,24 @@ impl LayoutTree {
...
+                }
+            }
+
             let parent_ix = self.tree.parent_of(child_ix)
                 .expect("Node had no parent");
             let children = self.tree.children_of(parent_ix);
             let index = children.iter().position(|&node_ix| node_ix == child_ix)
                 .map(|num| (num + 1).to_string());
-            if Some(child_ix) != self.active_container {
+            if !self.tree.on_path(child_ix) {
```
Timidger: Woops, spoke to soon. Found one regression with your changes. You can't use the `on_path` check, you have to check the `active_container` directly, otherwise it breaks border focus when a subsurface is spawned for a view.

A subsurface is usually a drop down menu, or a right click menu, or one of those popup "hover" bubbles. They are views in their own right, but they have another view as their parent. While they are treated as full views (e.g they have a path in the layout tree), they don't update the `active_container`. This is a bit of a hack, but the active path will _not_ always point to the `active_container` (yes that's confusing, I should perhaps re-think that later).

So if you could revert this change (to do the original `Some(child_ix) != self.active_container` check) that would fix this issue.

alexander-smoktal: Ok. I'll take a look

alexander-smoktal: About the change with vector drain: Don't you think it will be slow? Having just a hashset of indices is one thing. Creating a set every time and then vector from it may be too slow. Because of memory fragmentation for example. I don't like this condition with 2 pushes eather, but I'd say it's better. 

Maybe there is a reason to use some function to redraw just visible containers, not to pass children? It's a tree, so a list of visible containers will be relatively small. 
And the function will be much simpler. Something like:
```rust
while !visible.empty() {
    let node = visible.pop();
    if node.type == Stacked {
        let index = container.iter().position(|&node| node.ix == visible.head()));
        b.set_title(format!("{:?} ({}/{})", layout, index.unwrap_or("?".into()), node.children.len())
    }
}
```

What do you think about this?

alexander-smoktal: But as I understand, it doesn't fix an issue with my change. Maybe to store a list of visible containers instead of active container? Use a list of active containers? )

alexander-smoktal: What do you think about another condition
```rust
Some(child_ix) != self.active_container || !self.tree.on_path(child_ix) {
    self.set_borders(child_ix, borders::Mode::Inactive)?;
}
````
So if it's not on path or not and active container (container with popup will remain active), set inactive.

alexander-smoktal:
I've posted a lot of comment to the change request and pushed new version. Now I have no idea where all the comments went.

Timidger:
Is there a reason you need to use `tree.on_path` in your changes? It should only be updating the color of the borders of the active container, does it some way affect your tabbed/stacked title fix that I'm unaware of?

Timidger:
You're probably right about the constant allocation for new `HashSet`, I was hoping there would be a way to pre-allocate and re-use the old ones but I don't think there is. Your current implementation is fine. Just need to fix that regression

src/layout/actions/layout.rs
Timidger: A function to return visible children could be helpful and probably improve performance. There would need to be more testing necessary by me to ensure that there aren't any other border focusing bugs (there are weird edge cases, one of which I can think of is switching focus from one window to a window in another output. The first window needs to be "inactive" now, so it needs to still be considered visible even though its not in the active output).

If you want to make this change, I suggest you open a different PR. Lets keep these changes focused on this PR, and if you want to add the "visible" list we can talk about it on gitter to see what other tiling methods would benefit from this change.

Timidger:
Oh I see duh you added the `on_path` check because otherwise it doesn't check non-focused containers. Ok, this looks great. Thanks again!

PR #406, opened 2017-10-05
Add missing close bracket to README.md

samdoshi:
Every time I see that unbalanced open bracket...

(Maybe I'll be back if XMonad doesn't make the Wayland transition.)


Timidger:
Thanks!

PR #408, opened 2017-10-09
Update background to use desktop-shell protocol

### LINKED ISSUE #141: Commission an official background image

### LINKED ISSUE #159: Associate background surface with output nodes in the tree
Right now the background just exists regardless of the state of the tree. With multi-head  however, there is a need to have different backgrounds on different outputs. We should associate a background surface to an output container. This will make it easier to track. This change should also be bundled with #143, since they both go hand in hand.

Timidger:
This fixes the terrible, terrible hack that is the background program.

### Notable new features
* Allows the background client to specify which output to use (with `way-cooler-bg` now automatically covering all outputs, woohoo)
* No longer based on the name of the window (so clients named "background" will no longer suddenly become the background. Seriously, that was pretty silly)
* Adds the desktop-shell protocol 
* Switching TTYs no longer makes the background mysteriously disappear   

### Lost features
* Can't run `wc-bg` again while one is active, you need to manually kill the old process first. Note that restarting in place (e.g `mod+r`) still works

### This requires the latest version of `wc-bg` (formally `way-cooler-bg`). This is a backward incompatible change.

Fixes #141 (`wc-bg` now uses it by default), Fixes #159 and Fixes #159

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6,7 +6,7 @@ dependencies = [
  "cairo-rs 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "dbus 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "dbus-macros 0.0.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "dummy-rustwlc 0.6.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "dummy-rustwlc 0.6.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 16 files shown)


Timidger:
Build failing due to changes in rust-wlc not being propagated to dummy-rustwlc. Will fix

Timidger:
There is one issue with this PR still I need to address: Spawning another background program while one is active (e.g running `wc-bg` twice) causes a segfault

Timidger:
Issue has been "fixed" by not allowing you to spawn a background if one is active. Or rather, it will spawn just not as a background.

This is due to a whole bunch of issues, the end result being xwayland crashes because we deny the new background spawning due to a race condition in the protocol (it tries to remove the last background, but due to the order of event dispatch in wayland the surface comes before that request happens)

Issue #407, opened 2017-10-06
A Tabbed/Stacked nested in another Tabbed/Stacked container doesn't properly show the focused view


Timidger:
To reproduce issue:

1) Spawn Tabbed/Stacked container
2) Make a Tabbed/Stacked container inside of that first Tabbed/Stacked container
3) Add a new view to that nested container
4) Notice that you can only see the latest one, not any the previous ones that you spawned.

alexander-smoktal:
I'll try to fix this one

alexander-smoktal:
Fixed

Timidger:
Correct, not sure why it didn't close automatically haha

Timidger:
Note that this issue was not totally fixed by @alexander-smoktal patch, as it caused a regression. Going to keep this closed since it is fixed but refer to #413 for further discussion

PR #409, opened 2017-10-09
Fix Tabbed/Stacked nested container front view

alexander-smoktal:
Fix Issue #407:
A Tabbed/Stacked nested in another Tabbed/Stacked container doesn't properly show the focused view

### DIFF
--- a/src/layout/core/tree.rs
+++ b/src/layout/core/tree.rs
@@ -216,37 +216,31 @@ impl LayoutTree {
 
     /// Sets the active container to be the given node.
     pub fn set_active_node(&mut self, node_ix: NodeIndex) -> CommandResult {
-        if self.active_container != Some(node_ix) {
-            info!("Active container was {}, is now {}",
-                  self.active_container.map(|node| node.index().to_string())
-                    .unwrap_or("not set".into()),
-                  node_ix.index());
-        }
-        let id = self.tree[node_ix].get_id();
-        if let Some(fullscreen_id) = try!(self.in_fullscreen_workspace(id)) {
-            if fullscreen_id != id {
-                return Err(TreeError::Focus(FocusError::BlockedByFullscreen(id, fullscreen_id)))
+        let (container_id, container_type) = {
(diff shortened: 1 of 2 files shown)


alexander-smoktal:
Added fix. Also refactored a bit.

The fix is to add `handle.bring_to_front();` into `focus_on` function. As I understand if you focus on a view, you always want it to be in front of other windows?

alexander-smoktal:
It's not that I don't like refactored code, but I rearranged code to be more readable for a side developer. If you don't like it, I'll revert.

src/layout/core/tree.rs
```diff
@@ -216,37 +216,31 @@ impl LayoutTree {
...
-            ContainerType::View => try!(self.focus_on(id)),
+
+        match container_type {
+            ContainerType::View => self.focus_on(container_id)?,
             ContainerType::Container => {/* TODO implement this */},
             _ => return Err(
-                TreeError::UuidWrongType(id, vec!(ContainerType::View, ContainerType::Container)))
+                TreeError::UuidWrongType(container_id, vec!(ContainerType::View, ContainerType::Container)))
         }
-        self.tree.set_ancestor_paths_active(node_ix);
```
Timidger: Why was this line removed? Did this interfere and cause an issue?

Timidger:
Looks great, thanks for the refactoring. There's a fair amount of code that hasn't been brought up to date e.g using `?` instead of `try!`, and a lot of the code makes sense to me but not necessarily anyone else ;). 

Those small rewording/restructurings really help make it clear to anyone reading it what's going on without having to know the entire codebase and its conventions.

src/layout/core/tree.rs
alexander-smoktal: We have it inside `focus_on` call

Timidger: Ah, good catch. Thanks for finding and removing the dup

Timidger:
Thanks again for your contributions!

PR #413, opened 2017-10-14
Fixed floating display issue, and #407

Timidger:
cc @alexander-smoktal , there was an issue with your implementation so I modified it significantly. It's not nearly as pretty, but it fixes the original nested tabbed/stacked issue without causing floating views to hide behind tiled ones.

To reproduce the issue this fixes:

1) Spawn a tiled window (doesn't have to be tabbed or stacked)
2) Spawn a floating window
3) Focus on tiled window, notice how the floating window is hidden when it shouldn't be.

An alternative fix would be to keep your fix and enumerate the floating containers afterwards and make them come forward, but this is less complicated because it only touches the containers within the tiled/stacked container

### DIFF
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -39,8 +39,8 @@ impl LayoutTree {
                     Layout::Tabbed | Layout::Stacked => {
                         for child_ix in self.tree.children_of(parent_ix) {
                             match self.tree[child_ix] {
-                                Container::View { handle, .. } => {
-                                    if child_ix != node_ix {
+                                Container::View { handle, floating, .. } => {
+                                    if child_ix != node_ix && !floating {
                                         handle.send_to_back();
                                     }
                                 },
--- a/src/layout/actions/focus.rs
+++ b/src/layout/actions/focus.rs
@@ -32,10 +32,33 @@ impl LayoutTree {
         }
         let node_ix = self.tree.lookup_id(uuid)
             .ok_or(TreeError::NodeNotFound(uuid))?;
+        let parent_ix = self.tree.parent_of(node_ix)?;
+        match self.tree[parent_ix] {
+            Container::Container { layout, .. } => {
+                match layout {
+                    Layout::Tabbed | Layout::Stacked => {
+                        for child_ix in self.tree.children_of(parent_ix) {
+                            match self.tree[child_ix] {
+                                Container::View { handle, .. } => {
+                                    if child_ix != node_ix {
+                                        handle.send_to_back();
+                                    }
+                                },
+                                Container::Container { ..}  => {
+                                    // do nothing
+                                },
+                                _ => unreachable!()
+                            }
+                        }
+                    },
+                    _ => {}
+                }
+            }
+            _ => {}
+        }
         match self.tree[node_ix] {
             Container::View { handle, .. } => {
                 handle.focus();
-                handle.bring_to_front();
                 self.active_container = Some(node_ix);
             },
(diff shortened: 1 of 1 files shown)


alexander-smoktal:
I see

src/layout/actions/focus.rs
```diff
@@ -32,10 +32,33 @@ impl LayoutTree {
         }
         let node_ix = self.tree.lookup_id(uuid)
             .ok_or(TreeError::NodeNotFound(uuid))?;
+        let parent_ix = self.tree.parent_of(node_ix)?;
+        match self.tree[parent_ix] {
+            Container::Container { layout, .. } => {
+                match layout {
+                    Layout::Tabbed | Layout::Stacked => {
+                        for child_ix in self.tree.children_of(parent_ix) {
```
alexander-smoktal: In this case you get children in random order. So eventually top window (not floating) will be the last one, but this is not the window you focus on. 
From my point of view, it's impossible to implement this logic w/o storing floating window pointer somewhere in the tree struct. In this case you could bring focused window in front and then floating.

alexander-smoktal: Actually this is possible, but floating window is something outstanding, so why not to store pointer it somewhere else

Timidger: It's not [actually random order](https://github.com/way-cooler/way-cooler/blob/master/src/layout/core/graph_tree.rs#L580) it's sorted by the focus weight, but yes for this purpose it might as well be.

If you look down at line 43, the only thing this loop is doing is sending the other windows to the back (i.e, just moving them behind the other tabbed/stacked that is already there). When you send all but the one you want to see to the back, the one you want to focus on is naturally at the front.

Is there an issue you can reproduce even with these changes? Or are you commenting purely on the code quality (which I admit is messy)

alexander-smoktal: I'm not sure if this is because of this change, but with floating window, there're a lot of strange behaviours. For example when the active container is not parent container of a floating window. If you change active views and later change floating window, behaviour is unpredictable. Sometimes static view becames just black.

alexander-smoktal: What if keep the code with `bring_to_front` of ficused view and then lookup for a floating window and make `bring_to_front` for it?

Timidger: That could work as well, feel free to make those changes and see if the floating issues you reported persist. Otherwise I can do it later tonight.

alexander-smoktal:
Ok, I know how to reproduce the issue, but don't know how to fix it:
- Create a bunch of stacked containers;
- Make the last view of the last container floating;
- Switch active to the middle of stacked containers;
- Make active view floating;

Result: When you switch containers, the topmost container is inactive. 
Even more. Topmost view is often not the active one.

alexander-smoktal:
Also I'm able to reproduce this without changes. It looks like there's some issue with floating window.

alexander-smoktal:
Ok. I'm retarded. There are a couple of floating windows. They don't replace each other.

alexander-smoktal:
Another question then: Should active floating window be on-top of all floating windows?

Timidger:
>Another question then: Should active floating window be on-top of all floating windows?

Yes

alexander-smoktal:
Hi. Sorry, was a little busy last week. The change fixes my regression. With topmost floating window, I'll create another issue and try to fix it.

Timidger:
Ugh I see what you mean now, the windows are hidden behind the background with this change.

/sigh

PR #414, opened 2017-10-15
Lockscreen protocol update

Timidger:
Based on changes made in [this pr](https://github.com/way-cooler/way-cooler-lock/pull/5).

### Notable Features
* Ability to lock all the screens, not just one (means we are no longer compliant with desktop shell protocol though, might fix that in the future)
* Fixes the lockscreen so it's based on an actual Wayland protocol instead of a bad hack

### Breaking Changes
* `"lock_screen"` is no longer a built-in command, instead explicitly call `wc-lock`
* Can no longer lock screen using D-Bus (that method has been removed from the interface)
* Old versions of `wc-lock` will not work with this patch
* Scrape now takes an output as an argument, **this is a breaking change for `wc-grab` as well**. See [this pr for more details](https://github.com/way-cooler/way-cooler-grab/pull/4)
    + To use the old behaviour, simply pass `0` for the inputs to the `Scrape` D-Bus method

**Old versions of `wc-lock` are incompatible with this change. Please use a version >= 0.2**

### DIFF
--- a/protocols/desktop-shell.xml
+++ b/protocols/desktop-shell.xml
@@ -18,6 +18,7 @@
     </request>
 
     <request name="set_lock_surface">
+      <arg name="output" type="object" interface="wl_output"/>
       <arg name="surface" type="object" interface="wl_surface"/>
     </request>
 
--- a/src/modes/lock_screen.rs
+++ b/src/modes/lock_screen.rs
@@ -108,7 +108,10 @@ impl LockScreen {
 #[allow(unused)]
 impl Mode for LockScreen {
     fn output_render_post(&mut self, output: WlcOutput) {
-        // Don't allow screen scraping during lock screen
+        Default.output_render_post(output);
+        // TODO Uncomment
+        // we should allow this some other way
+        /*// Don't allow screen scraping during lock screen
         let need_to_fetch = read_screen_scrape_lock();
         if *need_to_fetch {
             if let Ok(mut scraped_pixels) = scraped_pixels_lock() {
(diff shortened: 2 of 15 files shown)


protocols/desktop-shell.xml
```diff
@@ -18,6 +18,7 @@
     </request>
 
     <request name="set_lock_surface">
+      <arg name="output" type="object" interface="wl_output"/>
```
Timidger: This is where we modify the protocol. The output is sent along with the surface, so it knows which one to bind to.

src/modes/lock_screen.rs
```diff
@@ -166,97 +179,78 @@ impl Mode for LockScreen {
...
-    match path {
-        None => warn!("No lock screen program set!"),
-        Some(path) => {
-            let path = path.into();
-            lock_screen_with_path(path).unwrap_or_else(|err| {
-                warn!("Could not lock screen: {:?}", err)
-            })
+pub fn lock_screen(client: *mut wl_client, output: WlcOutput) {
+    let mut mode = write_current_mode();
+    warn!("Locking screen w/ mode {:#?}", mode);
```
Timidger: Remove

src/modes/lock_screen.rs
```diff
@@ -166,97 +179,78 @@ impl Mode for LockScreen {
...
+    {
+        match *mode {
+            Modes::LockScreen(ref mut lock_mode) => {
+                warn!("Found another view");
```
Timidger: Remove

Timidger:
Ok finally merged this after testing for way too long

PR #417, opened 2017-10-25
Hotfix for hiding behind backgrounds tabbed stacked

Timidger:
Fixes my screw up with https://github.com/way-cooler/way-cooler/pull/413
Fixes warnings


Timidger:
The floating is still an issue, but it's actually only an issue when you create a view. I didn't test enough, and it's actually an issue with view creation calling the `focus_on` method which forces it to obscure floating views.

I can worry about that fix in a different issue, but it's really not a big deal in the short term

PR #423, opened 2017-10-28
Can now load default Awesome rc.lua without it crashing

Timidger:
Loading the Awesome file doesn't _do_ anything, but the interface is set up so that contributors can easily work on different parts of it (and there's already skeleton code there to get them started).

Still some work needs to be done before I'm ok merging this in:

- [x] Unify the "dummy" types and functions so it's easy to ensure they're not needed because they'll be marked as "dead code"
- [x] Make issues for each of the components that are so far defined.
- [x] Enable the debug library in rlua (see ~[this issue](https://github.com/chucklefish/rlua/issues/52)~ [this pr](https://github.com/chucklefish/rlua/pull/54))
- [x] Fix `package.path` and `package.cpath` setup in Lua init
- [x] Keep old `init.lua`, add `rc.lua` that is copied from Awesome (we are targeting the default, to keep things simple for now)

There's more of the awesome interface that hasn't been encoded, but that can wait until we have the default `rc.lua` actually running in Way Cooler 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ cache:
 addons:
   apt:
     packages:
-      - lua-compat53
       - libnih-dbus1
       - libdbus-1-3
       - libcurl4-openssl-dev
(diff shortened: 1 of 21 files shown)


Timidger:
Ugh, ok looks like the debug library really *is* super unsafe. So we are just going to have to mock debug calls as they come up. This will make debugging the code way, way harder than it needs to be, which is a shame...

Timidger:
Holding off on this until I can get the debug library working without crashing

Timidger:
This has been fixed in rlua, in 0.9.7. Will update the version, remove debug mocks, do a final review, and merge this in later tonight. Then it will finally be possible to work asynchronously on the various Awesome interfaces :)

PR #425, opened 2017-10-29
Fixed #415

### LINKED ISSUE #415: Topmost floating window is not always the active one
Topmost floating window is always the same. But it'd be better to have active floating window on top. 
To reproduce:
1. Create 2 floating windows;
2. Switch to the first one;

Result: Active floating window is hidden by the last one.

Timidger:
@alexander-smoktal Just to double check, does this fix the issue? 

It's not a very performant fix, but it should fix it.


alexander-smoktal:
Sorry, just noticed the message

Issue #436, opened 2017-11-12
Display tabbed and stacked containers' title bars like i3


Arnaz87:
Display tabs' title bars side to side on the container title bar, instead of how they are being displayed right now, counting the contained tabs. Similarly, stacked container should display the titles of it's children on top of one another.

Timidger:
Your welcome to send a patch, but I am not planning to do this in the near future.

Timidger:
This has mostly been fixed, there are a few remaining issues. Namely #440, #442, and #443.

Going to go ahead and close this though.

PR #439, opened 2017-11-22
Tabbed/Stacked container titlebars

### LINKED ISSUE #410: Title bar doesn't work properly in case its size differs from the size of borders
Steps to reproduce:
1. Open `~/.config/way-cooler/init.lua`
2. Change `way_cooler.windows.borders.size` to `10`
3. Run binary

Result: container title bar is too narrow. There is a hole below.

Arnaz87:
Display titlebars of tabbed and stacked containers like i3 does. Children containers/views still draw their titlebars though.

Probably fixes #410, but needs more testing to confirm.

### DIFF
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -931,10 +931,8 @@ impl LayoutTree {
             // self.tree, and I need self.tree to get the children's titles
             let titles = children.iter().map(|&node_ix| {
                 match self.tree[node_ix] {
-                    Container::Container { ref mut borders, .. } =>
-                        borders.as_ref().map( |b| b.title() )
-                            .unwrap_or("").into(),
-                    Container::View { ref mut borders, .. } =>
+                    Container::Container { ref mut borders, .. } |
+                    Container::View { ref mut borders, .. } => 
                         borders.as_ref().map( |b| b.title() )
                             .unwrap_or("").into(),
                     ref child => {
@@ -952,13 +950,13 @@ impl LayoutTree {
                         if layout == Layout::Tabbed || layout == Layout::Stacked {
                             borders.as_mut().map(|b| {
 
-                                b.set_children(titles, index.unwrap());
+                                b.set_children(titles, index.expect("The node index was not a child"));
 
                                 // Still necesary to draw the title when this
                                 // container is inside another tabbed/stacked
                                 b.set_title(format!("{:?} ({}/{})",
                                                     layout,
-                                                    index_str.unwrap_or("?".into()),
+                                                    index_str.unwrap_or_else(|| "?".into()),
                                                     children.len()
                                 ));
                             });
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -901,8 +901,26 @@ impl LayoutTree {
             let parent_ix = self.tree.parent_of(child_ix)
                 .expect("Node had no parent");
             let children = self.tree.children_of(parent_ix);
-            let index = children.iter().position(|&node_ix| node_ix == child_ix)
-                .map(|num| (num + 1).to_string());
+            let index = children.iter().position(|&node_ix| node_ix == child_ix);
+            let index_str = index.map(|num| (num + 1).to_string());
+
+            // I can't do it inside the match ahead because it borrows
+            // self.tree, and I need self.tree to get the children's titles
+            let titles = children.iter().map(|&node_ix| {
+                match self.tree[node_ix] {
+                    Container::Container { ref mut borders, .. } =>
+                        borders.as_ref().map( |b| b.title() )
+                            .unwrap_or("").into(),
+                    Container::View { ref mut borders, .. } =>
+                        borders.as_ref().map( |b| b.title() )
+                            .unwrap_or("").into(),
+                    ref child => {
+                        error!("Container child is {:#?}", child);
+                        panic!("Container childr is not a Container nor a View")
+                    }
+                }
+            }).collect();
+
             if !self.tree.on_path(child_ix) && Some(child_ix) != self.active_container {
                 self.set_borders(child_ix, borders::Mode::Inactive)?;
             } else {
@@ -910,9 +928,14 @@ impl LayoutTree {
                     Container::Container { layout, ref mut borders, .. } => {
                         if layout == Layout::Tabbed || layout == Layout::Stacked {
                             borders.as_mut().map(|b| {
+
+                                b.set_children(titles, index.unwrap());
+
+                                // Still necesary to draw the title when this
+                                // container is inside another tabbed/stacked
                                 b.set_title(format!("{:?} ({}/{})",
                                                     layout,
-                                                    index.unwrap_or("?".into()),
+                                                    index_str.unwrap_or("?".into()),
                                                     children.len()
                                 ));
                             });
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -423,8 +423,13 @@ impl LayoutTree {
                 // NOTE Do _NOT_ use set_layout,
                 // the normalization causes the issue described in #344
                 match self.tree[parent_ix] {
-                    Container::Container { ref mut layout , ..} => {
-                        *layout = new_layout
+                    Container::Container { ref mut layout , ref mut borders, ..} => {
+                        *layout = new_layout;
+                        // This needs to be done manually because
+                        // apparently set_layout cannot be used.
+                        if let Some(ref mut _b) = borders.as_mut() {
+                            _b.layout = Some(new_layout);
+                        }
                     },
                     _ => unreachable!()
                 }
@@ -598,8 +603,13 @@ impl LayoutTree {
 
     pub fn set_layout(&mut self, node_ix: NodeIndex, new_layout: Layout) {
         match self.tree[node_ix] {
-            Container::Container { ref mut layout, .. } => {
+            Container::Container { ref mut layout , ref mut borders, .. } => {
                 *layout = new_layout;
+                // This needs to be done manually because
+                // for some reason set_layout cannot be used.
+                if let Some(ref mut _b) = borders.as_mut() {
+                    _b.layout = Some(new_layout);
+                }
             },
             ref container => {
                 warn!("Can not set layout on non-container {:#?}", container);
@@ -821,22 +831,35 @@ impl LayoutTree {
     fn update_container_geo_for_borders(&mut self, node_ix: NodeIndex,
                                         mut geometry: Geometry)
                                         -> Result<Geometry, TreeError> {
+        // Stacked containers need bigger title bars.
+        // This has to be done before borrowing self.tree to use the container
+        let child_count = self.tree.children_of(node_ix).len();
+
         let container = &mut self.tree[node_ix];
 
         match *container {
-            Container::Container { ref mut apparent_geometry,
+            Container::Container { layout,
+                                   ref mut apparent_geometry,
                                    geometry: ref mut actual_geometry,
                                    ref borders, .. } => {
                 *actual_geometry = geometry;
                 if borders.is_some() {
+                    let title_count = match layout {
+                        Layout::Stacked =>
+                            // I don't understand why I have to use this
+                            // instead of just child_count, but seems to work...
+                            (2*child_count-1) as u32,
+                        _ => 1
+                    };
                     let gap = Borders::gap_size();
                     let thickness = Borders::thickness() + gap;
                     let edge_thickness = thickness / 2;
                     let title_size = Borders::title_bar_size();
+                    let real_title_size = (title_size / 2) * title_count;
                     geometry.origin.y += edge_thickness as i32;
-                    geometry.origin.y += (title_size / 2) as i32;
+                    geometry.origin.y += real_title_size as i32;
                     geometry.size.h = geometry.size.h.saturating_sub(edge_thickness);
-                    geometry.size.h = geometry.size.h.saturating_sub(title_size / 2);
+                    geometry.size.h = geometry.size.h.saturating_sub(real_title_size);
                 }
                 *apparent_geometry = geometry;
             },
--- a/src/layout/core/borders/borders.rs
+++ b/src/layout/core/borders/borders.rs
@@ -6,6 +6,7 @@ use cairo::{ImageSurface, Format};
 use uuid::Uuid;
 use ::registry;
 use ::render::{Color, Renderable};
+use super::super::container::Layout;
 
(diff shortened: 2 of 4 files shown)


Timidger:
Awesome! I'll test this soon and get back to you. Quick peek at the code looks like you ran into a few borrowck issues, so I'll see if I can help you work around them without ugly code or unnecessary copying (if we need to though, not the end it the world)

src/layout/actions/layout.rs
```diff
@@ -423,8 +423,13 @@ impl LayoutTree {
                 // NOTE Do _NOT_ use set_layout,
                 // the normalization causes the issue described in #344
                 match self.tree[parent_ix] {
-                    Container::Container { ref mut layout , ..} => {
-                        *layout = new_layout
+                    Container::Container { ref mut layout , ref mut borders, ..} => {
+                        *layout = new_layout;
+                        // This needs to be done manually because
+                        // apparently set_layout cannot be used.
```
Timidger: To  work around issues you ran into with this, use this in the match statement instead:

```rust
match {
    ...
    ref mut container @ Container::Container { .. } => {
        container.set_layout(layout)?;
    }
}
````

File: src/layout/actions/layout.rs [hunk 2]
```diff
@@ -821,22 +831,35 @@ impl LayoutTree {
...
+                                   ref mut apparent_geometry,
                                    geometry: ref mut actual_geometry,
                                    ref borders, .. } => {
                 *actual_geometry = geometry;
                 if borders.is_some() {
+                    let title_count = match layout {
+                        Layout::Stacked =>
+                            // I don't understand why I have to use this
+                            // instead of just child_count, but seems to work...
+                            (2*child_count-1) as u32,
```
Timidger: This expression is used in a lot of places, is it possible to make this a function somewhere so this piece of code doesn't have to be copied around and then this concept can be given an abstract name to talk about it?

Probably best to place it as a function of `Layout` actually, since it seems dependent on layout (though it's only used for borders, so putting it on borders might make more sense...depends on what you think, I haven't looked into why you need to do that weird double-ing yet. If it's related to borders, put it on borders, if it's a weird quirk of math and how tabbed/stacked works then put it in `Layout` since it's more generic)

src/layout/actions/layout.rs
```diff
@@ -901,18 +924,41 @@ impl LayoutTree {
...
                             borders.as_mut().map(|b| {
+
+                                b.set_children(titles, index.unwrap());
+
+                                // Still necesary to draw the title when this
+                                // container is inside another tabbed/stacked
                                 b.set_title(format!("{:?} ({}/{})",
                                                     layout,
-                                                    index.unwrap_or("?".into()),
+                                                    index_str.unwrap_or("?".into()),
```
Timidger: It's a small thing, but technically the `"?".into()` will always evaluate due to how `unwrap_or` works. If you want to defer it, it can be achieved by putting it in a lambda like so:

```rust
index_str.unwrap_or_else(|| "?".into())
```

It's a little more verbose though, and this is probably not a performance bottle neck so idc too much either way.

src/layout/actions/layout.rs
```diff
@@ -901,18 +924,41 @@ impl LayoutTree {
...
+
             if !self.tree.on_path(child_ix) && Some(child_ix) != self.active_container {
                 self.set_borders(child_ix, borders::Mode::Inactive)?;
             } else {
                 match self.tree[parent_ix] {
                     Container::Container { layout, ref mut borders, .. } => {
                         if layout == Layout::Tabbed || layout == Layout::Stacked {
                             borders.as_mut().map(|b| {
+
+                                b.set_children(titles, index.unwrap());
```
Timidger: Though unlikely, I don't like this `unwrap` here. Can you instead make this a `TreeError` and return early instead?

At that point you won't loop through the rest if there's an error, but it's better than unwrapping.

src/layout/actions/layout.rs
```diff
@@ -901,18 +924,41 @@ impl LayoutTree {
...
+            // I can't do it inside the match ahead because it borrows
+            // self.tree, and I need self.tree to get the children's titles
+            let titles = children.iter().map(|&node_ix| {
+                match self.tree[node_ix] {
+                    Container::Container { ref mut borders, .. } =>
+                        borders.as_ref().map( |b| b.title() )
+                            .unwrap_or("").into(),
+                    Container::View { ref mut borders, .. } =>
+                        borders.as_ref().map( |b| b.title() )
+                            .unwrap_or("").into(),
```
Timidger: Because `borders` is the same for both, you can actually combine these two match statements to make the much more succinct:

```rust
                    Container::Container { ref mut borders, .. } |
                    Container::View { ref mut borders, .. } =>
                        borders.as_ref().map( |b| b.title() )
                            .unwrap_or("").into(),
```

src/layout/core/borders/borders.rs
```diff
@@ -16,6 +25,10 @@ pub struct Borders {
     pub title: String,
     /// The surface that contains the bytes we give to wlc to draw.
     surface: ImageSurface,
+    /// Children titles to be used tabbed/stacked layouts
+    pub children: Option<Children>,
+    /// Layout of the container, only used if the border is from a container
+    pub layout: Option<Layout>,
```
Timidger: Hmm, with all of these optional fields it might be prudent to split this up into a separate type of borders struct (e.g `ContainerBorders` or `ViewBorders`).

I don't know how annoying that would be though, especially as they would probably share a _lot_ of code (and there's no real notion of inheritance in Rust like you see in more traditional OO languages obviously).

You could probably make it work with an enum though, where only some of the functions split based on the type and other wise just do the same thing. It would increase the size of the borders struct slightly because there needs to be a tag, so you wouldn't get the size benifits of splitting them up but it would be easier to follow the logic and harder to mess up (and create another layer of abstraction, e.g borders for containers vs borders for views instead of the current setup where it's just borders that are sometimes used for one or the other).

If this is a lot of work, we can push this to a future patch/issue.

src/layout/core/borders/borders.rs
```diff
@@ -96,7 +111,16 @@ impl Renderable for Borders {
     fn reallocate_buffer(mut self, mut geometry: Geometry) -> Option<Self>{
         // Add the thickness to the geometry.
         let thickness = Borders::thickness();
-        let title_size = Borders::title_bar_size();
+
+        let title_count = match (self.layout, &self.children) {
+            (Some(Layout::Stacked), &Some(ref children)) =>
```
Timidger: You can remove this slightly perplexing looking `&` on the option to avoid moving the value by doing this (and removing the ref too as an added bonus):

```rust
        let title_count = match (self.layout, self.children.as_ref()) {
            (Some(Layout::Stacked), Some(children)) =>
```

src/layout/core/borders/borders.rs
```diff
@@ -345,6 +369,13 @@ impl Borders {
     pub fn set_title(&mut self, title: String) {
         self.title = title;
     }
+
+    pub fn set_children(&mut self, titles: Vec<String>, index: usize) {
+        self.children = Some(Children{
```
Timidger: Put an assert in here to verify that the `index` is a valid index for the vector.

Also add that in the doc comment, so it's obvious to the user of the function to do that.

src/layout/core/borders/container_draw.rs
```diff
@@ -33,21 +33,96 @@ impl ContainerDraw {
...
-
-        // Draw title text
-        self.base.move_to(title_x, title_y);
-        self.base = try!(self.base.check_cairo());
-        self.base.set_color_source(title_font_color);
-        self.base = try!(self.base.check_cairo());
-        self.base.show_text(title.as_str());
-        self.base = try!(self.base.check_cairo());
+        // TODO: This shouldn't be clone for performance reasons, but self
+        // needs to not be borrowed later with all the paint operations, and
```
Timidger: Hmm, yes this is tricky, but I think can see a way around it:

All you care about is which index is the active one, so if `childrens` exist[1] you just loop through and get the index that is the active one (which will copy out that little `usize`) and use that in the match statements. No more needless copying and now your match statements have less duplicate code as a bonus.


[1] and it shouldn't unless there's a case where a container can be not tabbed/stacked and still have children, in which case this is unnecessary useless work. Though you can always hide it behind a lambda that caches, though that makes it so much more complicated it wouldn't be worth it and you should instead just do the needless calculation

src/layout/core/borders/container_draw.rs
```diff
@@ -33,21 +33,96 @@ impl ContainerDraw {
...
+                for (i, title) in children.titles.iter().enumerate() {
+                    let active = children.index == i;
+
+                    let title_color = if active { title_color }
+                        else { Borders::default_title_color() };
+
+                    let title_font_color = if active { title_font_color }
+                        else { Borders::default_title_font_color() };
+
+                    let x = x + tab_width*i as f64;
```
Timidger: Hmm, it took me a couple of moments to realize you are rebinding due to the loop and you need the previous `x` value.

I would suggest maybe trying to make a different name, as this looks confusing and very easy to introduce a bug by removing this "useless" looking rebinding.

If you did this by copying me, then this was a bad mistake I have done and I will endeavour not to make this same mistake since it's obviously rather confusing.

src/layout/core/borders/container_draw.rs
```diff
@@ -33,21 +33,96 @@ impl ContainerDraw {
...
+                    self.base.move_to(title_x, title_y);
+                    self.base = try!(self.base.check_cairo());
+                    self.base.set_color_source(title_font_color);
+                    self.base = try!(self.base.check_cairo());
+                    self.base.show_text(title.as_str());
+                    self.base = try!(self.base.check_cairo());
+                }
+            }
+            _ => {
+                let title: String = self.inner().title().into();
```
Timidger: This is an unnecessary copy. You get borrowck errors if you just return the `&str` here, but you can avoid them by just changing line 121 to be:

```rust
self.base.show_text(self.inner().title());
```

and not bother binding it before that at all.

Timidger:
Also you should mention this fixes #436 so it's closed once this is merged.

For the other non-i3 parts (e.g client windows still rendered when they don't need to, nested containers being different and weird) other issues can be spun up if you think they should be fixed. I'm happy with this currently, but if there's a want to make this more like i3 I won't object.

Timidger:
Found a _really_ small bug (i.e not a showstopper).

Having three tabbed containers causes the middle one to have a sliver on the left side (next to the title part, just when it should meet with the container on the left) that exposes the background (e.g its left side doesn't extend far enough).

Focusing either it or the one on the left either makes it go away or makes it unnoticeable with the background I use. I.e you have to focus the far right one.

Also noticed it only happens on my bigger, second monitor. So might be only an issue at higher resolutions. If the cause isn't immediately obvious, probably nothing to worry about for this PR.

src/layout/actions/layout.rs
Arnaz87: That's not the problem, a few lines up it says in a comment that set_layout cannot be used because it introduces a bug, and down in `fn set_layout` it's done the same way, so I assumed it can be related to the same bug and I did the same.

src/layout/actions/layout.rs
Arnaz87: Yeah, I actually was going to do that, a method `title_bar_size` for the border that did this calculation, but it's not used everywhere, specifically one place I don't remember, it actually calculated the title bar size as expected, just the size times the child count, and if I replaced that with the weird one it could introduce a bug.

So I didn't know how to name the method or where to put it because I don't understand what it's doing.

src/layout/core/borders/borders.rs
Arnaz87: I feel the same, at this point there are too many differences between the borders of views vs containers and even the borders of each layout, because ideally (going i3) the other layouts shouldn't even draw borders (vertical/horizontal), but I do think this should be part of a future redesign.

src/layout/core/borders/container_draw.rs
Arnaz87: I don't think rebindings are that hard to tell, I did notice yours and didn't have difficulty understanding what they were, and I also use them a lot, but I don't really know if they are generally considered good or bad practice.

Arnaz87:
Yeah I also noticed the seam in the tabbed title bar, it's probably because of integer arithmetic rounding the results.

src/layout/actions/layout.rs
Timidger: Ah, ok then. Yeah keep it the way it is then, I'll see if I can improve the comment explaining what the bug is if you do that then.

src/layout/core/borders/container_draw.rs
Timidger: Ok, if you found it easy to follow for my previous code  it's probably not a big deal. 

Rebinding is generally frowned upon but it is context sensitive (eg I do it quite often when I need to eg turns Rust string to a CString but I want to take in a Rust string for ergonomics. There it's pretty clear, here it's only a little more dubious)

src/layout/actions/layout.rs
Arnaz87: What does it mean when index is None?

Timidger: When the `index` is `None`, that means the object couldn't be found in the list (e.g the node wasn't a child of the parent). Now, technically this is impossible with the code we have now. This is a thing that should never happen (tm). So an unwrap is _ok_, but even for the "impossible" errors I think it's best we handle them gracefully.

If that seems like too much work, doing `expect("The node index was not a child!")` would be better because it at least gives a reason why this occured.

Timidger:
Cool thanks, I'll test these changes when I get home just to be doubly-sure it works before I merge it. Thanks again for your contribution! :)

src/layout/core/borders/container_draw.rs
```diff
@@ -33,21 +33,113 @@ impl ContainerDraw {
...
+        match (self.base.inner().layout, child_data) {
+            (Some(Layout::Tabbed), Some((child_count, active_index))) => {
+
+                let tab_width = w / child_count as f64;
+
+                // Iterate over the indices to not borrow self
+                for i in 0..child_count {
+                    let active = i == active_index;
+
+                    let title = match self.base.inner().children.as_ref() {
```
Arnaz87: I couldn't find out how to do this without copying the String

Timidger: You can get around it by doing this:

```rust
                    self.base.show_text({
                        match self.base.inner().children.as_ref() {
                            Some(children) => &children.titles[i],
                            None => unreachable!()
                        }
                    });
```

Arnaz87: Nice, I tried something around that but didn't borrow check, but this does so i'm doing it

PR #450, opened 2017-12-05
Don't show view title twice for Tabbed/Stacked borders

Timidger:
This completes #439, to make the Tabbed/Stacked borders behave exactly like they do in i3.

Now, the titles of views no longer show up twice in Tabbed/Stacked layouts.

@Arnaz87 visually, is there anything I missed? Combined with #448 and #447 this should make complete the i3-style tiling for Way Cooler.

Example of what Tabbed and Stacked looks like:
proper-i3

### DIFF
--- a/config/init.lua
+++ b/config/init.lua
@@ -19,6 +19,7 @@ way_cooler.windows = {
     size = 0, -- The width of gaps between windows in pixels
   },
   borders = { -- Options for borders
+    root_borders = false, -- Display borders in root containers by default
     size = 20, -- The width of the borders between windows in pixels
     inactive_color = 0x386890, -- Color of the borders for inactive containers
     active_color = 0x57beb9 -- Color of active container borders
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -291,7 +291,8 @@ impl LayoutTree {
     /// This will change the active container, but **not** the active path,
     /// it will remain pointing at the previous parent container.
     pub fn float_container(&mut self, id: Uuid) -> CommandResult {
-        let node_ix = try!(self.tree.lookup_id(id).ok_or(TreeError::NodeNotFound(id)));
+        let node_ix = self.tree.lookup_id(id)
+            .ok_or(TreeError::NodeNotFound(id))?;
         if self.tree.is_root_container(node_ix) {
             return Err(TreeError::InvalidOperationOnRootContainer(id))
         }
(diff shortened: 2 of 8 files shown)


Arnaz87:
Nice!
In i3 the upper border of the view is not visible, i3 doesn't draw them only the title bar, that's the main difference I see. I think this should be configurable.
Also the text gets weird when I change the sizes, but that's a different subject.

Timidger:
Are you talking about the client borders? That isn't controllable by the compositor in Wayland unfortunately

Arnaz87:
No, i'm talking about the way-cooler border, the same border whose size is configurable in init.lua, i3 doesn't display the top border, only left right and bottom borders, the top only has the title bar.

Arnaz87:
Another difference I just noticed is the Horizontal/Vertical container title bar, i3 doesn't show them either, only the views' title bars.

Timidger:
Oh I see.

The top bar will be easy, when the boolean is true we can just also not display the top border either.

The horizontal/vertical one is solved with #451 by making that configurable

Timidger:
Ok pushed some changes that should mean this does the right thing. Feel free to test, as I'm only rather sure there's no bugs rather than definitely sure ;)

Timidger:
Here's an update on what it looks like now:

i3-like

Note that the "horizontal" border will be removable later with #451.

Arnaz87:
They look wrong, the stacked and tabbed containers for some reason give less space to the title bars to draw themselves.

The container at the center is just part of the normal horizontal container, but the tabbed at the left and the stacked at the right cut the title bars at the bottom. The horizontal container's title bar is cut off as well.
The view at the left has the configuration i'm using so this can be replicated. Specifically `way_cooler.windows.borders.size = 10` and `way_cooler.windows.title_bar.size = 20`

Timidger:
Thanks for the bug report @Arnaz87, I'll get that fixed up. My goal is to release 0.8 but not a 0.8.1 and these bug reports help me do that :)

Arnaz87:
What do you mean by not a 0.8.1?

Timidger:
@Arnaz87 Sorry for the delay, the latest patch should fix your issue.

Arnaz87:
Well, the titles are no longer cut, but the sizes are still wrong.

In my configuration, borders have size 10, and title_bars size 20.
waycooler sizes

Notice these things:
- Container titlebar is 15 pixels, instead of 20
- Tabbed titlebar is 15 pixels, instead of 20
- Stacked titlebars are 20 pixels, but the last one is 15 pixels (I tried with more stacked views, but I didn't take the picture)
- View titlebar **is** 20 pixels, but it's drawing the top border, unlike the containers, making it seem like 25 pixels.

I don't know if it's expected or feasible to fix all of this for the issue adressed in this pull request, since calculations of titlebar sizes have been wrong since before, only here they are exacerbated and I'm further shaking the nest.

Timidger:
Yeah lets fix that in a future PR.

Issue #421, opened 2017-10-26
Segfault when running inside awesome


l0b0:
Just built and installed [way-cooler-git](https://aur.archlinux.org/packages/way-cooler-git/), but it fails catastrophically.

```
$ rustc --version
rustc 1.21.0 (3b72af97e 2017-10-09)
```

[pkgbuild.log](https://github.com/way-cooler/way-cooler/files/1417360/pkgbuild.log)
[segfault.log](https://github.com/way-cooler/way-cooler/files/1417361/segfault.log)

Timidger:
Can you get a core dump / back trace from the segfault?

l0b0:
```
$ coredumpctl dump --output=way-cooler.core /usr/bin/way-cooler 26053
           PID: 26053 (way-cooler)
           UID: 1000 (victor)
           GID: 100 (users)
        Signal: 11 (SEGV)
     Timestamp: Fri 2017-10-27 20:19:49 NZDT (6min ago)
  Command Line: way-cooler
    Executable: /usr/bin/way-cooler
 Control Group: /user.slice/user-1000.slice/session-c5.scope
          Unit: session-c5.scope
         Slice: user-1000.slice
       Session: c5
     Owner UID: 1000 (victor)
       Boot ID: f0128ca74eff441e9fa52ee5f3e1b72b
    Machine ID: 9d30538abea1471ca2f5225e78cc05c2
      Hostname: l0b0k
       Storage: /var/lib/systemd/coredump/core.way-cooler.1000.f0128ca74eff441e9fa52ee5f3e1b72b.26053.1509088789000000.lz4
       Message: Process 26053 (way-cooler) of user 1000 dumped core.
                
                Stack trace of thread 26053:
                #0  0x00007f8516da4137 wl_list_remove (libwayland-server.so.0)
                #1  0x00007f8518c3ce99 n/a (libwlc.so.0)
                #2  0x00007f8518c1ea47 n/a (libwlc.so.0)
                #3  0x00007f8518c3a52c n/a (libwlc.so.0)
                #4  0x00007f8518c3ade6 n/a (libwlc.so.0)
                #5  0x00007f8516d9fb82 n/a (libwayland-server.so.0)
                #6  0x00007f8516d9fce1 wl_client_destroy (libwayland-server.so.0)
                #7  0x00007f8516d9fdb9 n/a (libwayland-server.so.0)
[...]

Issue #369, opened 2017-08-15
Raspberry Pi support


Timidger:
See [this sway issue for more details](https://github.com/SirCmpwn/sway/pull/1327)

Timidger:
Note that there might also be incorrectly defined types in rust-wlc, so there might need to be changes made to that. It will be obvious if there's problems, because the compiler will complain

Timidger:
This is a nice to have, but if this is too much work I will not include this in the 0.7 release.

Timidger:
Dupe of #445

Timidger:
Actually, there's explicit things we need to check for raspi...

PR #452, opened 2017-12-06
Update wayland-rs to 0.12

### LINKED ISSUE #395: Way Cooler can't find libwayland-server.so.0
On Ubuntu 17.04, for whatever reason, `libwayland-server.so.0` and `libwayland-server.so.0.1.0` exist, but `libwayland-server.so` does not exist, and `libloading` (the library that ultimately searches for the library) can not find `libwayland-server.so.0` from the given library name `libwayland-server.so`.

I was able to fix this issue on my end by making `libwayland-server.so` symlinked to `libwayland-server.so.0` however it might be better to look for `libwayland-server.so.0` instead.

Timidger:
Had some breaking changes, but nothing too big. Will leave open for a while to test any possible regressions with redshift.

Fixes #395

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,11 @@ version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "bitflags"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "c_vec"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


asymmetric:
When can we expect a new release including this change?

Timidger:
Soon™.

To be more precise, when all of these are marked closed I'll cut a release: https://github.com/way-cooler/way-cooler/milestone/18. I already have the changelog written I'm just waiting on the actual code to drop in master.

Note that most of these have PRs open already that I either need to review or I'm waiting on changes for. 

The only thing that doesn't have a PR open is to fix #421 (which I can't reproduce so I'll probably just remove from the release) and #369 (which I'm going to work on in a few minutes).

If the PRs that I request changes from aren't fixed by tomorrow I'l fix them myself and merge them. I'll cut a release at that point (so probably Wed, to answer your question).

Note that after 0.7 drops I want to focus on wlroots-rs so if there are any bugs they will either have to wait to be fixed (unless they are so bad they affect me being able to work in Way Cooler) or they'll need someone other than myself to submit a patch :).

PR #453, opened 2017-12-09
Updated to new mouse movement bindings in rust-wlc.

Idleless:
Hacked together, should be redone by someone who knows what they are
doing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "way-cooler"
 description = "Customizeable Wayland compositor written in Rust"
-version = "0.6.2"
+version = "0.7.0"
 repository = "https://github.com/Immington-Industries/way-cooler/"
 keywords = ["Wayland", "compositor", "window", "manager", "wlc"]
 readme = "README.md"
@@ -10,7 +10,7 @@ authors = ["Snirk Immington <snirk.immington@gmail.com>", "Timidger <apragmaticp
 build = "build.rs"
 
 [dependencies]
-rustwlc = { version = "0.6.4", features = ["wlc-wayland"] }
+rustwlc = { version = "0.7.0", features = ["wlc-wayland"], git = "https://github.com/Idleless/rust-wlc" }
 lazy_static = "0.2"
 log = "0.3"
 env_logger = "0.3"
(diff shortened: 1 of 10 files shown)


Idleless:
I didn't notice the dummy-rustwlc dependency. Unfortunately I have to work in a bit so I probably wont be able to fix that till Monday.

Timidger:
No worries I can fix the dummy-rustwlc dependency. It's literally just requires a mock for the functions you wrote.

src/awesome/mouse.rs
```diff
@@ -67,15 +67,15 @@ fn coords<'lua>(lua: &'lua Lua, (coords, _ignore_enter): (rlua::Value<'lua>, rlu
...
         },
         _ => {
             // get the coords
             let coords = lua.create_table();
-            let Point { x, y } = input::pointer::get_position();
-            coords.set("x", x)?;
-            coords.set("y", y)?;
+            let (x, y) = input::pointer::get_position_v2();
+            coords.set("x", x as i32)?;
+            coords.set("y", y as i32)?;
```
Timidger: Hm, according to the [awesome docs](https://awesomewm.org/doc/api/libraries/mouse.html) it uses integers mainly. However this is more "correct" with actually doing what the user wants...

I would say set this as an f64 instead.

Also right now your doing half of each: You return an integer but you (can) take in a floating number. It should return a floating point number.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ authors = ["Snirk Immington <snirk.immington@gmail.com>", "Timidger <apragmaticp
 build = "build.rs"
 
 [dependencies]
-rustwlc = { version = "0.6.4", features = ["wlc-wayland"] }
+rustwlc = { version = "0.7.0", features = ["wlc-wayland"], git = "https://github.com/Idleless/rust-wlc" }
```
Timidger: I have published this version now on crates.io, please remove this git link.

Timidger: I have also updated dummy-rustwlc. Please update the `Cargo.toml` listing to point to version 0.7.0 for dummy-rustwlc and the build should now pass.

Timidger:
@Idleless bump? I want to include this in the 0.7 release and everything else is pretty much done. If you can't get this done I can make the necessary changes but I can't do that without closing this PR and creating a new one (can't push to your branch)

Timidger:
See #464 for the continuation of this PR

PR #455, opened 2017-12-10
fix crash on layout changed when have floating window.

sbwtw:
The active window (variable `index` in layout.rs) is calculate without the floating filter.
when some floating window exist. the index may out of `titles.len()`

BTW. another bug is the active index of `Container` can be a floating window when switch active window by shortcut key. Maybe is same reason that not filtered the floating ones.

### DIFF
--- a/src/layout/actions/layout.rs
+++ b/src/layout/actions/layout.rs
@@ -923,21 +923,19 @@ impl LayoutTree {
 
             let parent_ix = self.tree.parent_of(child_ix)
                 .expect("Node had no parent");
-            let children = self.tree.children_of(parent_ix);
-            let index = children.iter().position(|&node_ix| node_ix == child_ix);
+            let grounded_children = self.tree.grounded_children(parent_ix);
+
+            let index = grounded_children.iter().position(|&node_ix| node_ix == child_ix);
             let index_str = index.map(|num| (num + 1).to_string());
 
             // I can't do it inside the match ahead because it borrows
             // self.tree, and I need self.tree to get the children's titles
-            let titles = children.iter()
-                .filter(|&&node_ix| {
-                    !self.tree[node_ix].floating()
-                })
+            let titles = grounded_children.iter()
                 .map(|&node_ix| {
                     match self.tree[node_ix] {
                         Container::Container { ref borders, .. } |
                         Container::View { ref borders, .. } =>
-                            borders.as_ref().map( |b| b.title() )
+                            borders.as_ref().map(|b| b.title())
                             .unwrap_or("").into(),
                         ref child => {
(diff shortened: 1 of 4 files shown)


src/layout/actions/layout.rs
```diff
@@ -924,27 +924,32 @@ impl LayoutTree {
...
             // self.tree, and I need self.tree to get the children's titles
-            let titles = children.iter()
-                .filter(|&&node_ix| {
-                    !self.tree[node_ix].floating()
-                })
-                .map(|&node_ix| {
+            let mut titles = Vec::<String>::with_capacity(children.len());
+            let mut index = None;
+            let mut filtered_idx = -1;
+            for &node_ix in children.iter() {
```
Timidger: I don't really like this iterative style. I find it much harder to read compared to the previous version. It might be more efficient due to not looping through the list twice, but the list will always be small and it will be easier to read that way.

One thing that will help: change children creation on line 926 to use `self.tree.grounded_children`, I'm not sure why I didn't switch to that in the first place to be honest. That will remove the need for the filter.

src/layout/actions/layout.rs
```diff
@@ -924,27 +924,32 @@ impl LayoutTree {
...
                         ref child => {
                             error!("Container child is {:#?}", child);
                             panic!("Container childr is not a Container nor a View")
                         }
                     }
-                }).collect();
+                );
+
+                filtered_idx += 1;
+                if node_ix == child_ix { index = Some(filtered_idx as usize); }
```
Timidger: It looks like you're just getting the index in the list. Again, I don't like this iterative solution so this probably won't matter but for the future you can use `enumerate` to have the index passed to you (like it works in Python, for example).

src/layout/core/borders/borders.rs
```diff
@@ -370,8 +370,8 @@ impl Borders {
         self.title = title;
     }
 
-    pub fn set_children(&mut self, titles: Vec<String>, index: usize) {
-        assert!(index < titles.len());
+    pub fn set_children(&mut self, titles: Vec<String>, index: Option<usize>) {
+        // assert!(index < titles.len());
```
Timidger: Go ahead and delete this line completely .

Timidger:
Thanks for your contribution! There's a few issues I have with the structure of the code, but thanks for the fix! I had noticed a bug, but I wasn't sure what caused it. Thanks for fixing it so fast that I didn't even have time to make an issue :).

sbwtw:
Hi @Timidger .
I updated the commit and using `grounded_children` instead of previous `children`.
This problem appears because I add input method `fcitx` into my auto startup programs. and that will create a floating window

PR #456, opened 2017-12-10
Awesome wm/awesome and class fixups drawin impl

Timidger:
This has some crucial changes needed for AwesomeWM compatibility. There are also some additions for drawin, drawable, and others etc. but they are incomplete. I'm merging them anyways so I don't have to duplicate the effort later but those APIs should still be considered incomplete.

With these changes the `rc.lua` no longer properly runs (it fails trying to access a layoutbox property). I will attempt to fix that however because I want the master `rc.lua` always to be usable for testing purposes.

### DIFF
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -28,6 +28,9 @@ if awesome.startup_errors then
                      title = "Oops, there were errors during startup!",
                      text = awesome.startup_errors })
 end
+naughty.notify({ preset = naughty.config.presets.critical,
+                 title = "Oops, there were errors during startup!",
+                 text = awesome.startup_errors })
 
 -- Handle runtime errors after startup
 do
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -28,9 +28,6 @@ if awesome.startup_errors then
                      title = "Oops, there were errors during startup!",
                      text = awesome.startup_errors })
 end
-naughty.notify({ preset = naughty.config.presets.critical,
-                 title = "Oops, there were errors during startup!",
-                 text = awesome.startup_errors })
 
 -- Handle runtime errors after startup
 do
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -94,7 +94,29 @@ fn register_xproperty<'lua>(_: &'lua Lua, _: Value<'lua>) -> rlua::Result<()> {
 
 /// Get layout short names
 fn xkb_get_group_names<'lua>(_: &'lua Lua, _: ()) -> rlua::Result<String> {
-    use std::process::{Stdio, Command};
+    use xcb::{xkb, Connection};
+    use std::os::unix::io::AsRawFd;
+    // TODO Init somewhere else
+    let con = Connection::connect(None/*Some("wayland-0")*/).unwrap().0;
+    // Tell xcb we are using the xkb extension
+    {
+        let cookie = xkb::use_extension(&con, 1, 0);
+
+        match cookie.get_reply() {
+            Ok(r) => {
+                if !r.supported() {
+                    panic!("xkb-1.0 is not supported");
+                }
+            },
+            Err(_) => {
+                panic!("could not get xkb extension supported version");
+            }
+        };
+    }
+    let id = xkb::ID_USE_CORE_KBD as _;
+    let names_cookie = xkb::get_names_unchecked(&con, id, xkb::NAME_DETAIL_SYMBOLS);
+    let names_r = names_cookie.get_reply().unwrap();
+    /*use std::process::{Stdio, Command};
     // TODO FIXME wow this is _really_ _really_ dumb.
     // like really dumb
     // but not sure how else to do this to communicate to xwayland...
@@ -121,6 +143,8 @@ fn xkb_get_group_names<'lua>(_: &'lua Lua, _: ()) -> rlua::Result<String> {
             warn!("err: {:#?}", err);
             Some("pc+us+inet(evdev)".into())
         }).unwrap())
+        */
+    Ok("pc+us+inet(evdev)".into())
 }
 
 /// Restart Awesome by restarting the Lua thread
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -93,9 +93,11 @@ fn register_xproperty<'lua>(_: &'lua Lua, _: Value<'lua>) -> rlua::Result<()> {
 }
 
 /// Get layout short names
-fn xkb_get_group_names<'lua>(_: &'lua Lua, _: ()) -> rlua::Result<()> {
+fn xkb_get_group_names<'lua>(_: &'lua Lua, _: ()) -> rlua::Result<String> {
     warn!("xkb_get_group_names not supported");
-    Ok(())
+    warn!("Returning dummy data so I can move forward");
+    warn!("Replace me before release!");
+    Ok("pc+us+inet(evdev)".into())
 }
 
 /// Restart Awesome by restarting the Lua thread
@@ -169,9 +171,11 @@ fn xkb_set_layout_group(_: &Lua, _group: i32) -> rlua::Result<()> {
     Ok(())
 }
 
-fn xkb_get_layout_group(_: &Lua, _group: i32) -> rlua::Result<()> {
+fn xkb_get_layout_group(_: &Lua, _: ()) -> rlua::Result<i32> {
     warn!("xkb_get_layout_group not supported; Wait until wlroots");
-    Ok(())
+    warn!("Returning dummy data so I can move forward");
+    warn!("Replace me before release!");
+    Ok(0)
 }
 
 fn xrdb_get_value(lua: &Lua, (_resource_class, resource_name): (String, String))
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -1,7 +1,8 @@
 //! TODO Fill in
 
 use std::{ptr, mem};
-use xcb::{xkb, ffi, Connection};
+use xcb::{xkb, Connection};
+use xcb::ffi::{self, xproto};
 use nix::{self, libc};
 use render;
 use std::ffi::{CStr, CString};
@@ -13,6 +14,11 @@ use std::thread;
 use std::default::Default;
 use rlua::{self, Table, Lua, UserData, ToLua, Value, LightUserData};
 use super::{XCB_CONNECTION_HANDLE, signal};
+use super::xproperty::{XProperty, XPropertyType, PROPERTIES};
+
+// TODO FIXME
+// Often we are getting some raw pointers from the xcb replies
+// we need to free them because the memory management for them is manual.
 
 // TODO this isn't defined in the xcb crate, even though it should be.
 // A patch should be open adding this to its generation scheme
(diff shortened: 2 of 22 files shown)


config/rc.lua
```diff
@@ -28,6 +28,9 @@ if awesome.startup_errors then
                      title = "Oops, there were errors during startup!",
                      text = awesome.startup_errors })
 end
+naughty.notify({ preset = naughty.config.presets.critical,
```
SethBarberee: Did you just duplicate this by accident??

Timidger: I put this in here to test notifications. I was _hoping_ to get them working when I first started on this PR (notice a lot of the changes relate to adding cairo and xcb stuff to awesome bindings). Alas, no such luck. I'll remove this as this is a debug thing I forgot to remove.

I'm going to review this again properly and merge it in soon (tm). There's some changes here that are fairly critical (fix a lot of issues with Awesome classes that will be needed down the line. Plus it shows how to connect and use xcb)

SethBarberee: That would be great! I was going to base the wallpaper work off of this branch anyway!

Timidger: I'll work on cleaning this up and merging it in today.

Timidger:
Currently we are doing the objects allocation wrong here. We are allocating the tables weirdly, and we need to be 1-1 on what Awesome does. Once that is setup properly, writing the APIs will be easy enough that it can be done in parallel.

cc @psychon We talked a bit about this on the gitter channel, but I know it's hard to read this Rust when you're not familiar with it. Any quick run-down on how the objects (e.g like a `client`) should look on the Lua side? Right now this current PR seems to allocate them fine since we are running it against the default rc.lua and it's accessing the objects fine but it's still almost certainly not correct.

psychon:
> Any quick run-down on how the objects (e.g like a client) should look on the Lua side?

Well... it's a userdata with some object-specific fields, some of them read-only. The general "stuff" is:
- the signal API (`add_signal`, `connect_signal`, `disconnect_signal`, `emit_signal`),
- a `valid` field that is always `true` and becomes `false` after the object's `__gc` metamethod runs (implemented by setting a different metatable, all other indices cause an error to be thrown),
- a `data` table that the C code saves but does not otherwise care about; this is used to e.g. save the `floating` state of a client and was needed so that Lua's garbage collector knows that this table is only kept alive by the client and can be garbage collected together with it (previously some magic with weak tables were used and if you did `awful.client.property.set(c, "foo", c)` this magic would break down and the client object would be un-garbage-collectable).

Each object has a class, e.g. a client belongs to the `client` global exposed by awesome. This has some fields, too:
- the signal API again; one thing that is easy to miss is that all signals on objects are also emitted on its class, e.g. doing `c:emit_signal("foo", true)` calls all callbacks for "foo" on this client with argument `true` and also all callbacks on the `client` global with arguments `c, true`,
- a counter of "how many instances of this class do currently exist?" as `.instances` (this is a function returning the number, I think); this is not used anywhere, so can be left out, but it still exists and is sometimes useful in debugging leaks ("oh, the number of live client objects is way too high"),
- A special "callback registering" API `set_index_miss_handler` and `set_newindex_miss_handler`; I think this gets the same arguments as the corresponding metafunctions and is used in Lua to implement some additional magic *on individual objects, not the class itself* (e.g. makes `c.floating` return `c.data.floating`, but is also used to implement read-only properties etc; internally this calls getter and setter functions) (implemented e.g. in `lib/awful/client.lua` by calling `gears.object.properties(client, { getter_class = awful.client.object, setter_class = awful.client.object, getter_fallback = client.property.get, setter_fallback = client.property.set })`.

I'll try to give your rust code a close look later.

src/awesome/drawable.rs
```diff
@@ -1,24 +1,35 @@
...
 
 #[derive(Clone, Debug)]
 pub struct DrawableState {
-    // TODO Fill in
-    dummy: i32
+    // TODO Pixmap?
+    pub surface: Option<ImageSurface>,
+    geo: Geometry,
+    refreshed: bool,
+    // TODO Refresh callback
```
psychon: You are using an ImageSurface which allocates its own memory to draw to; awesome uses a XcbSurface drawing to the pixmap. Thus, you don't need a pixmap member here, it's just the "target" for drawing in awesome.

The refresh callback is just an internal implementation detail. For drawins it copies from the drawable's pixmap to the screen and for titlebars it does basically the same. However, for your case it should be enough to just always copy the contents of the surface to the screen on each frame and basically ignore the refresh function that Lua calls (it is always valid to use the contents of the surface, the refresh callback from Lua is just a "I changed something"-signal to force a repaint; this can of course be optimized later).

src/awesome/drawin.rs
```diff
@@ -71,15 +97,29 @@ fn method_setup<'lua>(lua: &'lua Lua, builder: ClassBuilder<'lua>) -> rlua::Resu
     // TODO Do properly
     use super::dummy;
     builder.method("connect_signal".into(), lua.create_function(dummy))?
-           .method("__call".into(), lua.create_function(dummy_create))
+           .method("__call".into(), lua.create_function(|lua, _: Value| Drawin::new(lua)))
```
psychon: Oh and the global (e.g. `drawin`) when called get a table of properties to be applied, for example `visible = true` (but the table can also contains keys that the C code does not understand and thus ignores). This code looks like you explicitly ignore these extra arguments.

src/awesome/awesome.rs
```diff
@@ -94,10 +94,33 @@ fn register_xproperty<'lua>(_: &'lua Lua, _: Value<'lua>) -> rlua::Result<()> {
...
+                        let last = line.split(" ").last().unwrap();
+                        return Some(last.trim().into())
+                    }
+                }
+                None
+            })
+        }).unwrap_or_else(|err| {
+            warn!("err: {:#?}", err);
+            Some("pc+us+inet(evdev)".into())
+        }).unwrap())
```
psychon: I liked your previous "just return static fake data" version of this more. :-)
On a more serious note, I would expect a wayland compositor to already have the required data here somewhere. From a quick look at wlroots, this data seems to be [in struct roots_keyboard_config](https://github.com/swaywm/wlroots/blob/a081b3bf7b99c0e044da26cd00fc74e367c66e63/include/rootston/config.h#L41-L45) (which is part of the example compositor, so perhaps wlroots requires the compositor to handle this itself?).

src/awesome/awesome.rs
```diff
@@ -105,8 +105,34 @@ fn property_setup<'lua>(lua: &'lua Lua, awesome_table: &Table<'lua>) -> rlua::Re
 
 /// Registers a new X property
 /// This actually does nothing, since this is Wayland.
-fn register_xproperty<'lua>(_: &'lua Lua, _: Value<'lua>) -> rlua::Result<()> {
-    warn!("register_xproperty not supported");
+fn register_xproperty<'lua>(lua: &'lua Lua, (name, v_type): (String, String)) -> rlua::Result<()> {
```
psychon: I think I would implement this purely in Lua (somehow). In awesome, awesome can be restarted while the X11 server continues running, so there it makes sense to save things in the X11 server. For way-cooler, a restart surely loses all the state (right?), so emulating xproperties in Lua is okay.
What do you think?
(No, I don't know how to actually do this; it's just a thought.)

src/awesome/screen.rs
```diff
@@ -57,4 +60,27 @@ fn method_setup<'lua>(lua: &'lua Lua, builder: ClassBuilder<'lua>) -> rlua::Resu
...
+    let table = lua.create_table();
+    // TODO Make sure I'm doing this right...
+    let screen_table = Screen::new(lua)?.to_lua(lua)?;
+    table.set("screen", screen_table)?;
+    builder.add_to_meta(table)?
+           .property(Property::new("screen".into(),
+                                   // TODO
+                                   Some(lua.create_function(screen_new)),
+                                   Some(lua.create_function(get_visible)),
+                                   Some(lua.create_function(set_visible))))
```
psychon: Huh? Screens don't have a `visible` property and Lua cannot (that easily) create screens.

Timidger:
Thanks for the detailed writeup! I'm going to work on this PR more and get it to be compliant ASAP. I'm mostly trying to focus on wlroots, since that's going to necessitate mostly starting from scratch on the compositor, but this is really important to enable both me and others to work on whatever part of the Awesome API that seems easiest / most interesting to work on.

> a data table that the C code saves but does not otherwise care about; this is used to e.g. save the floating state of a client and was needed so that Lua's garbage collector knows that this table is only kept alive by the client and can be garbage collected together...

Ahhh, ok that was the missing piece in my information. Previously I tried to store these things in the object itself but ran into a lot of issues which prompted the discussion in [this rlua issue thread](https://github.com/chucklefish/rlua/issues/41). Here's the relevant takeaway from the thread:

> Just so you know though, this is almost NEVER a good idea. This bindings system is not designed to have Lua handles inside UserData, because Lua itself cannot garbage collect through a rust UserData, so you can very easily end up with unbreakable cycles.

src/awesome/awesome.rs
Timidger: In wlc (the current compositor framework Way Cooler uses) that information is not exposed (or at least, not anywhere I can see). It, along with xwayland stuff, is all handled internally.

When we switch to wlroots though that will definitely be what we do. Also yes you're correct we have to handle it ourselves--that's the main benefit of wlroots actually is that we have to manage more stuff (compared to wlc). This increases the complexity of the compositor but also gives enough control that we can do stuff like this.

src/awesome/awesome.rs
Timidger: We may still need to implement this actually, because we will be relying on xwayland a lot and, AFAIK, the xwayland server lets you register xproperties to the clients still.

If the main use case is to store state in case of a restart though, that means we may need to do some hoops to achieve this...

You're right in that if way-cooler was to restart (e.g the process died) then we would lose all state because Way Cooler is not only the window manager but also the display server. If the use case is primarily recovery from the process dying (e.g from an error) then we would need to achieve this via some other way.

One thing we _could_ do is run Lua in a totally separate process and communicate between Way Cooler and Lua via IPC. Though that would obviously be:

1) Really slow
2) Way more complicated

That would achieve the same thing Awesome currently does with X11 (since they are separate processes there) but it would mean more work.

Depending on the use case, this could drastically affect the design. But it shouldn't affect implementing the interface, as I'll be moving things around when we move to wlroots anyways. I'll be thinking about this though, thank you.

src/awesome/screen.rs
Timidger: Whoops, looks like an oversight. I'll remove those.

src/awesome/drawin.rs
Timidger: Nice catch, I'll fix that when I fix the object creation (so I can throw those unknown keys into `data`)

src/awesome/drawable.rs
Timidger: Thanks, I had a feeling I didn't need a pixmap but I hadn't looked into whether or not I needed it.

Ok so I don't even really need to add it as a field, I can just have a refresh function that checks the bool and only draws if it's true then. That will simplify this.

src/awesome/awesome.rs
psychon: Well, the "global state" (xproperties on `root`) is not all that interesting, I think. And xproperties on individual client objects will still be lost since when the display server dies, all clients die with it.

Yes, this xproperty API is from a time when awesome was still restarting whenever the screen configuration changed. Ever since awesome stopped doing that, I haven't seen much use of this API.

src/awesome/drawin.rs
psychon: I don't think awesome stores unknown keys into `data`, it just really completely ignores them.
This seems to be done in `luaA_class_new` in `common/luaclass.c`: All strings for which it finds a `lua_class_property_t` are handled, all others are just silently ignored.
(This is like this mostly for "historic reasons", I guess; I don't know if it makes sense or not)

config/rc.lua
Timidger: Ok "today" was a bit optimistic. Going to try to finish this up ASAP, but you should still base your work off this branch.

Timidger:
Ok I think I fixed all the issues I had with this PR. Still some rough edges around the core object stuff, but nothing blocking [working on the other issues](https://github.com/way-cooler/way-cooler/issues?q=is%3Aissue+is%3Aopen+label%3AAwesome). Gonna give this one more look through and then merge it later today.

PR #468, opened 2018-01-05
Small improvement to drawin_geometry

psychon:
Commit 487b8d6e5 made the argument to the geometry method of a drawin
optional, so that the geometry can be queried without setting it. This
was done by directly matching on the Lua value.

However, rlua implements FromLua for Option<T>, so instead of
implementing this optionality directly, Option can be used instead.


Timidger:
Thanks!

PR #471, opened 2018-01-14
Allow colors to contain alpha values

timo-schmid:
Hi

This patch will allow ARGB values to be defined:

* `0xbbggrr` - Normal (non-transparent) color (still works)
* `0xaabbggrr` - Transparent color

That's also the reason I chose ARGB over RGBA - it makes parsing the value very simple. This way, I can just parse the u32, and don't have to look at the original string (eg. the alpha bits are optional).

Looking at the string would have some advantages however:
* Distinct between formats like: "0xrgb", "0xrrggbb" and "0xaarrggbb"
* Allow different parsing methods for different formats

I think this could be adressed in a later patch, with a notice communicated to the users.

I've noticed, the colors in the hex value are defined as `0xbbggrr` - which is a bit weird - but I didn't want to change it yet, as it breaks backwards-compatibility with old config files (users probably will be confused when their colors suddenly look different). Thus, I haven't changed it. Swapping `red` and `blue` in `from` would easily fix this.

Thank you all for developing this software, its very fun to use and simple to set up!

### DIFF
--- a/src/render/color.rs
+++ b/src/render/color.rs
@@ -3,8 +3,12 @@
 use std::convert::From;
 use std::str::Chars;
 
+const WLC_BUG_INVERT_RED_BLUE: bool = true;
 
 /// Color to draw to the screen, including the alpha channel.
+/// NOTE: At this point, the parsed colors return the colors red and blue switched.
+/// This is due to a bug in WLC, causing the colors to be switched when drawing.
+/// Example: "00FF0000" will draw as red (correct), but the Color structure will contain 0 for `red` and 255 for `blue`.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub struct Color {
     red: u8,
@@ -18,11 +22,24 @@ impl Color {
 
     /// Creates a new color with an alphachannel
     pub fn rgba(r: u8, g: u8, b: u8, a: u8) -> Color {
-        Color {
-            red: r,
-            green: g,
-            blue: b,
-            alpha: a
+	if WLC_BUG_INVERT_RED_BLUE {
+            // There is a bug in wlc, causing red and blue to be inverted:
+            // https://github.com/Cloudef/wlc/issues/142
+            // We can work around it, by just switching red with blue, until the issue is resolved.
+            // When the bug is fixed, this code path can be deleted and the tests must be adjusted.
+            Color {
+                red:   b,
+                green: g,
+                blue:  r,
+                alpha: a
+            }
+        } else {
+            Color {
+                red:   r,
+                green: g,
+                blue:  b,
+                alpha: a
+            }
         }
     }
 
@@ -46,7 +63,7 @@ impl Color {
         } else if s.starts_with("0x") {
             Color::parse(s[2..].to_string())
         } else if s.len() == 8 {
-            Color::parse_rgba(s)
+            Color::parse_argb(s)
         } else if s.len() == 6 {
             Color::parse_rgb(s)
         } else {
@@ -55,12 +72,18 @@ impl Color {
     }
 
     /// Parses an ARGB String into a Color
-    fn parse_rgba(s: String) -> Option<Color> {
+    fn parse_argb(s: String) -> Option<Color> {
         if s.len() == 8 {
             let mut chars = s.chars();
-            Color::parse_color(&mut chars)
-                .and_then(|a| Color::parse_rgb(chars.as_str().to_string())
-                    .map(|rgb| Color::rgba(rgb.red, rgb.green, rgb.blue, a)))
+            let alpha = Color::parse_color(&mut chars);
+            let color = Color::parse_rgb(chars.as_str().to_string());
+            if WLC_BUG_INVERT_RED_BLUE {
+		// Due to the bug, the colors are already inverted, so in the returned color
+                // red is blue and blue is red.
+                alpha.and_then(|a| color.map(|rgb| Color::rgba(rgb.blue, rgb.green, rgb.red, a)))
+            } else {
+                alpha.and_then(|a| color.map(|rgb| Color::rgba(rgb.red, rgb.green, rgb.blue, a)))
+            }
         } else {
             None
         }
@@ -70,10 +93,10 @@ impl Color {
     fn parse_rgb(s: String) -> Option<Color> {
         if s.len() == 6 {
             let mut chars = s.chars();
-            Color::parse_color(&mut chars)
-                .and_then(|r| Color::parse_color(&mut chars)
-                    .and_then(|g| Color::parse_color(&mut chars)
-                        .map(|b| Color::rgba(r, g, b, 255))))
+            let red   = Color::parse_color(&mut chars);
+            let green = Color::parse_color(&mut chars);
+            let blue  = Color::parse_color(&mut chars);
+            red.and_then(|r| green.and_then(|g| blue.map(|b| Color::rgba(r, g, b, 255))))
         } else {
             None
         }
@@ -81,41 +104,28 @@ impl Color {
 
     /// Parses exactly one single color value from a String (eg "AA", "RR", "GG" or "BB")
     fn parse_color(chars: &mut Chars) -> Option<u8> {
-        chars.next().and_then(Color::hex_to_num)
-            .and_then(|i1| chars.next()
-                .and_then(Color::hex_to_num).map(|i2| (i1 << 4) | i2))
+        let mut next_two_chars = chars.take(2);
+        let digit1 = Color::parse_next_char(next_two_chars.next());
+        let digit2 = Color::parse_next_char(next_two_chars.next());
+        digit1.and_then(|i1| digit2.map(|i2| (i1 << 4) | i2))
     }
 
-    /// Converts a hex char into a number
-    fn hex_to_num(c: char) -> Option<u8>{
-        match c {
-            '0' => Some(0),
-            '1' => Some(1),
-            '2' => Some(2),
-            '3' => Some(3),
-            '4' => Some(4),
-            '5' => Some(5),
-            '6' => Some(6),
-            '7' => Some(7),
-            '8' => Some(8),
-            '9' => Some(9),
-            'a' | 'A' => Some(10),
-            'b' | 'B' => Some(11),
-            'c' | 'C' => Some(12),
-            'd' | 'D' => Some(13),
-            'e' | 'E' => Some(14),
-            'f' | 'F' => Some(15),
-             _   => None
-         }
+    fn parse_next_char(c: Option<char>) -> Option<u8> {
+        c.and_then(Color::hex_to_u8)
+    }
+
+    /// Converts a hex char into a u8
+    fn hex_to_u8(c: char) -> Option<u8> {
+        c.to_digit(16).map(|x| (x as u8))
     }
 
 }
 
 impl From<u32> for Color {
     fn from(val: u32) -> Self {
-        let blue = ((val & 0xff0000) >> 16) as u8;
-        let green = ((val & 0x00ff00) >> 8) as u8;
-        let red = (val & 0x0000ff) as u8;
+        let red   = ((val & 0xff0000) >> 16) as u8;
+        let green = ((val & 0x00ff00) >> 8)  as u8;
+        let blue  =  (val & 0x0000ff)        as u8;
         Color::rgba(red, green, blue, 255)
     }
 }
@@ -125,29 +135,7 @@ mod test {
     use ::render::Color;
 
     #[test]
-    fn test_old_logic_failure() {
-        let hex_red   = 0xFF0000;
-	let hex_green = 0x00FF00;
-        let hex_blue  = 0x0000FF;
-	let r: Color = hex_red.into();
-	let g: Color = hex_green.into();
-	let b: Color = hex_blue.into();
-	// test red values
-	assert_eq!(0xFF, r.red);
-	assert_eq!(0x00, r.green);
-	assert_eq!(0x00, r.blue);
-	// test green values
-	assert_eq!(0x00, g.red);
-	assert_eq!(0xFF, g.green);
-	assert_eq!(0x00, g.blue);
-	// test blue values
-	assert_eq!(0x00, b.red);
-	assert_eq!(0x00, b.green);
-	assert_eq!(0xFF, b.blue);
-    }
-
-    #[test]
-    fn test_old_logic_working() {
+    fn test_from_u32() {
         let hex_red   = 0xFF0000;
 	let hex_green = 0x00FF00;
         let hex_blue  = 0x0000FF;
@@ -217,9 +205,9 @@ mod test {
         assert_eq!(0,   rgb_black.blue);
         assert_eq!(255, rgb_black.alpha);
         let rgb_red   = Color::parse_rgb("ff0000".to_string()).unwrap();
-        assert_eq!(255, rgb_red.red);
+        assert_eq!(0,   rgb_red.red);
         assert_eq!(0,   rgb_red.green);
-        assert_eq!(0,   rgb_red.blue);
+        assert_eq!(255, rgb_red.blue);
         assert_eq!(255, rgb_red.alpha);
         let rgb_green = Color::parse_rgb("00ff00".to_string()).unwrap();
         assert_eq!(0,   rgb_green.red);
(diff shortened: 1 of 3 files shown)


Timidger:
Just going to test this out to ensure there's no regressions. Looks great though, thanks!

As for addressing the bizarre way colours are written in the configuration: Since the configuration file is going to change _very_ drastically in v0.9.0 or so, I'm open to either not changing it (and just waiting for this particular thing to be fixed with the whole move to AwesomeWM compatibility) or fixing it and releasing a v0.7.1. It's _kinda of_ a breaking change, but since it doesn't actually break anything, and this is still alpha software, I think a minor version bump is fine.

Timidger:
Testing it out, I noticed that for the borders the colour is not the same (so unlike you said in your comment, this changes how the colours are treated in the configuration file).

This patch should fix it so that old colours still work, but if you use the alpha channel then it uses the new format:

```diff
diff --git a/src/render/color.rs b/src/render/color.rs
index a6ea859..5583533 100644
--- a/src/render/color.rs
+++ b/src/render/color.rs
@@ -33,10 +33,22 @@ impl Color {
 
 impl From<u32> for Color {
     fn from(val: u32) -> Self {
-        let blue  =  (val & 0x000000ff) as u8;
-        let green = ((val & 0x0000ff00) >> 8) as u8;
-        let red   = ((val & 0x00ff0000) >> 16) as u8;
-        let alpha = ((val & 0xff000000) >> 24) as u8;
+        let blue;
+        let green;
+        let red;
+        let alpha;
+        // If it includes the alpha value
+        if val > 0x000ffffff {
+            blue  =  (val & 0x000000ff) as u8;
+            green = ((val & 0x0000ff00) >> 8) as u8;
+            red   = ((val & 0x00ff0000) >> 16) as u8;
+            alpha = ((val & 0xff000000) >> 24) as u8;
+        } else {
+            blue = ((val & 0xff0000) >> 16) as u8;
+            green = ((val & 0x00ff00) >> 8) as u8;
+            red = (val & 0x0000ff) as u8;
+            alpha = 0;
+        }
         Color::rgba(red, green, blue, (255 - alpha))
     }
 }
```

As nice as backwards compatibility is, I think it's ok if we just go with the new format everywhere. I'll do some more testing to ensure everything has been changed though.

Timidger:
Also it looks like the format is `bbggrr`, which is actually _not_ what it was before. Before it was actually `rrggbb`.

Does it work differently on your machine? I would be surprised if it did. Either way, `rrggbb` probably makes more sense to users.

timo-schmid:
Hmm, let me test again - but I'm fairly sure a hex value of `0xff0000` did not lead to red borders before - I remember being really confused by that. I'll recompile (before my commit) and test in a few minutes or so.

timo-schmid:
It seems you're right - the borders are actually red now using `0xff0000` - it must have been related to a change I made, i'll check it out!

Timidger:
> `0xffffffff` -> non-transparent white, correct
> `0xffffffff` non-transparent white, wrong!

Was there a typo? Those are the same values.

> It seems you're right - the borders are actually red now using 0xff0000 - it must have been related to a change I made, i'll check it out!

Ok so I think there's an endian issue now...because when I test your code that value gives me _blue_ borders, not red. Are you on a big endian machine? Unfortunately all of my machines are little endian, so I haven't tried this on a different machine.

> Another way to achieve this would be just to have "inverted" alpha-values (as it is in my patch, 0 = fully visible, ff = transparent). I have to say I really don't like this solution very much.

yeah `ff` should be fully visible, to fix this you should just be able to remove the `255 - alpha` in your code and it should do the right thing.

> The only way to really nicely solve it, would be to know the string of the config value, as both values 0xffffff and 0x00fffff have the same value, but mean different things. Is there a way to achieve this?

Those don't have the same value though. Right now we don't support colours that don't come in RR,GG,BB (and with this patch, AA, RR, GG, BB) and you gave AA, RR, GG, B. I don't think it's a big deal to force that format, as it makes it easy to parse and also makes it unambiguous to the user.

So as a takeaway, lets see if there is an endianness issue (and if so, it should be fixed) causing us to see different things. To help, here's what `0xff0000` looks like with your patch on my machine for active borders + active border title bars:

aeou

Finally, lets just support the old way RRGGBB and the new way (AARRGGBB). Your code handles this fine sans possible endian issues we might be experiencing.

timo-schmid:
> Was there a typo? Those are the same values.

Yes, actually there was - I meant to write `0xffffff` in the first one.

> Ok so I think there's an endian issue now...because when I test your code that value gives me blue borders, not red. Are you on a big endian machine? Unfortunately all of my machines are little endian, so I haven't tried this on a different machine.

I think there's a misunderstanding - I meant to say: The last commit (before my changes) is fine, borders look red when I use 0xff0000 - then I run the code with my commit applied, and they are blue, so it's just an issue in my code. I don't think it's endian-related (I'm on x68_64). You see the same effect on your end, right?

> yeah ff should be fully visible, to fix this you should just be able to remove the 255 - alpha in your code and it should do the right thing.

The problem I see with this is: When I use a hex number without any alpha values (eg. `0xff0000`) the expected result should be a red border. What will happen is, the bits for Alpha will be set to 00 and thus the border will not be shown. I could add an `if` to avoid this, but that would make full transparency impossible.

Anyway, if we still haven't figured it out tomorrow i'll hop on gitter, seems it's a bit unclear with such long messages here :+1:

Timidger:
> You see the same effect on your end, right?
Yes! Ok good. Started worrying for a second. Excuse my poor reading comprehension skills.

> I could add an if to avoid this, but that would make full transparency impossible.
Ahh, ok I got you.

Hmm...yeah this sounds like an issue. So the "parsing" of those values (as you probably guessed) happens from Lua because it converts the hex number to a regular number. So either we switch to strings (which has been suggested in the past, and I'm open to) like you suggested or we have the backwards way.

We are pre-1.0, so backwards compatibility, while super nice, isn't that important (and we can't be a real Awesome clone if we have good backwards compatibility guarantees now can we 😜). 

I'll enumerate the options here, for posterity and historian's sake, from maximum current compatibility to least:
* Swap the alpha value.
  - I don't like that for the reasons you mentioned, it's very confusing.
* Keep the old way but have an option to take a string that includes the alpha channel.
  - This is an interesting idea, but then it will look confusing to the user and complicate the code by having two paths.
* Take in a string and remove the old way
  - This has the least code complexity changes, and it fixes this bug while enabling simpler / different colour combinations.

I think I like the last way the best, though that will mean I should do a minor, rather than patch, version bump for this. But that's ok, we have plenty of numbers 😉.

Just one note: if you implement this change, this will mean that `wc-bg` will no longer take similar options as the Way Cooler configuration. To fix this is really easy though, you'll just need to open a similar PR [for this file in this repo](https://github.com/way-cooler/way-cooler-client-helpers/blob/master/src/color.rs) and I'll have to publish a new version of `wc-bg` as well.

Timidger:
@timo-schmid Thanks for putting the work in! I'm happy to see it's well tested too 👍 

I have some notes on how it can be improved:
* `hex_to_num` can just use [to_digit](https://doc.rust-lang.org/std/primitive.char.html#method.to_digit)
* Have you tried using [take](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take) to make [parse_color](https://gist.github.com/timo-schmid/f937a892e30d1e8cfdc561e160875349#file-lib-rs-L77) easier to follow? The nested `and_then` makes it a bit difficult to follow that you are taking two at a time.
* For [parse_rgba](https://gist.github.com/timo-schmid/f937a892e30d1e8cfdc561e160875349#file-lib-rs-L49) the doc comment says it parses ARGB.
* [This method](https://gist.github.com/timo-schmid/f937a892e30d1e8cfdc561e160875349#file-lib-rs-L66) gets a bit slanty to the right. Why don't you straighten it out by using `?` to return early if a `None` is returned from one of those calls, especially since you're letting them do the iteration.
* Is it necessary to do an allocation [here](https://gist.github.com/timo-schmid/f937a892e30d1e8cfdc561e160875349#file-lib-rs-L53)? It iterates over the characters from that `String`, so is there an ownership issue I'm not seeing in this code? Actually in fact can't you use `&str` everywhere in this code? You are doing a lot of allocations that aren't necessary.

Those notes are mostly over style/performance. Overall, looks really great though! Thanks for putting the time in to write this boring parsing code, that's the main thing that's put me off doing this for so long. 👍

Timidger:
@timo-schmid Any updates on getting those changes incorporated into this PR?

timo-schmid:
I didn't have much time to work on it last week, but I'll keep you posted with an update soon. Btw, using this code I still have the "inverted red and blue" bug - so I'll try to figure out why that is next.

timo-schmid:
I added two tests to see what the old behaviour exactly does, it seems it wasn't working at all how I expected it:    
```
#[test]
    fn test_old_logic_failure() {
        let hex_red   = 0xFF0000;
        let hex_green = 0x00FF00;
        let hex_blue  = 0x0000FF;
        let r: Color = hex_red.into();
        let g: Color = hex_green.into();
        let b: Color = hex_blue.into();
        // test red values
        assert_eq!(0xFF, r.red);
        assert_eq!(0x00, r.green);
        assert_eq!(0x00, r.blue);
        // test green values
        assert_eq!(0x00, g.red);
        assert_eq!(0xFF, g.green);
        assert_eq!(0x00, g.blue);
        // test blue values
        assert_eq!(0x00, b.red);
        assert_eq!(0x00, b.green);
        assert_eq!(0xFF, b.blue);
    }
    #[test]
    fn test_old_logic_working() {
        let hex_red   = 0xFF0000;
        let hex_green = 0x00FF00;
        let hex_blue  = 0x0000FF;
        let r: Color = hex_red.into();
        let g: Color = hex_green.into();
        let b: Color = hex_blue.into();
        // test red values
        assert_eq!(0x00, r.red);
        assert_eq!(0x00, r.green);
        assert_eq!(0xFF, r.blue);
        // test green values
        assert_eq!(0x00, g.red);
        assert_eq!(0xFF, g.green);
        assert_eq!(0x00, g.blue);
        // test blue values
        assert_eq!(0xFF, b.red);
        assert_eq!(0x00, b.green);
        assert_eq!(0x00, b.blue);
    }
```

Turns out, ` test_old_logic_failure` does not work, but `test_old_logic_working` runs without errors. So it looks to me, that the values of `red` and `blue` were inversed all along. Can you confirm this on your machine as well?

It gets even more weird - looking at `set_color_source` in `src/render/draw.rs` looks like it's passing the parameters to Cairo's `set_source_rgba` in the correct order - so I'm baffled right now it works the way it does.

Any idea what causes this?

Timidger:
I can confirm that the same happens on my machine, though you're right it's a bit strange...because I try to use the literal values `0xff0000` and `0x0000ff` and they show red and blue respectively...

Oh yeah I thought it had something to do with this...I apologize, I forget wlc had this bug as I haven't touched the color code in a while. [See this issue for the answer to your question](https://github.com/Cloudef/wlc/issues/142). TL;DR: wlc renders it in BGRA format even when you specify to use RGBA. 

Since this is a wlc issue (and will definitely be fixed in wlroots) I suggest changing your colour code to incorrectly parse the values (e.g the old way I do it) and document in it clearly. That will cause the least breakage, including when we switch to the main format. It does mean e.g `color.red` isn't actually red but blue, but there's no where in the code that I access that directly so it doesn't break anything.

Sorry for the confusion.

timo-schmid:
Ah, that explains it - thanks, I probably wouldn't have found that out by mself!

I've fixed most of the suggestions from the code review, thanks for the hints!

> Is it necessary to do an allocation here? It iterates over the characters from that String, so is there an ownership issue I'm not seeing in this code? Actually in fact can't you use &str everywhere in this code? You are doing a lot of allocations that aren't necessary.

This one is still TBD.

src/render/color.rs
```diff
@@ -1,8 +1,14 @@
 //! Colors used for drawing to a Cairo buffer
 
 use std::convert::From;
+use std::str::Chars;
+
+const WLC_BUG_INVERT_RED_BLUE: bool = true;
```
Timidger: I suggest removing this and leaving the comments. Once we switch away from WLC there's no need for this. Also it's unknown if this whole module will survive the trip there, though it probably will. It'll be pretty obvious when it doesn't work thankfully, and I think this boolean check just adds noise.

e.g I never see myself setting this to `false`. I'll just delete the old code.

src/render/color.rs
```diff
@@ -28,13 +48,272 @@ impl Color {
...
+    }
+}
+
+mod test {
+
+    use ::render::Color;
+
+    #[test]
+    fn test_from_u32() {
+        let hex_red   = 0xFF0000;
```
Timidger: use spaces, not tabs.

src/render/color.rs
```diff
@@ -28,13 +48,272 @@ impl Color {
...
-        let green = ((val & 0x00ff00) >> 8) as u8;
-        let red = (val & 0x0000ff) as u8;
-        Color::solid_color(red, green, blue)
+        let red   = ((val & 0xff0000) >> 16) as u8;
+        let green = ((val & 0x00ff00) >> 8)  as u8;
+        let blue  =  (val & 0x0000ff)        as u8;
+        Color::rgba(red, green, blue, 255)
+    }
+}
+
```
Timidger: add `#[cfg(test)]` to make the warning go away on build.

src/render/color.rs
```diff
@@ -28,13 +48,272 @@ impl Color {
...
+    /// Parses an ARGB String into a Color
+    fn parse_argb(s: String) -> Option<Color> {
+        if s.len() == 8 {
+            let mut chars = s.chars();
+            let alpha = Color::parse_color(&mut chars);
+            let color = Color::parse_rgb(chars.as_str().to_string());
+            if WLC_BUG_INVERT_RED_BLUE {
+		// Due to the bug, the colors are already inverted, so in the returned color
+                // red is blue and blue is red.
+                alpha.and_then(|a| color.map(|rgb| Color::rgba(rgb.blue, rgb.green, rgb.red, a)))
```
Timidger: `alpha` can be returned early so you don't need this `and_then` by simply doing above:

```rust
let alpha = Color::parse_color(&mtu chars)?;
```

src/render/color.rs
```diff
@@ -28,13 +48,272 @@ impl Color {
...
+    }
+
+    /// Parses a RGB String into a Color
+    fn parse_rgb(s: String) -> Option<Color> {
+        if s.len() == 6 {
+            let mut chars = s.chars();
+            let red   = Color::parse_color(&mut chars);
+            let green = Color::parse_color(&mut chars);
+            let blue  = Color::parse_color(&mut chars);
+            red.and_then(|r| green.and_then(|g| blue.map(|b| Color::rgba(r, g, b, 255))))
```
Timidger: Can use `?` here too, no need for `and_then`.

src/render/color.rs
```diff
@@ -28,13 +48,272 @@ impl Color {
...
+            None
+        }
+    }
+
+    /// Parses exactly one single color value from a String (eg "AA", "RR", "GG" or "BB")
+    fn parse_color(chars: &mut Chars) -> Option<u8> {
+        let mut next_two_chars = chars.take(2);
+        let digit1 = Color::parse_next_char(next_two_chars.next());
+        let digit2 = Color::parse_next_char(next_two_chars.next());
+        digit1.and_then(|i1| digit2.map(|i2| (i1 << 4) | i2))
```
Timidger: and here too (`?`)

Timidger:
You can definitely make them all take `&str`, just change the type signatures and remove all the `to_string` everywhere.

src/render/color.rs
timo-schmid: Fair point - I'll change it.

src/render/color.rs
Timidger: e.g `s.chars().take(2)?`

src/render/color.rs
timo-schmid: TIL 👍

timo-schmid: This doesn't work for the last line, does it? (eg. I'll still have to map over the last Option)

Timidger: You can do 

```rust
let color = Color::parse_rgb(chars.as_str())?
...
Some(Color::rgba(rgb.blue, rgb.green, rgb.red, a))
```

src/render/color.rs
timo-schmid: It's fixed on the lines below, this line doesn't seem to have tabs.

timo-schmid:
Allright, looks much better now using `&str`. Thanks for all the advices, very useful stuff!

I think I got all the issues you pointed out, let me know if I missed one!

Timidger:
Awesome! Thanks for all your hard work. I'll cut a 0.8 release with these changes tomorrow, unless I can think of some other breaking changes / small bugs that annoy me that I can put in that release too.

PR #472, opened 2018-01-21
WIP: Run a Glib main loop (GMainLoop)

psychon:
GLib provides a default GMainContext (`g_main_context_default`). Awesome's Lua code accesses this directly through LGI for things, e.g. `gears.timer.start_new`. However, this only works if a GMainLoop is actually running for the default context. Thus, if way-cooler wants to be compatible with awesome's Lua code, it needs to run a GMainLoop.

Lua by itself is not thread-safe. Thus, LGI implements locking and automatically locks/unlock stuff at the C/Lua transition. However, LGI cannot detect the Lua->C transition that happens when way-cooler loads the `rc.lua` and then at the end continues running. Thus, from LGI's point-of-view, it is still in Lua-land and so its mutex is locked. Thus, LGI must be loaded from the same thread that will also run the main loop (since the mutex is recursive; aka can be locked multiple times by the same thread).

The above considerations resulted in this patch series.

- The first commit unexports the LUA global so that it can be enforced that only one thread every runs Lua code.
- The second commit actually changes the API added in the previous commit to switch to the Lua thread. This is a separate commit since I had quite some problems in getting this to work and the resulting code is bad (especially the error handling, aka `unwrap()`)
- The next commit only happened accidentally, because without it the following commit would not work (the main thread tries to `send()` to the Lua thread immediately after starting it, but `send()` does not work yet because `RUNNING` is still `false` => panic). This is just a (bad) work-around which a later commit (silently) reverts.
- Next, Lua initialisation is actually moved to the thread running Lua code.
- The following commit switches to using `GMainLoop`. This also reverts the removal of `RUNNING` and instead removes `SENDER` (since `std::sync::mpsc` cannot easily be polled inside a `GMainLoop`).
- The last commit happened because I wanted to get rid of the LUA global. It is only accessed from the Lua thread and so I thought I could get away with a a `thread_local!`. Well... some banging my head against `rustc` later I gave up and instead "only" moved the `LUA` global into the only module using it.

This PR was only lightly tested, mainly by "does `require("gears.timer").start_new(function() print("I live!") return true end)` work?" and by "is there a deadlock?" (I'm actually surprised by how many deadlocks I encountered along the way and how long it took me to figure out what I just wrote in the second paragraph of this text about "LGI's lock is never unlocked"). Also, the resulting code feels "wrong/bad" to me. I'd be happy for suggestions on how to improve this.

If you want to, you can merge this (and just commit follow up commits to make the code less ugly (`run_with_lua`, the introduction of a new (untested) `ExecWithLua`, ...)). However, I feel bad for even proposing this and am quite unhappy with some of the ugliness lurking in here.

NOTE: While writing the above it occurred to me: I'm not actually sure if `send()` is still order-preserving, aka earlier `send()`s are handled before later ones. I guess it is, but if this might be a problem, please tell me to double-check.

NOTE: Calling `g_source_attach` from other threads is actually documented as being safe:

>  A GMainContext can only be running in a single thread, but sources can be added to it and removed from it from other threads.

https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#glib-The-Main-Event-Loop.description

Edit: Some docs on LGI's mutex: https://github.com/pavouk/lgi/blob/master/docs/guide.md#6-threading-and-synchronization

### DIFF
--- a/src/awesome/keygrabber.rs
+++ b/src/awesome/keygrabber.rs
@@ -1,6 +1,6 @@
 //! AwesomeWM Keygrabber interface
 
-use ::lua::LUA;
+use ::lua::run_with_lua;
 use rlua::{self, Lua, Table, Function, Value};
 use rustwlc::*;
 #[allow(deprecated)]
@@ -29,26 +29,24 @@ pub fn init(lua: &Lua) -> rlua::Result<()> {
 /// defined with the input.
 pub fn keygrabber_handle(mods: KeyboardModifiers, sym: Keysym, state: KeyState)
                          -> rlua::Result<()> {
-    let lua = LUA.lock()
-        .map_err(|err| rlua::Error::RuntimeError(
-            format!("Lua lock was poisoned {:#?}", err)))?;
-    let lua_state = if state == KeyState::Pressed {
-        "press"
-    } else {
-        "release"
-    }.into();
-    let lua_sym = sym.get_name()
-        .ok_or_else(|| rlua::Error::RuntimeError(
-                 format!("Symbol did not have a name: {:#?}", sym)))?;
-    let lua_mods = ::lua::mods_to_lua(&lua.0, mods.mods)?;
-    let res = call_keygrabber(&lua.0, (lua_mods, lua_sym, lua_state));
-    match res {
-        Ok(_) | Err(rlua::Error::FromLuaConversionError { .. }) => {Ok(())},
-        err => {
-            err
+    run_with_lua(move |lua| {
+        let lua_state = if state == KeyState::Pressed {
+            "press"
+        } else {
+            "release"
+        }.into();
+        let lua_sym = sym.get_name()
+            .ok_or_else(|| rlua::Error::RuntimeError(
+                     format!("Symbol did not have a name: {:#?}", sym)))?;
+        let lua_mods = ::lua::mods_to_lua(lua, mods.mods)?;
+        let res = call_keygrabber(lua, (lua_mods, lua_sym, lua_state));
+        match res {
+            Ok(_) | Err(rlua::Error::FromLuaConversionError { .. }) => {Ok(())},
+            err => {
+                err
+            }
         }
-    }
-
+    })
 }
 
 /// Call the Lua callback function for when a key is pressed.
--- a/src/awesome/mousegrabber.rs
+++ b/src/awesome/mousegrabber.rs
@@ -1,6 +1,6 @@
 //! AwesomeWM Mousegrabber interface
 
-use ::lua::LUA;
+use ::lua::run_with_lua;
 use rlua::{self, Lua, Table, Function, Value};
 use rustwlc::ButtonState;
 
@@ -21,13 +21,12 @@ pub fn init(lua: &Lua) -> rlua::Result<()> {
 
 pub fn mousegrabber_handle(x: i32, y: i32, button: Option<(u32, ButtonState)>)
                          -> rlua::Result<()> {
-    let lua = LUA.lock()
-        .map_err(|err| rlua::Error::RuntimeError(
-            format!("Lua lock was poisoned {:#?}", err)))?;
-    let button_events = button.map(|(button, button_state)|
-                                   ::lua::mouse_events_to_lua(&lua.0, button, button_state))
-        .unwrap_or_else(|| Ok(vec![false, false, false, false, false]))?;
-    call_mousegrabber(&lua.0, (x, y, button_events))
+    run_with_lua(move |lua| {
+        let button_events = button.map(|(button, button_state)|
+                                       ::lua::mouse_events_to_lua(lua, button, button_state))
+            .unwrap_or_else(|| Ok(vec![false, false, false, false, false]))?;
+        call_mousegrabber(lua, (x, y, button_events))
+    })
 }
 
 fn call_mousegrabber(lua: &Lua,
--- a/src/lua/mod.rs
+++ b/src/lua/mod.rs
@@ -18,10 +18,11 @@ unsafe impl Send for LuaWrapper{}
 
 
 lazy_static! {
-    pub static ref LUA: Mutex<LuaWrapper> = Mutex::new(LuaWrapper(Lua::new()));
+    // XXX: The Mutex is not actually needed. The Lua state will be thread-local to the Lua thread.
+    static ref LUA: Mutex<LuaWrapper> = Mutex::new(LuaWrapper(Lua::new()));
 }
 
 pub use self::types::{LuaQuery, LuaResponse};
 pub use self::thread::{init, on_compositor_ready, running, send, update_registry_value,
-                       LuaSendError};
+                       run_with_lua, LuaSendError};
 pub use self::utils::{mods_to_lua, mods_to_rust, mouse_events_to_lua};
--- a/src/lua/mod.rs
+++ b/src/lua/mod.rs
@@ -1,8 +1,5 @@
 //! Lua functionality
 
-use rlua::Lua;
-use std::sync::Mutex;
-
 #[cfg(test)]
 mod tests;
 
@@ -12,15 +9,6 @@ mod rust_interop;
 mod init_path;
 mod utils;
 
-pub struct LuaWrapper(pub Lua);
-
-unsafe impl Send for LuaWrapper{}
-
-
-lazy_static! {
-    // XXX: The Mutex is not actually needed. The Lua state will be thread-local to the Lua thread.
-    static ref LUA: Mutex<LuaWrapper> = Mutex::new(LuaWrapper(Lua::new()));
-}
 
 pub use self::types::{LuaQuery, LuaResponse};
 pub use self::thread::{init, on_compositor_ready, running, send, update_registry_value,
--- a/src/lua/thread.rs
+++ b/src/lua/thread.rs
@@ -90,12 +90,15 @@ pub fn update_registry_value(category: String) {
 
 // Reexported in lua/mod.rs
 /// Run a closure with the Lua state. The closure will execute in the Lua thread.
-pub fn run_with_lua<F, G>(func: F) -> G
-    where F: FnOnce(&mut rlua::Lua) -> G
+pub fn run_with_lua<F>(func: F) -> rlua::Result<()>
+    where F: 'static + FnMut(&mut rlua::Lua) -> rlua::Result<()>
 {
-    let mut lua = LUA.lock().expect("LUA was poisoned!");
-    let lua = &mut lua.0;
-    func(lua)
+    // TODO: Uhm, error handling?
+    match send(LuaQuery::ExecWithLua(Box::new(func))).unwrap().recv().unwrap() {
+        LuaResponse::Error(err) => Err(err),
+        LuaResponse::Pong => Ok(()),
+        _ => Err(rlua::Error::CoroutineInactive)
+    }
 }
 
 // Reexported in lua/mod.rs:11
@@ -336,6 +339,12 @@ fn handle_message(request: LuaMessage, lua: &mut rlua::Lua) -> bool {
         LuaQuery::ExecRust(func) => {
             let result = func(lua);
             thread_send(request.reply, LuaResponse::Variable(Some(result)));
+        },
+        LuaQuery::ExecWithLua(mut func) => {
+            match func(lua) {
+                Ok(()) => thread_send(request.reply, LuaResponse::Pong),
+                Err(e) => thread_send(request.reply, LuaResponse::Error(e))
+            };
         },
         LuaQuery::HandleKey(press) => {
             trace!("Lua: handling keypress {}", &press);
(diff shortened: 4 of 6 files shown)


psychon:
I don't know if this is better, but it compiles (and it expresses more clearly that the idle function is only run once (`FnOnce` vs `FnMut`)):
```diff
diff --git a/src/lua/thread.rs b/src/lua/thread.rs
index 746bc31..ae8ba81 100644
--- a/src/lua/thread.rs
+++ b/src/lua/thread.rs
@@ -10,6 +10,7 @@ use std::fmt::{Debug, Formatter};
 use std::fmt::Result as FmtResult;
 use std::sync::{Mutex, RwLock};
 use std::sync::mpsc::{channel, Sender, Receiver};
+use std::cell::RefCell;
 use std::io::Read;
 
 use glib::MainLoop;
@@ -83,9 +84,6 @@ pub fn running() -> bool {
 pub enum LuaSendError {
     /// The thread was not initialized yet, crashed, was shut down, or rebooted.
     ThreadClosed,
-    /// The sender had an issue, most likey because the thread panicked.
-    /// Following the `Sender` API, the original value sent is returned.
-    Sender(LuaQuery)
 }
 
 /// Appends this combination of category and key to the registry queue.
@@ -107,6 +105,19 @@ pub fn run_with_lua<F>(func: F) -> rlua::Result<()>
     }
 }
 
+fn idle_add_once<F>(func: F)
+    where F: Send + 'static + FnOnce() -> ()
+{
+    let cell = RefCell::new(Some(func));
+    idle_add(move || {
+        match cell.borrow_mut().take() {
+            Some(f) => f(),
+            None => { panic!("The impossible happened"); }
+        };
+        Continue(false)
+    });
+}
+
 // Reexported in lua/mod.rs:11
 /// Attemps to send a LuaQuery to the Lua thread.
 pub fn send(query: LuaQuery) -> Result<Receiver<LuaResponse>, LuaSendError> {
@@ -116,26 +127,10 @@ pub fn send(query: LuaQuery) -> Result<Receiver<LuaResponse>, LuaSendError> {
     // Create a response channel
     let (response_tx, response_rx) = channel();
     let message = LuaMessage { reply: response_tx, query: query };
-    // No idea how to give the query to the thread without this
-    let (useless_tx, useless_rx) = channel();
-    useless_tx.send(message).map_err(|e| {
-        LuaSendError::Sender(e.0.query)
-    })?;
-    idle_add(move || {
-        let message = match useless_rx.recv() {
-            Err(e) => {
-                error!("Lua thread: unable to receive message: {}", e);
-                error!("Lua thread: now panicking!");
-                *RUNNING.write().expect(ERR_LOCK_RUNNING) = false;
-
-                panic!("Lua thread: lost contact with host, exiting!");
-            },
-            Ok(m) => m
-        };
+    idle_add_once(move || {
         let mut lua = LUA.lock().expect("LUA was poisoned!");
         trace!("Handling a request");
         handle_message(message, &mut lua.0);
-        Continue(false)
     });
     Ok(response_rx)
 }
diff --git a/src/modes/custom_lua.rs b/src/modes/custom_lua.rs
index 46894de..d68e645 100644
--- a/src/modes/custom_lua.rs
+++ b/src/modes/custom_lua.rs
@@ -286,8 +286,6 @@ fn send_to_lua<Q: Into<String>>(msg: Q) {
     let msg = msg.into();
     match lua::send(LuaQuery::Execute(msg.clone())) {
         Ok(_) => {},
-        Err(LuaSendError::Sender(err)) =>
-            warn!("Error while executing {:?}: {:?}", msg, err),
         Err(LuaSendError::ThreadClosed) => {
             warn!("Thread closed, could not execute {:?}", msg)
         }
```

Timidger:
Thanks for doing this! I wasn't looking forward to digging into the glib API so thanks for doing the hard work for me!

I opened another branch that is based on this one with some fixes. It still passes the manual test (e.g of trying to run the basic gears timer).

I'm going to hold off on merging this until I cut the 0.8 release though. There is an unrelated colours change that I want to push out and I don't want to merge this in and possibly make the build unstable.

psychon:
It just occurred to me that I broke quit/restart. I'm not at the right computer for this currently, but here is a quick attempt (that likely does not even compile yet):
```diff
diff --git a/thread.rs.orig b/thread.rs
index 28154d9..e19b6a4 100644
--- a/thread.rs.orig
+++ b/thread.rs
@@ -31,6 +31,8 @@ use ::layout::{lock_tree, ContainerType};
 
 thread_local! {
     static LUA: RefCell<rlua::Lua> = RefCell::new(rlua::Lua::new());
+
+    static MAIN_LOOP: RefCell<MainLoop> = RefCell::new(MainLoop::new(None, false));
 }
 
 lazy_static! {
@@ -135,7 +137,9 @@ pub fn send(query: LuaQuery) -> Result<Receiver<LuaResponse>, LuaSendError> {
     idle_add_once(move || {
         LUA.with(|lua| {
             trace!("Handling a request");
-            handle_message(message, &*lua.borrow());
+            if (!handle_message(message, &*lua.borrow())) {
+                MAIN_LOOP.with(|loop| loop.borrow().quit());
+            }
         });
     });
     Ok(response_rx)
@@ -248,7 +252,7 @@ fn main_loop() {
     LUA.with(|lua| rust_interop::register_libraries(&*lua.borrow()))
         .expect("Could not register lua libraries");
     lua_init();
-    MainLoop::new(None, false).run();
+    MAIN_LOOP.with(|loop| loop.borrow().run());
     RUNNING.store(false, Ordering::Relaxed);
 }
 
@@ -282,6 +286,9 @@ fn handle_message(request: LuaMessage, lua: &rlua::Lua) -> bool {
             let _new_handle = thread::Builder::new()
                 .name("Lua re-init".to_string())
                 .spawn(move || {
+                    // TODO: Check that this works
+                    // TODO: If it works: Uhm, isn't this racy? This thread will set RUNNING to
+                    // false while the new thread will set RUNNING to true.
                     init().expect("Could not init");
                     keys::init();
                 });
```

Timidger:
> The panic in 991c378 should likely have a better error message (sorry for providing a bad one).

I fixed this by just `unwrap`ing. It's always going to be a `Some` so why bother? ;). I also turned the `RefCell` into a `Cell`, since the closure passed in doesn't need to be thread safe.

> 7183f95 removes the only error case from init(). Should its return value be removed? I kept the Lua thread initialisation in the main thread so that I didn't have to introduce a panic here, that's why I noticed this.

Nice catch. Also no longer made it public.

Timidger:
As for your new patch to fix restart/quit (which I haven't tested yet, so maybe it doesn't compile) but it won't be racy to do what you're doing because each thread has its own (lazy-initialized) `MainLoop`. So the new `MainLoop` would have its value updated correctly (and start running) while the old one will stop because we set it to `false`.

Timidger:
Your patch was fine (except `loop` is a reserved keyword), has been applied to my branch.

Timidger:
hmm, even with your patch restart is broken. I'll look into this.

Timidger:
Yeah this is super weird. I think the crate for glib isn't wrapped very well (e.g they implement `Send` and `Sync` for `MainLoop` even though I don't think it should be...). I was able to get a segfault just from spamming the restart command.

In the log I keep seeing this show up:

```
** (process:11796): CRITICAL **: g_type_info_get_tag: assertion 'GI_IS_TYPE_INFO (info)' failed
```

so I think that the `MainLoop` isn't being cleaned up properly. `MainLoop` doesn't have a `drop` impl, and we are for sure calling `quit` so I'm not sure if the issue is from something else...

How does Awesome implement restarting? If I remember correctly it was in place, so does it even tear down the Lua state at all?

At a certain point, the work on Awesome will start conflicting to the current version of Way Cooler so maybe it's better if this was merged into an "awesome" branch where all the work on awesome can go instead. In that branch we can remove the ability to restart in place / design it better to act like Awesome.

psychon:
I'll look at the difference between `Cell` and `RefCell`, but thanks for cleaning up. :-)

Also, when I have the time, I'll take a look at the segfault. The error message sounds like it could be coming from lgi (`g_type_info_get_tag` sounds like introspection to me). I know that lgi is not safe to unload, but there are work-arounds in place that should make sure that it doesn't actually get unloaded, even when the Lua state is thorn down? Anyway, I like that way-cooler can restart in place / without losing all state and would like to keep that ability...

On `glib.MainLoop` being Send and Sync: Well, for example `.quit()` actually works across threads while `.run()` of course may only be called from a single thread. I guess part of the problem is that GLib wasn't written with rust in mind. ;-) (and glib-rs is afaik (partly) auto-generated)

Timidger:
> I'll look at the difference between Cell and RefCell, but thanks for cleaning up. :-)

Basically for `Cell` the way you use it is either to get a mutable reference to the inner thing, or you copy it out and replace it with something else. Since for the `FnOnce` we only move it out, it's more obvious we don't actually need dynamic borrowing. Both are forms of interior mutability, but `Cell` is useful for types that are copyable (e.g `i32`) and `RefCell` are useful for types that aren't.

> Anyway, I like that way-cooler can restart in place / without losing all state and would like to keep that ability...

It is a nice feature, so hopefully we can work around it. You know LGI/Glib more than I, so hopefully you can figure it out. I'm going to take a look this weekend and see if I can stop it from segfaulting.

> I guess part of the problem is that GLib wasn't written with rust in mind. ;-) (and glib-rs is afaik (partly) auto-generated)

I've had a lot of safety issues with the gtk-rs projects in general. I'm happy that they put the effort in to provide the wrappers/bindings, but there's a lot of issues with the actual API surface in terms of both safety and ergonomics. E.g for Cairo they could have built up a finite state automata using the Rust type system to ensure you could never misuse it (e.g do any draw operation before running the cairo start drawing function) but they didn't. 

But because `MainLoop` has `Sync` and `Send` we shouldn't need to put it behind a `RefCell` especially since all of its functions take `&self`.

psychon:
What's the status of this? Since you merged this to a branch, can this PR be closed and new PRs be submitted against that branch? Or do you have other plans?
As I remember, the only issue is that restart crash, right?

Timidger:
I haven't had a chance to look at this in a while. I do have a separate branch ([here](https://github.com/way-cooler/way-cooler/tree/glib_main_loop_fixes)) so this PR can be closed and the rest of the work can continue on that.

I would really like to solve the restart crash, since a segfault is pretty bad. If it can't be solved I'm fine merging it into a different, longer living branch that will serve as the Awesome rewrite. But to be honest if we are doing that I'd rather just restart from scratch designing how the Lua thread/handles are set up. There's issues with the current code beyond just the `lazy_static`/`thread_local` business (to list them briefly: code to get init file is convoluted, init of the Lua thread jumps around a lot and requires way too much knowledge of different parts of the code base, etc.).

I know you said you liked the restart functionality (and I do too, it's very helpful!) but that's not something that's in base Awesome and for the Awesome rewrite I don't see it as necessary to have in especially if it has this large safety hole. If I do start a separate branch I'll just remove this functionality and strip out everything I can so that there's a mostly fresh start on switching to be Awesome / use wlroots-rs (which is certainly not ready, but at the very least I can start the work).

psychon:
Okay, closing this PR then since its in "the" branch.

I haven't had time to look at the restart crash yet, but will likely be able to do so tomorrow. Is just appending `awesome.restart()` to the Lua code enough to trigger this? (Oh... this would call `g_mainloop_quit` before we reached `g_mainloop_run`, so the quit does not really work... perhaps this is already an explanation for the crash?).
Or... how else did you trigger the restart from Lua? A timer?

Timidger:
The restart is triggered with a special message sent to the Lua thread (E.g `LuaQuery::Restart`). The way it restarts is it kills the Lua thread (by returning `false` in `handle_message`) and then spawns another one (this is to clear all the Lua state easily). When it does this some weirdness I haven't pined down yet happens with the gmain loop stuff.

Because both the `LUA` and `MAIN_LOOP` (glib main loop variable) live in `thread_local`s a new glib `MainLoop` is made when the new thread spawns. I think this is where the problem is, as I have't looked at the code but I have a feeling starting two them clobbers some internal global state and is causing the crash.

One solution would be for the in-place restart to _not_ make a new `rlua::Lua` and instead do something else...which _should_ solve the problem. Though I'm not sure how you can be sure the state is clean.

Alternatively, if we just change the restart to behave like Awesome's (e.g `exec way-cooler` essentially) then it will almost certainly clean up properly but it will no longer be in place.

psychon:
Uhm, I meant, how do I make way-cooler send this Restart message to the thread? Is there some already existing mechanism for that or should I just hack something up?
Your theory about the thread start sounds convincing. Since there is a race with the `RUNNING` global anyway, it might make sense to be able to replace the complete Lua state without spawning a new thread. I guess this requires a `Cell` around the Lua state (or something similar...).

Timidger:
> Uhm, I meant, how do I make way-cooler send this Restart message to the thread? Is there some already existing mechanism for that or should I just hack something up?

Ah sorry. [Here is where we use it for the keybindings](https://github.com/way-cooler/way-cooler/blob/master/src/commands/defaults.rs#L180). Basically just send the query for restart via the channel.

> ... it might make sense to be able to replace the complete Lua state without spawning a new thread. I guess this requires a Cell around the Lua state (or something similar...).

Hm, I'll give that a try.

PR #475, opened 2018-02-04
More awesome stuff:refresh signal, override type() and have screens

psychon:


src/lua/rust_interop.rs
```diff
@@ -56,6 +57,37 @@ pub fn register_libraries(lua: &rlua::Lua) -> LuaResult<()> {
...
+        Value::Integer(_) => Ok("number".to_owned()),
+        Value::Number(_) => Ok("number".to_owned()),
+        Value::String(_) => Ok("string".to_owned()),
+        Value::Function(_) => Ok("function".to_owned()),
+        Value::Thread(_) => Ok("thread".to_owned()),
+        Value::Error(e) => Err(e),
+        Value::Table(t) => {
+            // Handle our own objects specially: Return the metatable's .__class.name if it exists,
+            // else return "table".
+            let unused_error = Err(rlua::Error::CoroutineInactive);
```
Timidger: Using a `match` here is a much cleaner way to handle this case:

```rust
match t.get_metatable() {
    None => Ok("table".to_owned()),
    Some(t) => t.raw_get::<_, rlua::Table>("__class")?.raw_get("name")
}
```

src/lua/thread.rs
```diff
@@ -111,6 +111,12 @@ pub fn run_with_lua<F>(func: F) -> rlua::Result<()>
     }
 }
 
+fn emit_refresh(lua: &rlua::Lua) {
+    // FIXME: This assumes that the Lua code does not e.g. unset the global. Can we somehow emit
+    // the signal directly?
+    let _ = lua.exec::<()>("awesome.emit_signal('refresh')", None);
```
Timidger: Can you achieve this by calling `signal::global_emit_signal("refresh".to_owned(), ())`? That's what `awesome.emit_signal` is bound to, and since the signals emitted are part of the global signals list this should work.

src/lua/rust_interop.rs
psychon: I don't like the `?` in there. When a table does not have `__class` member, then `type()` should return `"table"` instead of causing a Lua error. My "mess" was written with the goal to have `"table"` only once, but if you want, I can amend in the following change:
```diff
diff --git a/src/lua/rust_interop.rs b/src/lua/rust_interop.rs
index 7b3fb53..af0f1bb 100644
--- a/src/lua/rust_interop.rs
+++ b/src/lua/rust_interop.rs
@@ -79,11 +79,12 @@ fn type_override(_lua: &rlua::Lua, arg: rlua::Value) -> Result<String, rlua::Err
         Value::Table(t) => {
             // Handle our own objects specially: Return the metatable's .__class.name if it exists,
             // else return "table".
-            let unused_error = Err(rlua::Error::CoroutineInactive);
-            t.get_metatable().map_or(unused_error,
-                                     |t| t.raw_get("__class"))
-                .and_then(|t: rlua::Table| t.raw_get("name"))
-                .or(Ok("table".to_owned()))
+            match t.get_metatable() {
+                None => Ok("table".to_owned()),
+                Some(t) => t.raw_get("__class")
+                    .and_then(|t: rlua::Table| t.raw_get("name"))
+                    .or(Ok("table".to_owned()))
+            }
         }
     }
 }
```

src/lua/thread.rs
psychon: Thanks, that looks like exactly the right thing.

psychon: Sigh...
```
error[E0603]: module `signal` is private
  --> src/lua/thread.rs:27:5
   |
27 | use super::super::awesome::signal;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Should I make it public?

src/lua/rust_interop.rs
Timidger: Ah right good catch. 

Yeah that unused error is something that stood out to me and is the thing I wanted to avoid. Your patch looks good.

Don't forget to remove the line above that suppresses `unused`

Timidger: Also for the future: for unused variables you can use `_variable_name` to make the warning be suppressed.

src/lua/thread.rs
Timidger: Yeah go ahead

Timidger: also might be easier at that point to do `use awesome::signal` and not have to go up two layers instead

src/lua/rust_interop.rs
psychon: Well, it is not unused in that sense. If there is no metatable, then a `Result` with that error is created via `map_or` (so it *is* used), but then the `.or()` later replaces that error with `Ok("table".to_owned())`. So in this sense the error is unused, but it is not unused in the "prefix with `_`"-sense. Basically, I wanted to write `"table"` only once, so turned "no metatable" into some kind of error.
I'll amend the commit with the patch above later.

Timidger: Got it. Since I missed the case that you don't want to return early if the meta table has the wrong thing, the original is fine too. It can be kinda awkward doing these `Option` and `Result` conversions, so not a big deal

psychon:
Whoops, I pushed `:branch` instead of `+branch`. The branch is now back, but apparently that doesn't mean that GitHub let's me reopen this pull request. Should I open a new one or do you just want to pull from the branch directly?
Should be `git pull 	https://github.com/psychon/way-cooler.git  more-awesome-stuff`, if you want.

Timidger:
I just pulled it, so nothing more for you to do. No worries.

PR #478, opened 2018-02-09
Screen: Initialise geometry and workarea

psychon:
Up to now, all screens had size 0x0.

### DIFF
--- a/src/awesome/screen.rs
+++ b/src/awesome/screen.rs
@@ -80,9 +80,12 @@ impl <'lua> Screen<'lua> {
         Ok(Screen::allocate(lua, class)?.build())
     }
 
-    fn init_screens(&mut self, outputs: Vec<Output>) -> rlua::Result<()> {
+    fn init_screens(&mut self, output: WlcOutput, outputs: Vec<Output>) -> rlua::Result<()> {
         let mut state = self.state()?;
+        let resolution = output.get_resolution().unwrap();
         state.outputs = outputs;
+        state.geometry.size = resolution;
+        state.workarea.size = resolution;
         self.set_state(state)
     }
 
@@ -120,7 +123,7 @@ pub fn init(lua: &Lua) -> rlua::Result<Class> {
     let mut screens: Vec<Screen> = vec![];
     for output in WlcOutput::list() {
         let mut screen = Screen::cast(Screen::new(lua)?)?;
-        screen.init_screens(vec![output.into()])?;
+        screen.init_screens(output, vec![output.into()])?;
         // TODO Move to Screen impl like the others
         screens.push(screen);
     }


src/awesome/screen.rs
```diff
@@ -80,9 +80,12 @@ impl <'lua> Screen<'lua> {
         Ok(Screen::allocate(lua, class)?.build())
     }
 
-    fn init_screens(&mut self, outputs: Vec<Output>) -> rlua::Result<()> {
+    fn init_screens(&mut self, &output: &WlcOutput, outputs: Vec<Output>) -> rlua::Result<()> {
         let mut state = self.state()?;
+        let resolution = output.get_resolution().unwrap();
         state.outputs = outputs;
```
Timidger: Hmm, ok I'm a bit confused about this so maybe you can explain...

I assume "screens" in the original Awesome code referred to what we call "Outputs" here. So why does the screen state store a list of outputs (which was copied from the Awesome version). Are the screens actually tagged workspaces that the user sees or is it something else? Because now I'm confused why we a screen state can refer to multiple outputs.

If we should be storing a list, are we supposed to calculate resolution, workarea, etc. differently?

Or did I mess up the data structure and instead there should be a global list of screens or something?

src/awesome/screen.rs
```diff
@@ -80,9 +80,12 @@ impl <'lua> Screen<'lua> {
         Ok(Screen::allocate(lua, class)?.build())
     }
 
-    fn init_screens(&mut self, outputs: Vec<Output>) -> rlua::Result<()> {
+    fn init_screens(&mut self, &output: &WlcOutput, outputs: Vec<Output>) -> rlua::Result<()> {
```
Timidger: `WlcOutput` is a `Copy` type (since it's just internally a `uintptr_t`), so passing it by reference is unnecessary here (the borrow checker will be more ok with that, it's more ergonomic to use, and it will probably generate better code because it will be a layer less of indirection).

src/awesome/screen.rs
psychon: Imagine you have two monitors displaying the same thing. In XrandR-terms this means you have one CRTC that drives two OUTPUTs.
Awesome turns this one CRTC into one screen, and the screen has two outputs. I do not have two monitors displaying the same thing, but in awesome when you do ' screen["HDMI-1"]`, then `HDMI-1` is actually the name of an OUTPUT and awesome looks up the corresponding CRTC. Also, you can query the size in millimeter of outputs (here it shows only one table entry since I do not have two monitors displaying the same thing):
```
 $ awesome-client 'return require("gears.debug").dump_return(screen[1].outputs)'
   string "table: 0x55d5f4e3af40
  HDMI-1 : table: 0x55d5f4e3af80
    mm_width : 477 (number)
    mm_height : 268 (number)"
```
In awesome's data structures, `struct a_screen` has an `uint32_t xid;` (which is used differently depending on if RandR or just XINERAMA is available...) that identifies the CRTC and a `screen_output_array_t` containing a list of `struct screen_output_t`. Each of these has a `name`, `mm_width`, `mm_height` and an array of `xcb_randr_output_t` (the latest revision of RandR made the mapping again more complicated than what I explained above, hence this now needs to be an array :-( ).

TL;DR: A screen has a size in pixels and only one geometry/workarea. But since it can be displayed on different "physical monitors" at the same time, it needs a list of sizes in millimeters.

src/awesome/screen.rs
psychon: Okay, will do a force push adding the following diff:
```diff
diff --git a/src/awesome/screen.rs b/src/awesome/screen.rs
index 809495e..d7b5751 100644
--- a/src/awesome/screen.rs
+++ b/src/awesome/screen.rs
@@ -80,7 +80,7 @@ impl <'lua> Screen<'lua> {
         Ok(Screen::allocate(lua, class)?.build())
     }
 
-    fn init_screens(&mut self, &output: &WlcOutput, outputs: Vec<Output>) -> rlua::Result<()> {
+    fn init_screens(&mut self, output: WlcOutput, outputs: Vec<Output>) -> rlua::Result<()> {
         let mut state = self.state()?;
         let resolution = output.get_resolution().unwrap();
         state.outputs = outputs;
@@ -123,7 +123,7 @@ pub fn init(lua: &Lua) -> rlua::Result<Class> {
     let mut screens: Vec<Screen> = vec![];
     for output in WlcOutput::list() {
         let mut screen = Screen::cast(Screen::new(lua)?)?;
-        screen.init_screens(&output, vec![output.into()])?;
+        screen.init_screens(output, vec![output.into()])?;
         // TODO Move to Screen impl like the others
         screens.push(screen);
     }
```

src/awesome/screen.rs
Timidger: Ah ok I see, thanks for the explanation. Things will have to change in Way Cooler significantly with this though since:

1. Outputs are no longer just one long frame buffer like in X. They are separate frame buffers that Way Cooler has to manage. 
2. Xrandr obviously doesn't work.

Thanks for detailing it out, the work that needs to be discussed later I just wanted to bring it up while you were working in the file.

PR #480, opened 2018-02-10
lua::tests: Drop wait_for_thread()

psychon:
There is nothing wrong with sending a message to the thread before it is
ready to accept them. This just works fine.

There would be a deadlock if the thread never starts, but let's just
ignore that...

--------

By the way, it is a bit weird that these tests depend on the first tests starting the thread so that the other tests can use that....

### DIFF
--- a/src/lua/tests.rs
+++ b/src/lua/tests.rs
@@ -1,21 +1,9 @@
 //! Lua thread test.
 //! Spawns a lua thread!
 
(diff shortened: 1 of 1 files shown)


Timidger:
Oh yeah these tests were written very poorly a long time ago. They really need to not rely on global state.

Thanks for making it a bit simpler though, it'll make it easier to fix it later.

PR #481, opened 2018-02-10
emit_signal: Log errors

psychon:
Instead of returning errors that happen in the actual Lua callback
function to the caller, these are now just logged. This is the same
behaviour as with awesome.


Timidger:
If you could rebase on latest master when you get a chance, I can merge this in.

PR #482, opened 2018-02-10
Handle constructor object arguments

psychon:
Doing e.g. drawin{ x = 42 } is supposed to create a drawin with its x
property set to 42. This commit actually implements this for all
supported objects except for buttons.

For buttons this causes an error during loading of the awesome default
config and so the argument to buttons is still ignored for now.

  cause: RuntimeError("Lock is an invalid modifier")


Timidger:
To fix button: update `mod_to_rust` to accept `"Lock"` as a valid modifier.

Timidger:
Should be fixed with #434, please rebase

psychon:
Done, and I also got rid of the `FIXME`.

psychon:
Oh, I broke tests... Will investigate tomorrow.

psychon:
I couldn't sleep, so I fixed the tests instead. @Timidger if Travis doesn't complain this should be good to merge.

PR #486, opened 2018-02-10
screen: Fix screen index method

psychon:
Before this, screen[1] caused a panic:

  thread 'Lua thread' panicked at 'index out of bounds: the len is 1 but
  the index is 1', src/liballoc/vec.rs:1552:10


Timidger:
Freaking 1 based indexing ;)

Nice catch

psychon:
Oh and: Yay for rust making this so clear. In C, reading beyond the array would result in lots of fun.

PR #489, opened 2018-02-10
xrdb_get_value: Don't pretend DPI is one

psychon:
...and I was wondering why drawing fonts didn't work and the default
config constructed a tiny wibar.

The code in awesome can (and has to) handle a missing "Xft.dpi" value.
It does so by falling by to compute a DPI value via root.size() and
root.size_mm(). Since that doesn't produce useful values either, it just
falls back to a hardcoded 96. That produces way saner results than
trying to display text with 1 dot per inch. That's just too low-res.

This effectively reverts commit 4480fbc794ddcb4e5d1a659fca7b.

### DIFF
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -277,11 +277,8 @@ fn xkb_get_layout_group<'lua>(lua: &'lua Lua, _: ()) -> rlua::Result<Value<'lua>
     }
 }
 
-fn xrdb_get_value(lua: &Lua, (_resource_class, resource_name): (String, String))
+fn xrdb_get_value(_lua: &Lua, (_resource_class, _resource_name): (String, String))
                   -> rlua::Result<Value> {
-    if &resource_name == "Xft.dpi" {
-        return Ok("1".to_lua(lua)?)
-    }
     warn!("xrdb_get_value not supported");
     Ok(Value::Nil)
 }


psychon:
Done & nope, I am not seeing anything. The "tiny wibar" refers to the wibar ending up with a height of something like 2 pixels. This is because `beautiful.get_font_height` returns 1 pixel (that's already an inch with 1 DPI, way larger than the requested font size of 8 point!) and `awful.wibar` multiplies this with 1.5 and rounds.

So far I am just trying to get the code to render a PNG file containing the expected content (and at some point rendering the PNG failed because the surface had size 0x0 and that was the starting point for the recent series of PRs).

Timidger:
Ah ok cool.

I'm trying that, and I still get a 0 bytes png out of it...weird. Might need more work then. But yes please sleep, you've done more than enough today 😄

psychon:
> I'm trying that, and I still get a 0 bytes png out of it...weird.

That was a case of PEBKAC. The value that I was giving to `gears.surface` was `nil`. The "path" needs an additional `.drawable`. And I noticed only when my `println!` in `get_surface` didn't trigger...

So, the following code (assuming awesome's default config is also loaded)
```lua
require("gears.timer").start_new(1, function()
	require("gears.surface")(screen[1].mywibox._drawable.drawable.surface, false):write_to_png("/tmp/t.png")
end)
```
just produced the following `t.png` for me:
t

PR #490, opened 2018-02-11
Add awesome.systray

psychon:
The implementation of awesome.systray hardcodes that no systray icons
are registered and that the systray is currently not visible.


psychon:
Rebased on latest master (aka "the rlua update").

PR #492, opened 2018-02-11
Update to an rlua version that provides wrappers for `lua_setuservalue`

psychon:
BE CAREFUL WHEN MERGING THIS!

I bet that this together with some of my other pull requests does not build. Thus, I would suggest that you merge this PR first, since rebasing the other PRs should be simpler than rebasing this one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,7 +412,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "rlua"
-version = "0.9.7"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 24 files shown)


psychon:
Uhm... once again, I broke tests :-(

psychon:
Weird. I randomly get tests failing with `ThreadClosed`, but I do not know why the thread, well, closed.

Timidger:
> Weird. I randomly get tests failing with ThreadClosed, but I do not know why the thread, well, closed.

This is because you got rid of code that waited for the thread to start in the tests in 2345337. 

I suggest you revert that before merging this in, as eventually we'll have tests failing for no reason and that will be very annoying/confusing...

PR #479, opened 2018-02-10
Make sending to the Lua thread fail after it has panicked

psychon:

PR #493, opened 2018-02-11
Remove lua::running()

psychon:
Its only (remaining) caller is a function that is only called from Lua
anyway, so it does not really make sense to check if the Lua thread is
running.


psychon:
Updated so that this actually compiles and so that the `RUNNING` static is completely removed.

I rebased this onto latest master and verified that each individual commit passed `cargo test`.

psychon:
This interacts badly with #479, which adds a new kind of error to `send()`. That would basically mean reverting the commit which removed all the error handling. Back to the drawing board!

PR #499, opened 2018-02-24
Fix the type() override for the new object layout

psychon:
I guess this function does not really belong into lua::rust_interop, but
where else should it be put?

### DIFF
--- a/src/lua/rust_interop.rs
+++ b/src/lua/rust_interop.rs
@@ -69,22 +69,23 @@ fn type_override(_lua: &rlua::Lua, arg: rlua::Value) -> Result<String, rlua::Err
         Value::Nil => Ok("nil".to_owned()),
         Value::Boolean(_) => Ok("boolean".to_owned()),
         Value::LightUserData(_) => Ok("userdata".to_owned()),
-        Value::UserData(_) => Ok("userdata".to_owned()),
         Value::Integer(_) => Ok("number".to_owned()),
         Value::Number(_) => Ok("number".to_owned()),
         Value::String(_) => Ok("string".to_owned()),
         Value::Function(_) => Ok("function".to_owned()),
         Value::Thread(_) => Ok("thread".to_owned()),
+        Value::Table(_) => Ok("table".to_owned()),
         Value::Error(e) => Err(e),
-        Value::Table(t) => {
-            // Handle our own objects specially: Return the metatable's .__class.name if it exists,
-            // else return "table".
-            match t.get_metatable() {
-                None => Ok("table".to_owned()),
-                Some(t) => t.raw_get("__class")
-                    .and_then(|t: rlua::Table| t.raw_get("name"))
-                    .or(Ok("table".to_owned()))
-            }
+        Value::UserData(o) => {
+            // Handle our own objects specially: Get the object's class from its user value's
+            // metatable's __class entry. Then get the class name from the class's user value's
+            // metatable's name entry.
+            o.get_user_value::<rlua::Table>().ok()
+                .and_then(|table| table.get_metatable())
+                .and_then(|meta| meta.raw_get::<_, rlua::AnyUserData>("__class").ok())
+                .and_then(|class| class.get_user_value::<rlua::Table>().ok())
+                .map(|table| table.raw_get("name"))
+                .unwrap_or_else(|| Ok("userdata".to_owned()))
         }
     }
 }


src/lua/rust_interop.rs
```diff
@@ -69,22 +69,26 @@ fn type_override(_lua: &rlua::Lua, arg: rlua::Value) -> Result<String, rlua::Err
...
+            // metatable's name entry.
+            // This is in an extra function so that we can easily just return "userdata" if any of
+            // these steps fail.
+            fn identify_user_data(data: &rlua::AnyUserData) -> Result<String, rlua::Error> {
+                match data.get_user_value::<rlua::Table>()?.get_metatable() {
+                    None => Err(rlua::Error::UserDataTypeMismatch),
+                    Some(m) => m.raw_get::<_, rlua::AnyUserData>("__class")?.get_user_value::<rlua::Table>()?.raw_get("name")
+                }
             }
+            identify_user_data(&o).or(Ok("userdata".to_owned()))
```
Timidger: You shouldn't need to wrap this in a function in order to succinctly say "get the stored name, otherwise just "userdata" is fine". You should be able to write something like this (this is probably not correct, you might need nested `map` or to use `and_then` / `ok` in places. Basically just throw away the error we don't care about it at all and convert them all to `Option`s using `ok`.

```rust
o.get_user_value::<rlua::Table>().ok()
    .map(rlua::Table::get_metatable)
    .map(|meta| meta.raw_get::<_, rlua::AnyUserData>("__class"))
    .map(|class| class.get_user_value::<rlua::Table>())
    .map(|table| table.raw_get("name"))
    .unwrap_or_else(|| "userdata".to_owned())
```

Timidger:
Putting it in rust_interop is ok

src/lua/rust_interop.rs
psychon: `map` wants a function that returns the value directly, but `get_metatable` returns `Option<the value>`. I'll see how I can make this work...

psychon: The following works. Do you want me to amend this into the commit in this PR?
```diff
diff --git a/src/lua/rust_interop.rs b/src/lua/rust_interop.rs
index 612bfb1..d3ff4fc 100644
--- a/src/lua/rust_interop.rs
+++ b/src/lua/rust_interop.rs
@@ -80,15 +80,15 @@ fn type_override(_lua: &rlua::Lua, arg: rlua::Value) -> Result<String, rlua::Err
             // Handle our own objects specially: Get the object's class from its user value's
             // metatable's __class entry. Then get the class name from the class's user value's
             // metatable's name entry.
-            // This is in an extra function so that we can easily just return "userdata" if any of
-            // these steps fail.
-            fn identify_user_data(data: &rlua::AnyUserData) -> Result<String, rlua::Error> {
-                match data.get_user_value::<rlua::Table>()?.get_metatable() {
-                    None => Err(rlua::Error::UserDataTypeMismatch),
-                    Some(m) => m.raw_get::<_, rlua::AnyUserData>("__class")?.get_user_value::<rlua::Table>()?.raw_get("name")
-                }
-            }
-            identify_user_data(&o).or(Ok("userdata".to_owned()))
+            o.get_user_value::<rlua::Table>().ok()
+                .map(|table| table.get_metatable())
+                .unwrap_or(None)
+                .map(|meta| meta.raw_get::<_, rlua::AnyUserData>("__class").ok())
+                .unwrap_or(None)
+                .map(|class| class.get_user_value::<rlua::Table>().ok())
+                .unwrap_or(None)
+                .map(|table| table.raw_get("name"))
+                .unwrap_or_else(|| Ok("userdata".to_owned()))
         }
     }
 }
```
(Exercise for the reader: Explain why each `.ok()` and `.unwrap_or(None)` is necessary. Especially the later is lots of fun, I think.)

psychon: Uhhh, I found a simpler solution: With `.and_then` I get the `Option<X>` directly while `.map` produced an `Option<Option<X>>` that needed to be flattened via `.unwrap_or(None)`:
```diff
diff --git a/src/lua/rust_interop.rs b/src/lua/rust_interop.rs
index 612bfb1..3d5adb9 100644
--- a/src/lua/rust_interop.rs
+++ b/src/lua/rust_interop.rs
@@ -80,15 +80,12 @@ fn type_override(_lua: &rlua::Lua, arg: rlua::Value) -> Result<String, rlua::Err
             // Handle our own objects specially: Get the object's class from its user value's
             // metatable's __class entry. Then get the class name from the class's user value's
             // metatable's name entry.
-            // This is in an extra function so that we can easily just return "userdata" if any of
-            // these steps fail.
-            fn identify_user_data(data: &rlua::AnyUserData) -> Result<String, rlua::Error> {
-                match data.get_user_value::<rlua::Table>()?.get_metatable() {
-                    None => Err(rlua::Error::UserDataTypeMismatch),
-                    Some(m) => m.raw_get::<_, rlua::AnyUserData>("__class")?.get_user_value::<rlua::Table>()?.raw_get("name")
-                }
-            }
-            identify_user_data(&o).or(Ok("userdata".to_owned()))
+            o.get_user_value::<rlua::Table>().ok()
+                .and_then(|table| table.get_metatable())
+                .and_then(|meta| meta.raw_get::<_, rlua::AnyUserData>("__class").ok())
+                .and_then(|class| class.get_user_value::<rlua::Table>().ok())
+                .map(|table| table.raw_get("name"))
+                .unwrap_or_else(|| Ok("userdata".to_owned()))
         }
     }
 }
```

psychon:
I amended the commit with the change I proposed above.

Note that your original `.map(rlua::Table::get_metatable)` does not work:
```
error[E0631]: type mismatch in function arguments
  --> src/lua/rust_interop.rs:84:18
   |
84 |                 .and_then(rlua::Table::get_metatable)
   |                  ^^^^^^^^
   |                  |
   |                  expected signature of `fn(rlua::Table<'_>) -> _`
   |                  found signature of `for<'r> fn(&'r rlua::Table<'_>) -> _`
```
(no idea what this error message wants to tell me)

Timidger:
Yes good thank that's exactly what I meant, I just can never remember when to use `map` or `and_then` (well, I do remember I just don't remember everything that's returned)

PR #500, opened 2018-02-24
Get rid of set state

psychon:
Some reasons are given in the first commit message


Timidger:
Yes you were correct, `set_state` is very silly.

PR #510, opened 2018-03-24
Keep a list of tags and implement root.tags()

psychon:
Active tags are now saved in a table in the registry. Adding is easy,
but removing is a bit complicated...

root.tags() is implemented by returning a copy of exactly this table.

### DIFF
--- a/src/awesome/root.rs
+++ b/src/awesome/root.rs
@@ -84,6 +84,57 @@ fn set_wallpaper<'lua>(_: &'lua Lua, _pattern: *mut cairo_pattern_t) -> rlua::Re
 
 fn tags<'lua>(lua: &'lua Lua, _: ()) -> rlua::Result<Table<'lua>> {
     let table = lua.create_table()?;
-    // TODO FIXME Get tags
+    let activated_tags = lua.named_registry_value::<Table>(super::tag::TAG_LIST)?;
+    for pair in activated_tags.pairs::<Value, Value>() {
+        let (key, value) = pair?;
+        table.set(key, value)?;
+    }
     Ok(table)
+}
+
+#[cfg(test)]
+mod test {
+    use rlua::Lua;
+    use super::super::root;
+    use super::super::tag;
+
+    #[test]
+    fn tags_none() {
+        let lua = Lua::new();
+        tag::init(&lua).unwrap();
+        root::init(&lua).unwrap();
+        lua.eval(r#"
+local t = root.tags()
+assert(type(t) == "table")
+assert(type(next(t)) == "nil")
+"#, None).unwrap()
+    }
+
+    #[test]
+    fn tags_does_not_copy() {
+        let lua = Lua::new();
+        tag::init(&lua).unwrap();
+        root::init(&lua).unwrap();
+        lua.eval(r#"
+local t = tag{ activated = true }
+local t2 = root.tags()[1]
+assert(t == t2)
+t2.name = "Foo"
+assert(t.name == "Foo")
+"#, None).unwrap()
+    }
+
+    #[test]
+    fn tags_some() {
+        let lua = Lua::new();
+        tag::init(&lua).unwrap();
+        root::init(&lua).unwrap();
+        lua.eval(r#"
+local first = tag{ activated = true }
+local second = tag{ activated = true }
+local t = root.tags()
+assert(t[1] == first)
+assert(t[2] == second)
+"#, None).unwrap()
+    }
 }
--- a/src/awesome/root.rs
+++ b/src/awesome/root.rs
@@ -137,4 +137,19 @@ assert(t[1] == first)
 assert(t[2] == second)
 "#, None).unwrap()
     }
+
+    #[test]
+    fn tags_removal() {
+        let lua = Lua::new();
+        tag::init(&lua).unwrap();
+        root::init(&lua).unwrap();
+        lua.eval(r#"
+local first = tag{ activated = true }
+local second = tag{ activated = true }
+first.activated = false
+local t = root.tags()
+assert(t[1] == second)
+assert(type(t[2]) == "nil")
+"#, None).unwrap()
+    }
 }
--- a/src/awesome/tag.rs
+++ b/src/awesome/tag.rs
@@ -154,10 +154,11 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
                 if tag_ref == &*value.borrow::<TagState>()? as *const _ {
                     found = true;
                     // Now remove this by shifting everything down...
-                    for index in key .. activated_tags_count-1 {
+                    for index in key .. activated_tags_count {
                         activated_tags.set(index,
                                            activated_tags.get::<_, Value>(index + 1)?)?;
                     }
+                    activated_tags.set(activated_tags_count, Value::Nil)?;
                     break;
                 }
             }
--- a/src/awesome/tag.rs
+++ b/src/awesome/tag.rs
@@ -79,7 +79,11 @@ fn method_setup<'lua>(lua: &'lua Lua, builder: ClassBuilder<'lua>) -> rlua::Resu
            .property(Property::new("activated".into(),
                                    Some(lua.create_function(set_activated)?),
                                    Some(lua.create_function(get_activated)?),
-                                   Some(lua.create_function(set_activated)?)))
+                                   Some(lua.create_function(set_activated)?)))?
+           .property(Property::new("clients".into(),
+                                   None,
+                                   Some(lua.create_function(get_clients)?),
+                                   None))
 }
 
 impl_objectable!(Tag, TagState);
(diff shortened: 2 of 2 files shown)


psychon:
Heh, this seems to "break" `cargo run` by making the taglist notice some missing API:
```
2018-03-24 17:49:27 E: awesome: Error during a protected call: /usr/share/awesome/lib/awful/widget/taglist.lua:342: attempt to call a nil value (method 'clients')
stack traceback:
	[C]: in method 'clients'
	/usr/share/awesome/lib/awful/widget/taglist.lua:342: in function 'awful.widget.taglist.taglist_label'
	(...tail calls...)
	/usr/share/awesome/lib/awful/widget/common.lua:117: in function 'awful.widget.common.list_update'
	/usr/share/awesome/lib/awful/widget/taglist.lua:472: in upvalue 'taglist_update'
	/usr/share/awesome/lib/awful/widget/taglist.lua:570: in function </usr/share/awesome/lib/awful/widget/taglist.lua:568>
	[C]: in function 'xpcall'
	/usr/share/awesome/lib/gears/protected_call.lua:36: in function </usr/share/awesome/lib/gears/protected_call.lua:35>
	(...tail calls...)
	/usr/share/awesome/lib/gears/timer.lua:231: in function </usr/share/awesome/lib/gears/timer.lua:229>
```
I'll add another commit to this PR that stubs `tag.clients`.

src/awesome/tag.rs
```diff
@@ -132,7 +135,30 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
...
+        // Find and remove the tag in/from the list of tags
+        {
+            let tag_ref = &*obj.borrow::<TagState>()? as *const _;
+            let mut found = false;
+            for pair in activated_tags.clone().pairs::<Integer, AnyUserData>() {
+                let (key, value) = pair?;
+                if tag_ref == &*value.borrow::<TagState>()? as *const _ {
+                    found = true;
+                    // Now remove this by shifting everything down...
+                    for index in key .. activated_tags_count-1 {
```
Timidger: The upper bound is exclusive, so this should just be `activated_tags_count` correct?

src/awesome/root.rs
```diff
@@ -84,6 +84,57 @@ fn set_wallpaper<'lua>(_: &'lua Lua, _pattern: *mut cairo_pattern_t) -> rlua::Re
 
 fn tags<'lua>(lua: &'lua Lua, _: ()) -> rlua::Result<Table<'lua>> {
     let table = lua.create_table()?;
-    // TODO FIXME Get tags
+    let activated_tags = lua.named_registry_value::<Table>(super::tag::TAG_LIST)?;
+    for pair in activated_tags.clone().pairs::<Value, Value>() {
```
Timidger: Is this `clone` necessary? It compiles fine without it.

src/awesome/tag.rs
```diff
@@ -132,7 +135,30 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
...
+                        activated_tags.set(index,
+                                           activated_tags.get::<_, Value>(index + 1)?)?;
+                    }
+                    break;
+                }
+            }
+            assert!(found);
```
Timidger: Do we actually want to crash the Lua thread if the tag isn't found?

src/awesome/tag.rs
```diff
@@ -132,7 +135,30 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
...
+    let activated_tags_count = activated_tags.len()?;
+    if val {
+        let index = activated_tags_count + 1;
+        activated_tags.set(index, obj.clone())?;
+    } else {
+        // Find and remove the tag in/from the list of tags
+        {
+            let tag_ref = &*obj.borrow::<TagState>()? as *const _;
+            let mut found = false;
+            for pair in activated_tags.clone().pairs::<Integer, AnyUserData>() {
```
Timidger: Is this clone necessary as well?

src/awesome/root.rs
psychon: Nope, it's not. It only clones the reference anyway. I will remove this.

src/awesome/tag.rs
psychon: Dunno. This assertion should always hold, so I'd say that it is a "correct" use of `assert!` (put differently: "This assert can never fail!" (famous last words)). I don't know if something like `error!` makes more sense here. Or perhaps a `debug_assert!`?

src/awesome/tag.rs
psychon: This one is necessary. `pairs()` consumes the table reference, meaning that I would have to fetch a reference from the table from the registry again without it (so that I can remove the entry from the table). Cloning the reference seems like the better option.

src/awesome/tag.rs
Timidger: Ok, I'm fine with failing early here, it's better than letting a bug slip through.

src/awesome/tag.rs
psychon: Right. I'll add a test that would catch this.

Another problem: This actually breaks the list of tags. Something like the following is missing:
```rust
activated_tags.set(activated_tags_count, Value::Nil)?;
```

psychon:
I added another commit to this to fix the index mis-calculation and squashed the following into the first commit:
```diff
commit 352e6e75a519d96c5ccfa1d4268e56792cb0ead3
Author: Uli Schlachter <psychon@znc.in>
Date:   Sat Mar 24 18:18:38 2018 +0100

    fixup! Keep a list of tags and implement root.tags()

diff --git a/src/awesome/root.rs b/src/awesome/root.rs
index 7ab66a2..8481487 100644
--- a/src/awesome/root.rs
+++ b/src/awesome/root.rs
@@ -85,7 +85,7 @@ fn set_wallpaper<'lua>(_: &'lua Lua, _pattern: *mut cairo_pattern_t) -> rlua::Re
 fn tags<'lua>(lua: &'lua Lua, _: ()) -> rlua::Result<Table<'lua>> {
     let table = lua.create_table()?;
     let activated_tags = lua.named_registry_value::<Table>(super::tag::TAG_LIST)?;
-    for pair in activated_tags.clone().pairs::<Value, Value>() {
+    for pair in activated_tags.pairs::<Value, Value>() {
         let (key, value) = pair?;
         table.set(key, value)?;
     }
```

src/awesome/tag.rs
```diff
@@ -132,7 +139,31 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
...
     }
-    if !val {
+    let activated_tags = lua.named_registry_value::<Table>(TAG_LIST)?;
+    let activated_tags_count = activated_tags.len()?;
+    if val {
+        let index = activated_tags_count + 1;
+        activated_tags.set(index, obj.clone())?;
+    } else {
+        // Find and remove the tag in/from the list of tags
+        {
```
Timidger: This inner scope is unnecessary, because the `borrow` you do "dies" once you convert it to a raw pointer. So there's no need to have the scope.

src/awesome/tag.rs
psychon: I'd also rather have the tags as keys in the table so that I do not have to mess with an array like this... but awesome does not do that either (and "cheats" by using `memmove` to remove an entry, which is a lot cheapter than what has to be done here).

src/awesome/tag.rs
psychon: Sorry, but `$RUST` disagrees. The borrow lives until the end of the block. `rust test` fails:
```
thread 'awesome::root::test::tags_removal' panicked at 'called `Result::unwrap()` on an `Err` value: CallbackError { traceback: "stack traceback:\n\t[C]: in ?\n\t[C]: in metamethod \'__newindex\'\n\t[string \"?\"]:4: in main chunk", cause: CallbackError { traceback: "stack traceback:\n\t[C]: in ?\n\t[C]: in ?\n\t[C]: in ?\n\t[C]: in metamethod \'__newindex\'\n\t[string \"?\"]:4: in main chunk", cause: UserDataBorrowMutError } }'
```
The only relevant information in there is `UserDataBorrowMutError`. This happens in `set_selected` when it calls `get_object_mut`. That's why I added this scope.
I also tried `drop(tag_ref)`, but that was not enough.

Alternatively, the following patch also works (the type annotation with `Ref<TagState>` is just to make it clear what I am dropping here; it is not required):
```diff
diff --git a/src/awesome/tag.rs b/src/awesome/tag.rs
index 973c342..e8199e2 100644
--- a/src/awesome/tag.rs
+++ b/src/awesome/tag.rs
@@ -2,6 +2,7 @@
 
 use std::fmt::{self, Display, Formatter};
 use std::default::Default;
+use std::cell::Ref;
 use rlua::{self, Table, Lua, UserData, ToLua, Value, UserDataMethods, AnyUserData, Integer};
 use super::object::{self, Object, Objectable};
 use super::class::{self, Class, ClassBuilder};
@@ -146,8 +147,10 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
         activated_tags.set(index, obj.clone())?;
     } else {
         // Find and remove the tag in/from the list of tags
-        {
-            let tag_ref = &*obj.borrow::<TagState>()? as *const _;
+        //{
+            let tag_borrow: Ref<TagState> = obj.borrow::<TagState>()?;
+            let tag_ref = &*tag_borrow as *const _;
+            drop(tag_borrow);
             let mut found = false;
             for pair in activated_tags.clone().pairs::<Integer, AnyUserData>() {
                 let (key, value) = pair?;
@@ -163,7 +166,7 @@ fn set_activated<'lua>(lua: &'lua Lua, (obj, val): (AnyUserData<'lua>, bool))
                 }
             }
             assert!(found);
-        }
+        //}
         set_selected(lua, (obj.clone(), false))?;
     }
     signal::emit_object_signal(lua,
```
Would you prefer this version? (Without the type annotation and with re-indenting the block)

Timidger: Ah silly me, I forgot to test this works at runtime. Too used to compile time checks it seems.

This current version is fine, no need to change it.

Timidger:
Thanks!

PR #511, opened 2018-03-25
screen: Implement screen.count()

psychon:
This was tested via print(screen.count()) in init.lua.


Timidger:
Thanks!

Issue #501, opened 2018-02-25
The awesome global breaks after use


psychon:
[snip, I dare to ignore the template]

The following Lua code:
```Lua
awesome.set_preferred_icon_size(42)
awesome.set_preferred_icon_size(21)
```
causes the following (unhelpful) panic:
```
thread 'Lua thread' panicked at 'Unable to load pre-compiled init file: CallbackError { traceback: "stack traceback:\n\t[C]: in ?\n\t[C]: in metamethod \'__index\'\n\t[string \"init.lua <DEFAULT>\"]:2: in main chunk", cause: FromLuaConversionError { from: "nil", to: "table", message: None } }', src/libcore/result.rs:906:4
```
I do not know why the preferred icon size should be saved in Lua (well, okay, perhaps so that it is automatically reset if the code falls back to another config file due to an error...), but `set_preferred_icon_size` looses the associated user value of the `awesome` global, which is the `AwesomeState` instance.

Plus, the call to `AwesomeState::to_lua` in there can be detected by Lua since it creates a new global:
```Lua
local a = awesome
awesome.set_preferred_icon_size(42)
assert(awesome == a)
```
(This cannot be fixed by implementing the `__equals` meta method: `local t = { awesome = 42 } awesome.set_preferred_icon_size(42) assert(t[awesome] == 42)` would still fail.)

Technically, these are two separate issues, but I am putting it into a single issue since fixing the second should also automatically fix the first one.
[I'd assign this issue to myself, but I cannot]

PR #512, opened 2018-03-25
awesome.set_preferred_icon_size: Modify state in place

psychon:
The old function tried to replace the "awesome" global and it did so in
a broken way. Instead, this just borrows the userdatum mutably and
modifies it in place.

Tested via adding the following in this function and checking that it
prints the expected old value when changing the preferred icon size
twice in a row:

  println!("Old preferred was {}, now {}",
  awesome_state.preferred_icon_size, val);

Fixes: https://github.com/way-cooler/way-cooler/issues/501


Timidger:
Thanks!

Issue #513, opened 2018-03-29
Get the Awesome status bar rendering in wlroots 


Timidger:
Thanks to the work from @psychon on implementing parts of #389 and #388, the parts of Way Cooler that host the Lua libraries now run far enough to draw the status bar correctly.

A good first step to porting Way Cooler over to wlroots is to render that bar which will ensure that the pipeline from Lua -> wlroots will work fine and that the wlroots-rs API is usable.

There will be more work necessary to get it ready for real use, but this is a fun, relatively easy first step.

Timidger:
Note: this probably won't utilize surface layers at the onset, as that seems fairly complicated and is very volatile right now in wlroots. This is purely getting a cairo surface and rending it on the screen every frame.

psychon:
TL;DR: The hard part of this is figuring out the communication between threads.

Since it's closely related: Here is a patch for latest master that makes awesome's wibar show up with wlc: https://gist.github.com/psychon/354551ddd101494f361ecf6f8f2d3117

Most of this is relatively straight-forward. The hard part is figuring out how to share the cairo surface between the Lua thread and the main thread. This hack really tries to share things and wraps it in an `Arc` and a `Mutex`. That's a bad idea for many reasons. Thus, I would actually suggest that `Drawable::refresh()` somehow copies the surface's contents over. No idea into what it should copy things.

Timidger:
Copying the surface over could work, though it does seem a little silly to have to do that...

I'll probably do it the "dumb" way first just to get it working and then try to come up with a better system where the compositor thread needs to read it and the Lua thread needs to write to it and the compositor only reads from it once the Lua thread writes to it and then the Lua thread can't touch it until it's read from...maybe?

So you could do it that way which remove all the locks, but I think a lot of it comes down to how expensive such an operation is too. The entire bar isn't going to change often, only small things like e.g the time. You shouldn't need to send the entire buffer for such a small change.

But yeah that's step 2. Step 1 is doing this in wlroots, which will be slightly more involved than wlc.

psychon:
> and then the Lua thread can't touch it until it's read from...maybe?

Problem is, the Lua code has direct access to the `cairo_surface_t*` and doesn't believe in synchronisation... Wrapping a lock around all places that could run Lua code brings us back to "older times" and doesn't work anymore since Lua can directly register callbacks with the GLib main loop thanks to LGI.

Timidger:
[...]

PR #516, opened 2018-04-06
wlroots

### LINKED ISSUE #515: Split source structure into awesome and compositor
With the heavy amount of lifting that is required with wlroots, it would be prudent to have two main modules: `awesome`, where the awesome and Lua interop lives, and `compositor` which will actually do all the things the compositor needs to do.

From the current structure, `awesome` and `lua` will live in the new `awesome`. `compositor` will be entirely new things.

There will of course be `main.rs` which will bind them together, and there will probably be some top level files that will define some constants and channels so the two sides can communicate.

Timidger:
Removes wlc, adds the beginnings of a wlroots compositor.

Removes all of the tiling code, this will all be done in Lua now.

This is the most backward incompatible thing ever. The configuration file has literally been removed. 

Also moved to sr.ht for building, instead of travis.

This is the real start of Awesome on Wayland with Rust. Stay tuned.

Fixes #515 and #513 (see [this branch](https://github.com/way-cooler/way-cooler/tree/working-top-bar-wlroots) for that example)

Here is a taste of things to come...the top bar rendering in Way Cooler, with a weston-terminal beneath it:


Timidger:
Merging this in, even though it's not usable. I look forward to all the bug reports from people using the aur git package.

PR #519, opened 2018-04-15
Fix awesome.load_surface() in various ways

### LINKED ISSUE #518: Wrong color channels on big endian from `load_surface_from_pixbuf`?
I do not actually have a big endian machine to test this on, but I am fairly sure that `load_surface_from_pixbuf` does the Wrong Thing on big endian. Cairo's format are specified as using `uint32_t` for the individual pixels and shifting e.g. the alpha component by 24, so this is endian-specific. However, GdkPixbuf uses a byte-based format and thus is endian-independent.

psychon:
This fixes #518 and details are in the individual commits.

I wonder if it makes sense to add an unit test for this. Such a test would prepare a pixbuf with some specific content, call this function and then do some assertions on the contents of the returned ImageSurface. However, such a test would be relatively verbose, I fear.... Do you want me to write one?

On the various casts that this adds:
- `channels` is either 3 or 4, so safe to cast to usize
- The code in GdkPixbuf ensures that width/height are positive numbers (even though the interface uses signed integers for them...), so casting them to usize is okay (as long as usize is larger than i32, which I am fine with assuming)
- The rest of the added casts are bugfixes, so ought to be fine. :-)

### DIFF
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -349,18 +349,18 @@ pub fn load_surface_from_pixbuf(pixbuf: Pixbuf) -> ImageSurface {
                     // NOTE Four because of the alpha value we ignore
                     cairo_pixels_index2 += 4;
                 } else {
-                    let mut r = pixels[pix_pixels_index];
-                    let mut g = pixels[pix_pixels_index + 1];
-                    let mut b = pixels[pix_pixels_index + 2];
-                    let a = pixels[pix_pixels_index + 3];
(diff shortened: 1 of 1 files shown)


Timidger:
> I wonder if it makes sense to add an unit test for this. Such a test would prepare a pixbuf with some specific content, call this function and then do some assertions on the contents of the returned ImageSurface. However, such a test would be relatively verbose, I fear.... Do you want me to write one?

Once this works it works, so I see no need to write a unit test.

Looks great though, thanks!

PR #520, opened 2018-04-15
awesome/screen: Ensure there is always a screen

psychon:
Currently, this code uses an empty list of outputs and turns that into
an empty list of awesome screens. Since awesome's Lua code reacts quite
badly if no screen exist, this commit adds some further code that just
fakes a static screen with size 1024x768.

This gets rid of some error that happens during startup.

### DIFF
--- a/src/awesome/screen.rs
+++ b/src/awesome/screen.rs
@@ -128,6 +128,18 @@ pub fn init<'lua>(lua: &'lua Lua) -> rlua::Result<Class<'lua>> {
         // TODO Move to Screen impl like the others
         screens.push(screen);
     }
+
+    // If no screens exist, fake one.
+    if screens.is_empty() {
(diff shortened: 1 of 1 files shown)


psychon:
Well, I just did the simplest approach possible to get rid of the Lua error. Note that current master of awesome also makes sure that there always is a screen, but the architecture is different enough that I cannot just copy things.

Sure, it is better to do it only when needed, but currently it is always needed. I would imagine that some PR that adds "proper" screens would get rid of this hack (or hides it in some way).

Alternatively, what do you think is the best place to do this instead? There is not really a central place since at least iterating over all screens and doing `screen[1]` would need this.

Timidger:
Hm, ok. I just don't want later bugs creeping in from this fake screen. Since there's nothing to break right now, I'm going to go ahead and approve this.

Issue #521, opened 2018-04-15
way_cooler.on_init()


psychon:
TL;DR: I think the following three lines of code need some kind of handling. I am not sure what kind of handling. I would just remove them (and the code referencing them), but since I do not know why they are there in the first place, I am reporting this as an issue instead.
```rust
const INIT_LUA_FUNC: &'static str = "way_cooler.on_init()";
const LUA_TERMINATE_CODE: &'static str = "way_cooler.on_terminate()";
const LUA_RESTART_CODE: &'static str = "way_cooler.on_restart()";
```
Also look at the end of the debug log for the exact error.

# Minimum issue requirements

**Search the issue tracker first, as it's very possible your question has been asked and answered**

Nope, I do not think I see anything relevant.

Please provide, at minimum:
* Way Cooler version (obtained via `way-cooler --version`
```
zsh: command not found: way-cooler
```
and
```
$ target/debug/way-cooler --version
Way Cooler 0.8.0 @ 23ab3331be821e7ff073e834f17af58f15a740e6
```
* rc file
  - None
* Debug log
```
Launching way-cooler...
 INFO [way_cooler] src/main.rs:206 Logger initialized 
 INFO [way_cooler] src/main.rs:211 USER: psychon 
 INFO [way_cooler] src/main.rs:211 XDG_SEAT: seat0 
 INFO [way_cooler] src/main.rs:211 XDG_SESSION_TYPE: x11 
 INFO [way_cooler] src/main.rs:211 HOME: /home/psychon 
 INFO [way_cooler] src/main.rs:211 DESKTOP_SESSION: lightdm-xsession 
 INFO [way_cooler] src/main.rs:211 XDG_SEAT_PATH: /org/freedesktop/DisplayManager/Seat0 
 INFO [way_cooler] src/main.rs:211 DBUS_SESSION_BUS_ADDRESS: unix:path=/run/user/1000/bus 
 INFO [way_cooler] src/main.rs:211 LOGNAME: psychon 
 INFO [way_cooler] src/main.rs:211 XDG_SESSION_ID: 2 
 INFO [way_cooler] src/main.rs:211 PATH: /usr/lib/ccache:/home/psychon/bin:/home/psychon/.luarocks/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games 
 INFO [way_cooler] src/main.rs:211 GDM_LANG: de_DE.utf8 
[...]

PR #522, opened 2018-04-15
Fix Lua startup

psychon:
This was running way_cooler.on_init(), which failed and caused an error.
Fix this by not running on_init(). While at it, this also removes
on_terminate() and on_restart().

Also, this inlines init() and lua_init() into their only caller.

Fixes: https://github.com/way-cooler/way-cooler/issues/521


Timidger:
Looks good to me, thanks!

PR #524, opened 2018-04-27
Remove dedicated Lua thread

psychon:
Before this commit, way-cooler had two threads.

One thread ran the GLib main loop (through g_main_loop_run()) and ran
all Lua code.  Using this main loop is necessary since Lua code has
direct access to this via LGI and can e.g. register callbacks.

The second thread ran the wayland event loop (through wl_display_run())
and was responsible for all compositor-related things.

This commit merges the two threads by making the GLib main loop "run
everything". This is possible, because the wayland event loop is
actually based on a simple epoll file descriptor. This commit adds some
C code that registers this file descriptor with GLib and dispatches the
wayland event loop whenever the file descriptor becomes readable.


psychon:
(Hm, at some point there should be a big refactoring and removal of some dead code in way-cooler. E.g. this commit makes the whole "dance" to send requests between the two threads unnecessary: Just call the necessary function directly. However, today is not that day...)
(Also, eventually it would of course be nice to get rid of the C code and make it possible to write new `GSource` implementations via `glib-rs`. However, for now I am perfectly fine with doing this call into C-land once to get things working. Plus, `GSource` is not exactly bindings friendly.)

PR #525, opened 2018-05-01
Proper top bar rendering

### LINKED ISSUE #513: Get the Awesome status bar rendering in wlroots

Timidger:
Yay the top bar is rendering on master (once this is merged) in a way that isn't stupid.

This is dependent on https://github.com/swaywm/wlroots-rs/pull/147 getting merged in so that we can continue to track master of wlroots-rs.

# Rendering
This sets up the rendering pipeline to load the Lua context and get the drawins. If they are ready to be rendered it renders them to the screen.

Things not currently handled:
* The time doesn't update, because the drawin isn't being updated for...some reason.
* It doesn't update when the screen is resized since we don't pass that event down yet.
* It can't be clicked or interacted with since that also hasn't been defined.

Fixes #513


src/compositor/output/output.rs
```diff
@@ -40,5 +44,34 @@ impl OutputHandler for Output {
...
+            let mut lock = drawin.image();
+            let mut image = match lock {
+                Err(_) => continue,
+                Ok(ref none) if none.is_none() => continue,
+                Ok(ref mut image) => image.as_mut().unwrap()
+            };
+            {
+                let cr = Context::new(&*image);
+                cr.set_source_surface(&*image, 0.0, 0.0);
+                cr.paint();
```
psychon: Uhm... is this painting a surface to itself?
- Why? (Copy&paste from something older?)
- That does not have defined semantics in Cairo.

Timidger: Yeah no idea why that code is there...Must be _oolllllddd_ code from the old wlc rendering code.

PR #526, opened 2018-05-12
use rlua builtin lua

acrisci:
Without this change, I get:

```
 = note: /usr/bin/ld: cannot find -llua5.3
          collect2: error: ld returned 1 exit status
```

Full build output available on request.

I think it's looking for [this build artifact](https://github.com/chucklefish/rlua/blob/31fa9173ae67628df7b7eaad4610263220d3a2f6/build.rs#L63).

Maybe on your system, `-llua5.3` will pick up the system lua? Mine won't.

```
$ pkg-config --libs lua        
-llua -lm -ldl
```

If there's a better way to fix this let me know.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ wlroots = { path = "./wlroots-rs" }
 lazy_static = "0.2"
 log = "0.3"
 env_logger = "0.3"
-rlua = { version = "0.12.0", default-features = false }
+rlua = "0.12.0"
 bitflags = "0.7"
 nix = "0.6"
 getopts = "0.2"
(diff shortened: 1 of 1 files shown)


Timidger:
We require the system Lua because awesome libs use LGI which is easier to setup for the system Lua.

You should be able to get ua53 from your package manager and it should "just work".

psychon:
@acrisci What distro are you on? Looks like some fun with `build.rs` is required (in rlua?) to find non-static Lua.

Timidger:
He's on Fedora, as reported on IRC.

acrisci:
Yeah, in rlua it needs to use pkg-config to find my system lua lib because it's using the wrong linker flag. I need to find out where it's generating the `-llua5.3` and make it say `-llua` on my system.

acrisci:
Made [this PR](https://github.com/chucklefish/rlua/pull/79) on rlua.

PR #530, opened 2018-05-17
update to latest wlroots-rs

acrisci:


Timidger:
Grrr submodules aren't updating correctly. Can't wait until I can put that on crates.io.

Thanks for doing this! I just merged in that other PR so you might want to rebase.

It looks like wlroots-sys got updated as well by mistake so the wlroots-rs package isn't building. Let's update the version in wlroots-rs first and then update wlroots-rs here so we don't have to do this twice so soon.

acrisci:
Oh it looks like everything I did here was already done in that PR.

PR #531, opened 2018-05-18
make the cursor work

acrisci:
Update to latest wlroots-rs, use the XCursorManager to make the cursor show up, and add pointer motion absolute event handler.

TODO:
* [ ] generalize cursor updates for relative/absolute motion into a single function

### DIFF
--- a/src/compositor/input/pointer.rs
+++ b/src/compositor/input/pointer.rs
@@ -20,7 +20,7 @@ impl PointerHandler for Pointer {
             with_handles!([(cursor: {&mut *cursor})] => {
                 let (x, y) = event.pos();
                 cursor.warp_absolute(event.device(), x, y);
-                update_cursor_position(cursor, seat, views);
+                update_view_position(cursor, seat, views);
             }).expect("Cursor was destroyed");
         }).unwrap();
     }
@@ -35,7 +35,7 @@ impl PointerHandler for Pointer {
             with_handles!([(cursor: {&mut *cursor})] => {
                 let (x, y) = event.delta();
                 cursor.move_to(event.device(), x, y);
-                update_cursor_position(cursor, seat, views);
+                update_view_position(cursor, seat, views);
             }).expect("Cursor was destroyed");
         }).unwrap();
     }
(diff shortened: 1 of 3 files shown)


Timidger:
Looks good, should be pretty easy to dry up that code. If there's issues you run into though feel free to point them out here

acrisci:
Ok what do you think about that?

The idea here is to not pass handles to the helper functions. I think the other way could get a bit hairy because it increases the overhead of helper functions since they need to unwrap everything in the server each time you want to add a new one which could get nasty a few functions deep in the stack.

This way the only time you ever unwrap handles is at the wlroots entry point and everybody else works with the real things.

src/compositor/input/pointer.rs
```diff
@@ -40,35 +48,52 @@ impl PointerHandler for Pointer {
...
+                    focus_under_pointer(seat, &mut **keyboards, None).expect("Could not focus view");
+                }
+            }).unwrap()
+        }).unwrap()
+    }
+}
+
+/// After the cursor has been warped, do all the things needed to change the cursor from the old
+/// position to the new position.
+fn update_cursor_position(cursor: &mut Cursor, seat: &mut Seat, views: &mut [View]) {
```
Timidger: This doesn't actually update the cursor position though. It updates the views to be where the cursor is we should move the window (which in this proto-way-cooler phase it always is). If the idea is it should do other generic stuff then I think we shouldn't try to shoe horn those things into into this function and just focus on updating views.

This function name and doc comment should be changed to reflect that. (e.g. `move_focused_views_to_cursor` or something to that effect).

acrisci: It will need to do other stuff too though like the resize.

Timidger: Ok, but that isn't really updating the cursor position then is it? It should be `update_view_position`

acrisci: We can call it that for now and then reevaluate the name later once you see the stuff that's going to go in there.

Timidger:
Thanks!

PR #532, opened 2018-05-20
Send pointer motion

acrisci:
Send pointer motion to the surface under the pointer.

Add the view `activate()` function.

Test by opening a window and clicking around to see it respond to pointer events.


src/compositor/input/pointer.rs
```diff
@@ -147,7 +171,6 @@ fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
                                       (keyboard: {keyboard})] => {
                     match shell.state() {
                         Some(&mut TopLevel(ref mut toplevel)) => {
-                            // TODO Don't send this for each keyboard!
```
acrisci: This has no performance impact.

Timidger: This is not a performance thing. We don't want to set all the views activated, only the focused one. The others should be deactivated

PR #534, opened 2018-05-20
Trigger Awesome keybindings for key events

Timidger:
With this patch keybindings work properly :tada:. Too bad they fail when you press them since nothing else is set up correctly ;).

# TODO
- [x] Error handling
- [x] ~Using a hashmap instead to store the global keybindings~ Can't do because this is part of the public API (you pass a "list", and we should return one in the same order).
- [x] Double checking the error handling used in `xcb` crate is correct for `xcb_key_symbols_get_keysym`
- [x] Make a deep clone of the getter for the root keybindings.
- [x] Parse numbers the same way awesome does it (e.g they handle `#[0-9]` in a specific way)

### DIFF
--- a/src/compositor/input/keyboard.rs
+++ b/src/compositor/input/keyboard.rs
@@ -68,7 +68,7 @@ impl KeyboardHandler for Keyboard {
     }
 }
 
-/// Emits the Awesome keybindinsg.
+/// Emits the Awesome keybindings.
 fn emit_awesome_keybindings(lua: &Lua,
                             event: &KeyEvent,
                             event_modifiers: KeyboardModifier)
--- a/src/compositor/input/keyboard.rs
+++ b/src/compositor/input/keyboard.rs
@@ -1,7 +1,11 @@
-use compositor::Server;
+use rlua;
 use wlroots::{key_events::KeyEvent,
               xkbcommon::xkb::{KEY_Escape, KEY_Super_L, KEY_Super_R}, CompositorHandle,
               KeyboardHandle, KeyboardHandler, WLR_KEY_PRESSED};
+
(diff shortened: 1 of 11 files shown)


acrisci:
Can you give a test config I can use to test the bindings?

Timidger:
The default config should have that already.

However it uses `meta` for the default. I used `"Control"` for the meta key to test in a nested environment. When you press e.g. `ctrl+s` when (or `super+s` with the default config) then it will show an error showing it tried to do something with the client library but failed (so it shows the keybinding part works).

Timidger:
Right now it seems to be panicking as well because it tries to restart when things don't work. That might be because it's a trigger from awesome that causes it to try to shutdown. When that happens it tries to mutably borrow the LUA global but it can't because we have immutably borrowed it to call the callback that gets us to that point.

We can get around that in two ways:
1) Crashing (as we would now, if we had the right panic catchers set up with wlroots-rs)
2) Handling it by saving that we _want_ to reset the Lua thread until we are back at the top of the glib loop call stack. That could cause a different error though as there's way to unwind back up that smoothly without executing a bunch more probably wrong code.

src/compositor/input/keyboard.rs
```diff
@@ -27,7 +25,47 @@ impl KeyboardHandler for Keyboard {
...
+        LUA.with(|lua| {
+                     let lua = lua.borrow();
+                     if let Err(err) = emit_awesome_keybindings(&*lua, event, modifiers) {
+                         warn!("Could not emit binding for {}: {:?}", event.keycode(), err);
+                     }
+                 });
+    }
+}
+
+/// Emits the Awesome keybindinsg.
```
acrisci: the what?

Timidger: Man I'm really bad at spelling. Thanks

PR #537, opened 2018-05-22
remove seat focus on surface destroy

acrisci:
Fixes a crash when you click somewhere after the focused view has been destroyed. The seat focus is a copy of the view.

I would like to handle removing the focused view from the seat in the first block, but I can't access the seat there. Is there a better way to do that?

### DIFF
--- a/src/compositor/shells/xdg_v6.rs
+++ b/src/compositor/shells/xdg_v6.rs
@@ -52,6 +52,16 @@ impl XdgV6ShellManagerHandler for XdgV6ShellManager {
             {
                 server.views.remove(pos);
             }
+
+            let focused = if let Some(ref focused) = server.seat.focused {
+                focused.shell == destroyed_shell
+            } else {
+                false
+            };
+
+            if focused {
+                server.seat.focused = None;
+            }
         }).unwrap();
     }
 }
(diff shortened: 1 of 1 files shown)


Timidger:
Two things:

1. It's slightly more rustic to say `server.seat.focused.take()` instead of `server.seat.focused = None`.

2. There's a couple ways you can go with this. I suggest really exploing [the Option page](https://doc.rust-lang.org/std/option/enum.Option.html). I look at it often when I try to do something with `Option` that's a little awkward like this. I came up with this, you might be able to do better:

```rust
            server.seat.focused = server.seat.focused.take().and_then(|focused| {
                if focused.shell == destroyed_shell {
                    None
                } else {
                    Some(focused)
                }
            });
```

I would have used [filter](https://doc.rust-lang.org/std/option/enum.Option.html#method.filter) but, like every other fucking thing in Rust, it hasn't stabilized because they had to [bikeshed over the name](https://github.com/rust-lang/rust/issues/45860).

acrisci:
Yeah I think that's good. I'm not entirely clear on why the `take()` is necessary to compile though.

Timidger:
The take actually moves any Some out of the Option (leaving None in its place, or just returning None if it was None). Otherwise you get into borrowing problems.

lfairy:
`Option::filter` will be stable [next Thursday](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1270-2018-06-21), so we can use it then :)

psychon:
...which makes me ask what the backwards compatibility plan is. Will there ever be support for Rust version `n-1` and perhaps even `n-10`?
(I am just asking "for a friend" who happens to use `apt-get install rustc cargo` and had to resort to Debian experimental to build way-cooler).

PR #538, opened 2018-05-23
run cargo fmt on code

acrisci:


Timidger:
Thanks. I think I've had the nightly installed on my machine and it overrode the stable version.

Can't wait until rustfmt is stable.

PR #539, opened 2018-05-23
basic move view implementation

acrisci:
To test, try to move a view around by the titlebar.

### DIFF
--- a/src/compositor/shells/xdg_v6.rs
+++ b/src/compositor/shells/xdg_v6.rs
@@ -1,6 +1,8 @@
-use compositor::{Server, Shell, View};
-use wlroots::{CompositorHandle, XdgV6ShellHandler, XdgV6ShellManagerHandler, XdgV6ShellState::*,
-              XdgV6ShellSurfaceHandle};
+use compositor::{Action, Server, Shell, View};
+use wlroots::{CompositorHandle, Origin, SurfaceHandle, XdgV6ShellHandler,
+              XdgV6ShellManagerHandler, XdgV6ShellState::*, XdgV6ShellSurfaceHandle};
+
+use wlroots::xdg_shell_v6_events::MoveEvent;
 
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct XdgV6 {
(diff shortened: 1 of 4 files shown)


acrisci:
It's getting a little cumbersome to work with clones of views.

src/compositor/shells/xdg_v6.rs
```diff
@@ -13,7 +15,33 @@ impl XdgV6 {
...
 
-impl XdgV6ShellHandler for XdgV6 {}
+impl XdgV6ShellHandler for XdgV6 {
+    fn move_request(&mut self,
+                    compositor: CompositorHandle,
+                    surface: SurfaceHandle,
+                    shell_surface: XdgV6ShellSurfaceHandle,
+                    event: &MoveEvent) {
+        with_handles!([(compositor: {compositor})] => {
+            let server: &mut Server = compositor.data.downcast_mut().unwrap();
```
Timidger: Use `compositor.into()` here instead of the downcast_mut. It makes it cleaner.

src/compositor/shells/xdg_v6.rs
```diff
@@ -13,7 +15,33 @@ impl XdgV6 {
...
+            let ref mut seat = server.seat;
+            let ref mut cursor = server.cursor;
+
+            seat.focused = seat.focused.take().and_then(|view| {
```
Timidger: You could do this better like this:

```rust
            if let Some(ref mut view) = seat.focused {
                let shell: Shell = shell_surface.into();
                let action = &mut seat.action;
                if view.shell == shell {
                    with_handles!([(cursor: {cursor})] => {
                        let (lx, ly) = cursor.coords();
                        let Origin { x: shell_x, y: shell_y } = view.origin;
                        let (view_sx, view_sy) = (lx - shell_x as f64, ly - shell_y as f64);
                        let start = Origin::new(view_sx as _, view_sy as _);
                        *action = Some(Action::Moving { start: start });
                    }).unwrap();
                }
            }
```

Note that we get around the closure borrowing the entire struct by taking a mutable reference to the field we want to mutate.

acrisci:
> What is it about the cloning that you don't like?

I have multiple representations to the real thing, but the only one that matters is the one in the list of views that the server has. So if I want to change view state, I have to mutate the right one which only the server can do. And when the view is destroyed, I have to clean up all the copies in that destroy callback.

src/compositor/shells/xdg_v6.rs
acrisci: Yeah this will be needed in multiple places as well, so I'll probably make it a server method.

PR #540, opened 2018-05-24
Shutdown properly on panic

Timidger:
This allows Way Cooler to shutdown properly on a panic.

# TODO
- [x] Wait for the wlroots-rs changes to be merged into master.

### DIFF
--- a/src/compositor/input/keyboard.rs
+++ b/src/compositor/input/keyboard.rs
@@ -1,5 +1,5 @@
 use rlua::{self, Lua};
-use wlroots::{self, key_events::KeyEvent,
+use wlroots::{key_events::KeyEvent,
               xkbcommon::xkb::{KEY_Escape, KEY_Super_L, KEY_Super_R}, CompositorHandle,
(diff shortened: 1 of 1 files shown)


src/awesome/root.rs
```diff
@@ -93,6 +95,29 @@ fn tags<'lua>(lua: &'lua Lua, _: ()) -> rlua::Result<Table<'lua>> {
...
 }
 
+/// Get or set global key bindings.
+///
+/// These bindings will be available when you press keys on the root window.
+fn root_keys<'lua>(lua: &'lua Lua, key_array: rlua::Value) -> rlua::Result<rlua::Value<'lua>> {
+    match key_array {
+        // Set the global keys
+        Value::Table(key_array) => {
+            lua.globals().set(ROOT_KEYS_HANDLE, key_array)?;
```
psychon: How about `set_named_registry_value` instead of saving this in the global table?

psychon: Never mind, the next commit seems to do just that

PR #544, opened 2018-05-25
meta key move

### LINKED ISSUE #543: keyboard on_key() handler sends incorrect mod mask
Pressing a mod is not immediately reflected in `keyboard.get_modifier_masks()` which is used to send the modifier state to clients.

To reproduce:

Hold shift and type into a weston-terminal window. The first key will be lowercase and only the subsequent keys after will be capitalized.

acrisci:
Basic meta key detection for the seat. To test, move a view around with the meta key depressed. With this, we can test that the rest of the meta key move code paths work correctly.

Open issues:

* How is meta key configured?
* If multiple meta keys are pressed, and one is released, then the meta move will be released even though the other meta key is pressed. This problem comes from us not keeping a list of currently pressed keysyms.

fixes #543

### DIFF
--- a/src/compositor/input/keyboard.rs
+++ b/src/compositor/input/keyboard.rs
@@ -1,7 +1,13 @@
 use compositor::Server;
-use wlroots::{key_events::KeyEvent, xkbcommon::xkb::{KEY_Escape, KEY_F1}, CompositorHandle,
+use wlroots::{key_events::KeyEvent,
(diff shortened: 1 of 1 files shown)


Timidger:
> How is meta configured?

See the configuration file. The meta key is defined and passed in with the rest of the keys. See in particular line 454.

Timidger:
My thoughts exactly

PR #546, opened 2018-05-27
render and input fixes

acrisci:
Trying to get mostly usable shell surfaces.

* [x] Render focused view on top of the other views
* [x] Revert the focus to the next one in the list when a focused shell surface closes
* [x] Set the focus to a newly created shell surface
* [x] General keyboard fixes
* [x] Add view on map, not create
* [x] reference count the views
* [x] make nicer view abstractions
* [x] Update to latest wlroots-rs

### DIFF
--- a/src/compositor/input/pointer.rs
+++ b/src/compositor/input/pointer.rs
@@ -56,19 +56,27 @@ impl PointerHandler for Pointer {
                     send_pointer_button(seat, event).expect("Could not send pointer button");
                     return
                 }
-                let (view, _surface, _sx, _sy) = view_at_pointer(views, cursor);
+                let view = {
+                    let (view, _surface, _sx, _sy) = view_at_pointer(views, cursor);
+                    if let Some(view) = view {
+                        Some(view.clone())
+                    } else{
+                        None
+                    }
+                };
+
                 match view {
-                    Some(view) => {
-                        focus_under_pointer(seat, &mut **keyboards, { &mut *view }).expect("Could not focus \
+                    Some(mut view) => {
+                        focus_under_pointer(seat, &mut **keyboards, &mut view, views).expect("Could not focus \
                                                                                     view");
                         let meta_held_down = seat.meta;
                         if meta_held_down && event.button() == BTN_LEFT {
-                            move_view(seat, cursor, view, None);
+                            move_view(seat, cursor, &mut view, None);
                         }
                         send_pointer_button(seat, event).expect("Could not send pointer button");
                     },
                     None => {
-                        focus_under_pointer(seat, &mut **keyboards, None).expect("Could not focus view");
+                        focus_under_pointer(seat, &mut **keyboards, None, views).expect("Could not focus view");
                     }
                 }
             }).unwrap()
@@ -135,7 +143,8 @@ fn view_at_pointer<'view>(views: &'view mut [View],
 /// Focus the view under the pointer.
 fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
                                  keyboards: &mut [KeyboardHandle],
-                                 view: V)
+                                 view: V,
+                                 views: &mut Vec<View>)
                                  -> HandleResult<()>
     where V: Into<Option<&'view mut View>>
 {
@@ -149,7 +158,7 @@ fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
             })
         }
         Some(view) => {
-            if let Some(mut focused_view) = seat.focused.take() {
+            if let Some(mut focused_view) = seat.focused.as_mut() {
                 if focused_view.shell == view.shell {
                     return Ok(());
                 }
@@ -157,6 +166,12 @@ fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
             }
             seat.focused = Some(view.clone());
             view.activate(true);
+
+            if let Some(idx) = views.iter().position(|ref v| v.shell == view.shell) {
+                let v = views.remove(idx);
+                views.insert(0, v);
+            }
+
             for keyboard in { &mut *keyboards } {
                 with_handles!([(seat: {&mut seat.seat}),
                 (surface: {view.surface()}),
--- a/src/compositor/input/pointer.rs
+++ b/src/compositor/input/pointer.rs
@@ -160,7 +160,7 @@ fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
         Some(view) => {
             if let Some(mut focused_view) = seat.focused.as_mut() {
                 if focused_view.shell == view.shell {
-                    return Ok(());
+                    return Ok(())
                 }
                 focused_view.activate(false);
             }
--- a/src/compositor/input/pointer.rs
+++ b/src/compositor/input/pointer.rs
@@ -55,7 +55,7 @@ impl PointerHandler for Pointer {
             with_handles!([(cursor: {&mut *cursor})] => {
                 if event.state() == WLR_BUTTON_RELEASED {
                     seat.action = None;
-                    send_pointer_button(seat, event).expect("Could not send pointer button");
+                    seat.send_button(event);
                     return
                 }
                 let view = {
@@ -75,7 +75,7 @@ impl PointerHandler for Pointer {
                         if meta_held_down && event.button() == BTN_LEFT {
                             move_view(seat, cursor, view, None);
                         }
-                        send_pointer_button(seat, event).expect("Could not send pointer button");
+                        seat.send_button(event);
                     },
                     None => {
                         seat.clear_focus();
@@ -163,12 +163,4 @@ fn move_view<O>(seat: &mut compositor::Seat, cursor: &mut Cursor, view: Rc<View>
             view.origin.replace(pos);
         }
     };
-}
-
-fn send_pointer_button(seat: &mut compositor::Seat, event: &ButtonEvent) -> HandleResult<()> {
-    with_handles!([(seat: {&mut seat.seat})] => {
-        seat.pointer_notify_button(Duration::from_millis(event.time_msec() as _),
-                                   event.button(),
-                                   event.state() as u32);
-    })
 }
--- a/src/compositor/input/pointer.rs
+++ b/src/compositor/input/pointer.rs
@@ -5,6 +5,7 @@ use wlroots::{pointer_events::*, CompositorHandle, HandleResult, KeyboardHandle,
               PointerHandle, PointerHandler, SurfaceHandle, WLR_BUTTON_RELEASED};
 
 use compositor::{self, Action, Seat, Server, Shell, View};
+use std::rc::Rc;
 
 #[derive(Debug, Default)]
 pub struct Pointer;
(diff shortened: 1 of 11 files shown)


Timidger:
The shadows on weston-terminal look really bizzare on the wayland backend.

Is that a bug in wlroots? Way Cooler shouldn't be doing anything strange with that AFAIK.

Timidger:
Holding down the mouse to highlight some text, moving it out of the window, and then releasing it causes it to still be "stuck". We need to send the click release even after the client no longer has the mouse.

src/compositor/input/pointer.rs
```diff
@@ -56,19 +56,27 @@ impl PointerHandler for Pointer {
                     send_pointer_button(seat, event).expect("Could not send pointer button");
                     return
                 }
-                let (view, _surface, _sx, _sy) = view_at_pointer(views, cursor);
+                let view = {
+                    let (view, _surface, _sx, _sy) = view_at_pointer(views, cursor);
+                    if let Some(view) = view {
+                        Some(view.clone())
```
Timidger: You can just `clone` the view if the inner type implements `clone`.

acrisci:
Yeah it's a [rootston bug](https://github.com/swaywm/wlroots/issues/613) too so I guess I'll have to fix it there too.

src/compositor/input/pointer.rs
```diff
@@ -149,8 +158,20 @@ fn focus_under_pointer<'view, V>(seat: &mut compositor::Seat,
...
+                    return Ok(())
+                }
+                focused_view.activate(false);
+            }
             seat.focused = Some(view.clone());
             view.activate(true);
+
+            if let Some(idx) = views.iter().position(|ref v| v.shell == view.shell) {
+                let v = views.remove(idx);
+                views.insert(0, v);
```
Timidger: So wait, the focused view is always the top of the list? So it's a stack?

I don't really like this, as it's not very performant and seems rather hacky. I also see why you can't just have use a `focused`, since that might differ from what's in the view list.

If upgrading a handle didn't require an `&mut` you could just use `Rc` here and call it a day. I'll have to re-investigate why I made it require `&mut`. I believe I found a safety hole when it was simply `&`, but it might be me being over zealous.

src/compositor/shells/xdg_v6.rs
```diff
@@ -60,8 +60,23 @@ impl XdgV6ShellManagerHandler for XdgV6ShellManager {
         if is_toplevel {
             with_handles!([(compositor: {compositor})] => {
                 let server: &mut Server = compositor.into();
-                server.views
-                    .push(View::new(Shell::XdgV6(shell_surface.into())));
+                let Server { ref mut seat,
+                             ref mut views,
+                             .. } = *server;
+                views.insert(0, View::new(Shell::XdgV6(shell_surface.into())));
```
Timidger: If this is the data structure we want to go with, I think having a linked list might be more performant (or something that doesn't require shifting all the of the elements over everytime focus changes or when a new view is added).

Though again I think this approach could use a lot of work so don't waste time switching to that unless we can find a better solution.

acrisci: You could push it to the back and then render the list forward and find the surface for input backward.

src/compositor/input/pointer.rs
Timidger: [Found the issue where I made it &mut](https://github.com/swaywm/wlroots-rs/pull/76).

It's not UB, I'm not sure what I was talking about. If we switch the handles to only require a `&` to upgrade the handle then, at the cost of making it slightly easier to shoot yourself in the foot (e.g by immediantly re-upgrading from the same handle, which will just cause an error and never be an issue) we can have `Rc<View>` stored at `compositor::Seat.focused`  that can also be inside the `Server.views` list. That way they'll always be synced up. 

For the values inside `View` that still need to be mutated and shared (e.g right now just the `Origin`) we can add `Cell`s (in the case of the `Origin`, which is more ergonomic) or `RefCell`s in the future.

So the view will look like this:


```rust
pub struct View {
    pub shell: Shell,
    /// Allows us to mutate it behind an Rc
    pub origin: Cell<Origin>
}
```

and the seat will look like this:

```rust
pub struct Seat {
    pub seat: SeatHandle,
    /// Thanks to zero cost abstractions, 
    /// this Option<Rc> has no overhead compared to a raw pointer in terms of size.
    pub focused: Option<Rc<View>>,
    pub action: Option<Action>,
    pub meta: bool
}
```

and finally the `Server` will look like this:

```rust
pub struct Server {
    // ... pointless fields elided

    pub views: Vec<Rc<View>>
}
```

src/compositor/shells/xdg_v6.rs
Timidger: Focusing on another view still requires reordering everything (in the worst case) which is inefficient in a Vector.

psychon:
> Holding down the mouse to highlight some text, moving it out of the window, and then releasing it causes it to still be "stuck". We need to send the click release even after the client no longer has the mouse.

Sounds like what X11 calls a passive grab (which is exactly what you describe: A grab in general is an "dear X11 server, please send me all input events exclusively"; An active grab is actively started by a client while a passive grab is the client saying "I am interested in the left mouse button" and then you basically get the behaviour outlined above)

acrisci:
> Sounds like what X11 calls a passive grab

Yeah that's a lot like what's happening. Grabs in wayland are specified by the shell protocol. For instance, a popup will take a grab when it needs to get all of the pointer events so it can close when you click on a location outside of the popup (normally it would only receive input events that are in the popup itself). Grabs in wlroots are managed by the seat. The popup requests a grab from the seat, and then when you call the `seat.notify_[input event]` function, the seat sends the input to the surface with the grab for you automatically so you don't have to worry about it.

What's happening here is that we are leaving the pointer surface with a button depressed. When we leave the surface, it loses pointer focus. If the seat does not have pointer focus, when we notify to send the button, it will not send the event because it doesn't have a surface to send it to. The solution is to give the surface a grab of sorts while the button is down, and only send pointer enter/leave events when no button is pressed on the pointer input device.

acrisci:
Needs [wlroots-rs#167](https://github.com/swaywm/wlroots-rs/pull/167)

acrisci:
This needs the lenient-handle-upgrades branch to work.

acrisci:
Ok, this isn't terrible anymore. I think I can move onto the next feature and do more abstractions as they become convenient.

Timidger:
Ok. Will merge lenient handle upgrades tonight and then review this patch

Timidger:
lenient-handle-upgrades has been merged in.

src/compositor/seat.rs
```diff
@@ -26,6 +29,49 @@ impl Seat {
...
+            focused_view.activate(false);
+        }
+        with_handles!([(seat: {&mut self.seat})] => {
+            seat.keyboard_clear_focus();
+        }).unwrap();
+    }
+
+    pub fn focus_view(&mut self, view: Rc<View>, views: &mut Vec<Rc<View>>) {
+        if let Some(ref focused) = self.focused {
+            if focused == &view {
```
Timidger: (I don't believe this will trigger a move. If it does then don't do this).

`*focused == view` would be clearer

src/compositor/seat.rs
```diff
@@ -26,6 +29,49 @@ impl Seat {
...
+                return
+            }
+            focused.activate(false);
+        }
+        self.focused = Some(view.clone());
+        view.activate(true);
+
+        if let Some(idx) = views.iter().position(|v| *v == view) {
+            let v = views.remove(idx);
+            views.insert(0, v);
```
Timidger: If we are going with this strategy, we should switch to a [linked list](https://doc.rust-lang.org/std/collections/struct.LinkedList.html). 

Note I have **not** profiled the two approaches, I'm just going with my gut here. It's possible this will decrease performance.

We can leave this off for another patch.

src/compositor/shells/xdg_v6.rs
```diff
@@ -41,15 +42,11 @@ impl XdgV6ShellHandler for XdgV6 {
...
-
-pub struct XdgV6ShellManager;
 
-impl XdgV6ShellManagerHandler for XdgV6ShellManager {
-    fn new_surface(&mut self,
+    fn map_request(&mut self,
                    compositor: CompositorHandle,
-                   mut shell_surface: XdgV6ShellSurfaceHandle)
-                   -> Option<Box<XdgV6ShellHandler>> {
+                   _surface: SurfaceHandle,
```
Timidger: You can just use `_` here since the name `surface` doesn't add much.

src/main.rs
```diff
@@ -203,6 +203,7 @@ fn detect_raspi() {
 
 /// Initializes the logging system.
 pub fn init_logs() {
+    //wlroots::utils::init_logging(wlroots::utils::L_DEBUG, None);
```
Timidger: Why is this commented out?

src/compositor/input/pointer.rs
```diff
@@ -111,70 +119,37 @@ fn update_view_position(cursor: &mut Cursor, seat: &mut Seat, views: &mut [View]
...
-        }
-    }
-}
-
 /// Start moving a view if you haven't already by passing `start: None`.
 ///
 /// Otherwise, update the view position relative to where the move started,
 /// which is provided by Action::Moving.
-fn move_view<O>(seat: &mut compositor::Seat, cursor: &mut Cursor, view: &mut View, start: O)
+fn move_view<O>(seat: &mut compositor::Seat, cursor: &mut Cursor, view: Rc<View>, start: O)
```
Timidger: If you can use `&View` instead of `Rc<View>` you should (and I believe you can here).

`Rc<T>` derefs into a `&T`

src/compositor/seat.rs
psychon: Unscientific data point: Awesome's C code does not have any linked lists and uses a homegrown vector / array list for all its lists (stacking order, 'visible' order, configured key bindings, ...). This only ever was a problem with "stupid" code that repeatedly scanned the list from the beginning and caused quadratic behaviour that way.

acrisci: what about a VecDeque?

Timidger: Sounds good to me.

src/main.rs
acrisci: Do you want it on by default or toggle?

Timidger: Let's put it on by default for now. We can redecide later close to release, though I can't think of a scenario where it makes sense to disable it.

src/compositor/input/pointer.rs
```diff
@@ -80,16 +88,16 @@ impl PointerHandler for Pointer {
...
     match seat.action {
         Some(Action::Moving { start }) => {
-            if let Some(focused_shell) = seat.focused.as_ref().map(|f| f.shell.clone()) {
-                for v in views {
-                    if v.shell == focused_shell {
-                        move_view(seat, cursor, v, start);
-                    }
-                }
-            }
+            seat.focused = seat.focused.take().map(|f| {
```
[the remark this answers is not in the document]
acrisci: I wish that worked because it's a common problem for me. I can't take a reference to `focused` without borrowing the seat. And I need to change the action of the seat so it needs to be mutable.

```
  --> src/compositor/input/pointer.rs:87:27
   |
86 |             if let Some(ref focused) = seat.focused {
   |                         ----------- immutable borrow occurs here
87 |                 move_view(seat, cursor, focused.clone(), start);
   |                           ^^^^ mutable borrow occurs here
88 |             }
   |             - immutable borrow ends here
```

src/compositor/seat.rs
acrisci: Just found out that you can't slice a VecDeque though so I don't think you can take slices for those pointer methods.

src/compositor/input/pointer.rs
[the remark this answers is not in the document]
Timidger: Ok :(

src/compositor/seat.rs
Timidger: https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.as_slices

Timidger: Ah but you might need to combine the slices....

eh just leave this for now. It's fine.

Issue #545, opened 2018-05-27
Implement Pixbuf to surface for latest Awesome


Timidger:
The latest awesome has a new API call that is the same as load image except the Pixbuf is provided by the Lua code

PR #547, opened 2018-05-28
Lua: Implement awesome.pixbuf_to_surface()

psychon:
This is a new function that was added to awesome in commit ee944f4da83.
It gets a pointer to a GdkPixbuf instance as its argument and returns a
cairo image surface with the same content. It works similar to the
existing awesome.load_image(), but without loading a pixbuf from a file,
thus it is more general.

Fixes: https://github.com/way-cooler/way-cooler/issues/545

### DIFF
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -5,7 +5,7 @@ use super::{signal, XCB_CONNECTION_HANDLE};
 use awesome::lua::{load_config, rust_interop, LUA};
 use cairo::{self, ImageSurface, ImageSurfaceData};
 use gdk_pixbuf::{Pixbuf, PixbufExt};
-use glib::translate::ToGlibPtr;
+use glib::translate::{ToGlibPtr, FromGlibPtrNone};
 use nix::{self, libc};
 use rlua::{self, AnyUserData, LightUserData, Lua, MetaMethod, Table, ToLua, UserData,
            UserDataMethods, Value};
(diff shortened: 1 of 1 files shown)


psychon:
I forgot: I tested this with the following Lua code. Both resulting PNG files are equally broken, so this new code seems to work fine. I'll debug `load_surface_from_pixbuf` next...
```lua
local lgi = require("lgi")
local cairo = lgi.cairo
local GdkPixbuf = lgi.GdkPixbuf

local img = "/home/psychon/Wallpaper/flowers-on-black.png"

-- Old version, load_image
local s = cairo.Surface(awesome.load_image(img), true)
s:write_to_png("/tmp/method_old.png")

-- New version, pixbuf_to_surface
local pb = GdkPixbuf.Pixbuf.new_from_file(img)
local s = cairo.Surface(awesome.pixbuf_to_surface(pb._native), true)
s:write_to_png("/tmp/method_new.png")

print("done")
```
(Oh and: I only tested the old method in awesome, my currently running awesome instance is too old for `pixbuf_to_surface`)

acrisci:
:+1: I can actually use the default config with this.

Timidger:
Getting the raw pointer is an issue in aware of for using the Cairo library.

The way you solved it is my preferred way. Though I have not reached out to the Cairo-rs people on changing it.

src/awesome/awesome.rs
```diff
@@ -248,6 +249,19 @@ fn load_image<'lua>(lua: &'lua Lua, file_path: String) -> rlua::Result<Value<'lu
     rlua::LightUserData(surface_ptr as _).to_lua(lua)
 }
 
+/// Convert a pixbuf to a cairo image surface.
+/// Returns either a cairo surface as light user data, nil and an error message
+fn pixbuf_to_surface<'lua>(lua: &'lua Lua, pixbuf: LightUserData) -> rlua::Result<Value<'lua>> {
```
Timidger: The raw pointer being passed in here from Lua with no checks possible for validity makes me sad. But there's nothing we can do I suppose.

Perhaps once Way Cooler is stable we can look into making that safer.

psychon:
> The way you solved it is my preferred way. Though I have not reached out to the Cairo-rs people on changing it.

That's just copy&paste from `load_image` (/me hides).
I saw that a `to_glib_none` exists and tried to use it, but  hit an `unimplemented!()` :-(

PR #551, opened 2018-06-02
use current_time() in surface.send_frame_done

acrisci:
See this [wlroots-rs](https://github.com/swaywm/wlroots-rs/issues/180) issue.


Timidger:
Thanks!

PR #553, opened 2018-06-09
Implemented dehandle! everywhere it makes sense

Timidger:
Look how much sexier that is...

\*cough\* \*cough*

Excuse me.

Implement `dehandle!` everywhere we used to use `with_handles!`. Should not be merged until wlroots-rs is and we fix the removal of the input manager callbacks.


acrisci:
I'm ok with this :+1: .

acrisci:
This broke the compositor xdgv6 `map_request` function. It says the compositor handle is already borrowed.

<blockquote>
thread 'main' panicked at 'Could not upgrade unhandle_name and set the result to handle_name: AlreadyBorrowed', libcore/result.rs:945:5
</blockquote>

Timidger:
yikes, my bad.

Also looks like that error message is basically garbage, which is fun..

Will fix after work.

Timidger:
This has been fixed, sorry about that.

PR #557, opened 2018-06-11
basic move resize

acrisci:
A move-resize operation must be handled asynchronously because it is done with a request to the shell surface and we need to wait for it to update its buffer before we can update the view size. Sometimes, updating the size of the view must be done at the same time as updating one of the view x/y coordinates so that must be kept in sync.

That's what the view `move_resize` function does. It takes an `Area` of the new position of the view, makes a `set_size` request on the toplevel, and then does both the x/y position change and the width/height update at the same time. Without that, you would update x/y first and width/height at some later time causing a flash of partially rendered content.

The view now has a `PendingMoveResize` struct which is set in the `view.move_resize` function and read in the surface `commit` handler. On commit, if the configure serial matches the `PendingMoveResize` serial, it does the update.

The seat got a new mode called `Resizing`. Here we store the offset of the cursor on the surface, and the original position of the view. We calculate how much the cursor has moved from the original spot and adjust the final view position based on that.

TODO:

* [x] async move/resize
* [x] seat resize state
* [ ] bug: jitter in the bottom corner on draging the top-left corner and moving rapidly (exists in rootston as well) that sort of looks like a rounding issue.
* [x] put the `WLR_EDGE` enum somewhere (maybe in wlroots-rs).
* [ ] can this be simpler? (maybe not).

### DIFF
--- a/src/compositor/seat.rs
+++ b/src/compositor/seat.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use wlroots;
 use wlroots::events::seat_events::SetCursorEvent;
 use wlroots::pointer_events::ButtonEvent;
-use wlroots::utils::current_time;
+use wlroots::utils::{current_time, Edges};
 use wlroots::{Area, CompositorHandle, Cursor, CursorHandle, DragIconHandle, Origin, SeatHandle,
               SeatHandler, Size, SurfaceHandle, SurfaceHandler, XCursorManager};
 
@@ -22,7 +22,7 @@ pub enum Action {
         start: Origin,
         offset: Origin,
         original_size: Size,
-        edges: u32
+        edges: Edges
     }
 }
 
@@ -115,7 +115,7 @@ impl Seat {
                         cursor: &mut CursorHandle,
                         view: Rc<View>,
                         views: &mut [Rc<View>],
-                        edges: u32) {
+                        edges: Edges) {
         self.focus_view(view.clone(), views);
         with_handles!([(cursor: {cursor})] => {
             let Origin { x: view_x, y: view_y } = view.origin.get();
@@ -185,23 +185,17 @@ impl Seat {
                     let Size { mut width,
                                mut height } = original_size;
 
-                    let EDGE_NONE = 0;
-                    let EDGE_TOP = 1;
-                    let EDGE_BOTTOM = 2;
-                    let EDGE_LEFT = 4;
-                    let EDGE_RIGHT = 8;
-
-                    if edges & EDGE_BOTTOM != 0 {
+                    if edges & Edges::WLR_EDGE_BOTTOM != Edges::empty() {
                         height += dy;
-                    } else if edges & EDGE_TOP != 0 {
+                    } else if edges & Edges::WLR_EDGE_TOP != Edges::empty() {
                         view_y += dy;
                         height -= dy;
                     }
 
-                    if edges & EDGE_LEFT != 0 {
+                    if edges & Edges::WLR_EDGE_LEFT != Edges::empty() {
                         view_x += dx;
                         width -= dx;
-                    } else if edges & EDGE_RIGHT != 0 {
+                    } else if edges & Edges::WLR_EDGE_RIGHT != Edges::empty() {
                         width += dx;
                     }
 
--- a/src/compositor/seat.rs
+++ b/src/compositor/seat.rs
@@ -6,8 +6,8 @@ use wlroots;
 use wlroots::events::seat_events::SetCursorEvent;
 use wlroots::pointer_events::ButtonEvent;
 use wlroots::utils::current_time;
-use wlroots::{CompositorHandle, Cursor, DragIconHandle, Origin, SeatHandle, SeatHandler,
-              SurfaceHandle, SurfaceHandler, XCursorManager};
+use wlroots::{Area, CompositorHandle, Cursor, CursorHandle, DragIconHandle, Origin, SeatHandle,
+              SeatHandler, Size, SurfaceHandle, SurfaceHandler, XCursorManager};
 
 #[derive(Debug, Default)]
 pub struct SeatManager;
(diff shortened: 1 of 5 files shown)


src/compositor/seat.rs
```diff
@@ -144,6 +170,41 @@ impl Seat {
...
+                    let Origin { x: mut offs_x,
+                                 y: mut offs_y } = offset;
+                    let Origin { x: mut view_x,
+                                 y: mut view_y } = start;
+                    let (mut dx, mut dy) =
+                        (cursor_lx as i32 - offs_x - view_x, cursor_ly as i32 - offs_y - view_y);
+                    let Size { mut width,
+                               mut height } = original_size;
+
+                    if edges & Edges::WLR_EDGE_BOTTOM != Edges::empty() {
```
Timidger: Just use `edges.contains(...)` here for these comparisons.

src/compositor/shells/xdg_v6.rs
```diff
@@ -17,6 +17,27 @@ impl XdgV6 {
 }
 
 impl XdgV6ShellHandler for XdgV6 {
+    fn resize_request(&mut self,
+                      compositor: CompositorHandle,
+                      _: SurfaceHandle,
+                      shell_surface: XdgV6ShellSurfaceHandle,
+                      event: &ResizeEvent) {
+        println!("got a resize event: {:?}", event);
```
Timidger: Remove this

src/compositor/shells/xdg_v6.rs
```diff
@@ -17,6 +17,27 @@ impl XdgV6 {
...
+        with_handles!([(compositor: {compositor})] => {
+            let server: &mut Server = compositor.into();
+            let Server { ref mut seat,
+                         ref mut views,
+                         ref mut cursor,
+                         .. } = *server;
+            let resizing_shell = shell_surface.into();
+            let idx = views.iter().position(|view| view.shell == resizing_shell);
+            if let Some(idx) = idx {
+                let view = views.get(idx).unwrap().clone();
```
Timidger: Use `views[idx].clone()` here since you know the `idx` is within bounds.

In fact since you're not using the index for anything _really_ you can do this instead and you have guaranteed elided bounds checks if you do this instead:

```rust
for view in &views {
    if view.shell == resizing_shell {
        seat.begin_resize(cursor, view.clone(), views, event.edges());
        break;
    }
}
```

Timidger: Or actually event better, use [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find). Those functions usually generate better code than the iterative version:

```rust
if let Some(view) = views.iter().find(|view| view.shell == resizing_shell).cloned() {
    seat.begin_resize(cursor, view.clone(), views, event.edges())
}
```

src/compositor/shells/xdg_v6.rs
```diff
@@ -43,6 +64,49 @@ impl XdgV6ShellHandler for XdgV6 {
...
+        };
+
+        let surface = shell_surface.into();
+        with_handles!([(compositor: {compositor})] => {
+            let server: &mut Server = compositor.into();
+            let Server { ref mut seat,
+                         ref mut views,
+                         ref mut cursor,
+                         .. } = *server;
+            let idx = views.iter().position(|view| view.shell == surface);
```
Timidger: Do the same thing as you did above here using `find` for no bounds checking.

src/compositor/shells/xdg_v6.rs
```diff
@@ -43,6 +64,49 @@ impl XdgV6ShellHandler for XdgV6 {
...
+            if let Some(idx) = idx {
+                let view = views.get(idx).unwrap().clone();
+                if let Some(move_resize) = view.pending_move_resize.get() {
+                    if move_resize.serial > 0 && move_resize.serial >= configure_serial {
```
Timidger: Can/should a serial ever by < 0? Should we switch this to e.g. a `u64` or something?

src/compositor/view.rs
```diff
@@ -40,6 +50,55 @@ impl View {
         }
     }
 
+    pub fn get_size(&self) -> Size {
+        match self.shell {
+            Shell::XdgV6(ref xdg_surface) => {
+                with_handles!([(xdg_surface: {xdg_surface})] => {
+                    let Area { origin: _, size } = xdg_surface.geometry();
```
Timidger: You can just do `Area { _, size }` here

src/compositor/shells/xdg_v6.rs
acrisci: configure_serial is uint32 in wlroots.

That check isn't required though since we already check for 0 serial earlier. That's the case when no configure was sent because the properties are the same as we are trying to set.

src/compositor/view.rs
acrisci: That seems to be a compiler error.

src/compositor/seat.rs
```diff
@@ -104,6 +111,25 @@ impl Seat {
...
+        self.focus_view(view.clone(), views);
+        with_handles!([(cursor: {cursor})] => {
+            let Origin { x: view_x, y: view_y } = view.origin.get();
+            let (lx, ly) = cursor.coords();
+            let (view_sx, view_sy) = (lx - view_x as f64, ly - view_y as f64);
+            let offset = Origin::new(view_sx as _, view_sy as _);
+            self.action = Some(Action::Resizing {
+                start: Origin { x: view_x, y: view_y },
+                offset,
+                original_size: view.get_size(), edges
```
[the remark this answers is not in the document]
acrisci: rustfmt didn't catch this one.

acrisci:
I dunno if I can fix that jitter. It doesn't seem to be a big deal anyway. Maybe leave that task up to someone who it bothers in the future.

Timidger:
Thanks!

PR #559, opened 2018-06-15
client.get()

acrisci:
Here's one way to do it.

Set the `CompositorHandle` in a lua `UserData` and set it to a named registry value. Now we can get the `Server` anytime we want it in the awesome code.

Now the `View` can have some kind of `to_lua()` function and we can just put them all in a table and then `client.get()` will be implemented.

What do you think?

### DIFF
--- a/src/awesome/client.rs
+++ b/src/awesome/client.rs
@@ -4,15 +4,15 @@ use super::object::{Object, Objectable};
 use rlua::{self, Lua, Table, ToLua, UserData, Value, AnyUserData, UserDataMethods, MetaMethod};
 use std::default::Default;
 use std::fmt::{self, Display, Formatter};
+use std::rc::Rc;
 use compositor::{Server, View};
+use wlroots::{compositor_handle, Origin, Size, Area};
 
 pub const CLIENTS_HANDLE: &'static str = "__clients";
 
 #[derive(Clone, Debug)]
 pub struct ClientState {
-    // TODO Fill in
-    pub lua_id: u32,
-    title: String,
+    pub lua_id: u32
 }
 
 #[derive(Clone, Debug)]
@@ -20,7 +20,7 @@ pub struct Client<'lua>(Object<'lua>);
 
 impl Default for ClientState {
     fn default() -> Self {
-        ClientState { lua_id: 0, title: "".into() }
+        ClientState { lua_id: 0 }
     }
 }
 
@@ -33,7 +33,6 @@ impl <'lua> Client<'lua> {
     pub fn init_client(&mut self, view: &View) -> rlua::Result<()> {
         let mut state = self.get_object_mut()?;
         state.lua_id = view.lua_id;
-        state.title = view.title();
         Ok(())
     }
 }
@@ -50,14 +49,49 @@ impl<'lua> ToLua<'lua> for Client<'lua> {
     }
 }
 
+impl ClientState {
+    fn view(&self) -> Rc<View> {
+        with_handles!([(compositor: {compositor_handle().unwrap()})] => {
+            let server: &mut Server = compositor.into();
+            server.views.iter().find(|view| view.lua_id == self.lua_id).unwrap().clone()
+        }).unwrap()
+    }
+}
+
 impl UserData for ClientState {
     fn add_methods(methods: &mut UserDataMethods<Self>) {
+        methods.add_method("geometry", |lua, state, (geometry,): (Value,)| {
+            match geometry {
+                Value::Table(table) => {
+                    let area = Area {
+                        origin: Origin { x: table.get("x")?, y: table.get("y")? },
+                        size: Size { width: table.get("width")?, height: table.get("height")? }
+                    };
+
+                    state.view().move_resize(area);
+
+                    Ok(Value::Nil.to_lua(lua))
+                },
+                _ => {
+                    let view = state.view();
+                    let table = lua.create_table()?;
+                    let geometry = view.geometry();
+                    let Origin { x, y } = view.origin.get();
+                    table.set("x", geometry.origin.x + x)?;
+                    table.set("y", geometry.origin.y + y)?;
+                    table.set("width", geometry.size.width)?;
+                    table.set("height", geometry.size.height)?;
+                    Ok(table.to_lua(lua))
+                }
+            }
+        });
+
         methods.add_meta_function(MetaMethod::Index, |lua, (class, index): (ClientState, Value)| {
             match index {
                 Value::String(value) => {
                     match value.to_str() {
                         Ok("title") => {
-                            Ok(class.title.to_lua(lua))
+                            Ok(class.view().title().to_lua(lua))
                         },
                         _ => Ok(Value::Nil.to_lua(lua))
                     }
--- a/src/awesome/client.rs
+++ b/src/awesome/client.rs
@@ -10,19 +10,13 @@ use wlroots::{compositor_handle, Origin, Size, Area};
 
 pub const CLIENTS_HANDLE: &'static str = "__clients";
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct ClientState {
-    pub lua_id: u32
+    pub view: Arc<View>
 }
 
 #[derive(Clone, Debug)]
 pub struct Client<'lua>(Object<'lua>);
-
-impl Default for ClientState {
-    fn default() -> Self {
-        ClientState { lua_id: 0 }
-    }
-}
 
 impl <'lua> Client<'lua> {
     pub fn new(lua: &Lua) -> rlua::Result<Object> {
@@ -30,9 +24,9 @@ impl <'lua> Client<'lua> {
         Ok(Client::allocate(lua, class)?.build())
     }
 
-    pub fn init_client(&mut self, view: &View) -> rlua::Result<()> {
+    pub fn init_client(&mut self, view: Arc<View>) -> rlua::Result<()> {
         let mut state = self.get_object_mut()?;
-        state.lua_id = view.lua_id;
+        state.view = view;
         Ok(())
     }
 }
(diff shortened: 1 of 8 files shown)


acrisci:
Here's a config I'm using to test this function with:

```lua
local awful = require("awful")
local gears = require("gears")

function test_func()
    print("client list begin")
    for _, c in ipairs(client.get()) do
        print(c.title)
    end
    print("client list end")
end

globalkeys = gears.table.join(
    awful.key({}, "s", test_func)
)

root.keys(globalkeys)
```

acrisci:
Oh ok I was not aware of that. That seems to work just as well.

psychon:
Sorry, for making everyone's life harder, but I'd like to point out that the following should work:
```lua
local t = {}
for _, c in ipairs(client.get()) do
  t[c] = true
end
for _, c in ipairs(client.get()) do
  assert(t[c])
end
```
Basically, two calls to `client.get()` have to return the same client object (not just equal according to the `__equal` metamethod, it really has to be the same). Thus, we cannot create client objects on-demand, but they somehow have to exist for as long as the corresponding View exists.

Timidger:
Better that's brought up now than in a bug report. Thanks for the heads up @psychon

acrisci:
ba26b46 makes my test config work and can be extended to the other properties pretty easily, but the approach is a little strange.

We give every view a `lua_id` and have separate state for the lua client and the compositor view.

When the compositor wants to update the lua state, it calls notification functions that update the lua state.

When lua wants to update the compositor state, it can get the compositor directly and call view methods.

The lookup mechanism is a bit of a compromise solution. I wish we could just store the view on the `ClientState` and the `ClientState` on the view, but that's not really easy to do because of the threading and lifetime requirements.

Timidger:
You can store it in the view if you do `unsafe impl Send for ClientState {}`. This is safe to do in this context because there's no other thread. AFAIK for our purposes there's no UB possible by doing that, but if you are considering a case I'm not I would like to hear it.

Not a huge fan of that weird `lua_id` thing because I'm not sure what it adds over the `unsafe impl Send for ClientState {}`

acrisci:
The problem is that the `ClientState` is owned by the `Client` which requires a lifetime (or a reference to the state with a lifetime) and if views require a lifetime, then that cascades throughout the server structs. I don't think I know enough about lifetimes to tell whether that is a workable solution. Or do you think there is a better way?

acrisci:
@psychon is setting the geometry synchronous? And is there a specific order the clients are supposed to be in in `client.get()`?

psychon:
> is setting the geometry synchronous?

Awesome immediately sends the resize request to the X11 server. But X11 itself is not synchronous here, so... what exactly do you mean? I would tend to say "yes", but I am not quite sure against what this should be synchronised.

> And is there a specific order the clients are supposed to be in in client.get()?

Well, `client.get()` actually has two optional arguments. The first one is a screen, in which case only clients from that screen are returned. The other one defaults to `false`, which means to return clients in creation order (oldest one first, I think). With `client.get(nil, true)`, clients are returned in stacking order (top to bottom, according to the docs).

Timidger:
> if views require a lifetime

But views don't require a lifetime, they have handles which can lives as long as they want.

`Client` indirectly owns `ClientState` through the `Object` using the user data values Lua provides. This is so the state stricts can store Lua data and not cause an unbreakable cycle for the gc.

Handles are only equal if they refer to the same underlying data (eg when you compare handled with  == it's doing an identity comparison for the inner pointer). You don't have to store Clients in Views, you can just have the views be in clients and gracefully handle them expiring (instead of `unwrapping` like we've been doing everywhere else.

acrisci:
> But views don't require a lifetime, they have handles which can lives as long as they want.

Views don't have handles. They are reference counted and stored in a list on the Server. `Rc` can't be `Send` so storing them in structs that live in awesome is not possible. I think the `lua_id` is kind of like a crappy way of doing handles. Do you want to properly do handles for the views as an alternative?

Timidger:
Ahh, I see. Sorry about that.

If you change them to Arcs it should work, but you'll have the atomic overhead cost unfortunately. I suggest you do that though because that's much easier to change in the future than this current system.

If the arc is too much overhead we can start using "proper" handles

src/compositor/shells/mod.rs
```diff
@@ -10,6 +10,13 @@ pub enum Shell {
                                     * TODO Xdg */
 }
 
+impl Default for Shell {
+    fn default() -> Self {
+        // The default shell is always invalid
+        Shell::XdgV6(XdgV6ShellSurfaceHandle::new())
```
acrisci: The only awkward thing about this is I needed a `Default` for the view to avoid putting an ugly `Option` in the `ClientState`.

Timidger: Ah that's a shame. Not a big deal though, as long as we always set it in the client state

acrisci:
I think that's all the features I want to do here, even though it is incomplete. I'm going to start chipping away on `awful.layout.set(awful.layout.suit.tile)` and just fix all the lua errors until it tiles.

src/awesome/client.rs
```diff
@@ -42,9 +43,60 @@ impl<'lua> ToLua<'lua> for Client<'lua> {
...
+                    }
+                },
+                _ => Ok(Value::Nil.to_lua(lua))
+            }
+        });
+    }
+}
+
+pub fn init<'lua>(lua: &'lua Lua, server: &Server) -> rlua::Result<Class<'lua>> {
+    let clients: &mut Vec<Client> = &mut Vec::default();
```
Timidger: Uh, is there a reason you did this instead of 

```rust
let mut clients: Vec<Client> = Vec::new()
```

src/awesome/client.rs
```diff
@@ -42,9 +43,60 @@ impl<'lua> ToLua<'lua> for Client<'lua> {
...
+
+    for view in &server.views {
+        let mut client = Client::cast(Client::new(lua)?)?;
+        client.init_client(view.clone())?;
+        clients.push(client);
+    }
+
+    lua.set_named_registry_value(CLIENTS_HANDLE, clients.clone().to_lua(lua)?)?;
```
Timidger: Why are you cloning the vector here? It looks like you don't use it anywhere else?

src/awesome/client.rs
```diff
@@ -55,11 +107,51 @@ fn method_setup<'lua>(lua: &'lua Lua,
...
-fn dummy_table<'lua>(lua: &'lua Lua, _: rlua::Value) -> rlua::Result<Table<'lua>> {
-    Ok(lua.create_table()?)
+fn get_clients<'lua>(lua: &'lua Lua, _: rlua::Value) -> rlua::Result<Table<'lua>> {
+    let clients: Vec<Client> = lua.named_registry_value::<Vec<AnyUserData>>(CLIENTS_HANDLE)?
+                                      .into_iter()
+                                      .map(|obj| Client::cast(obj.into()).unwrap())
+                                      .collect();
+    let table = lua.create_table()?;
+
+    for (i, client) in clients.iter().enumerate() {
```
Timidger: rlua implement `ToLua` for `Vec<T>` [see here](https://docs.rs/rlua/0.13.0/rlua/prelude/trait.ToLua.html#impl-ToLua%3C'lua%3E-15). It will automatically convert it into what your  doing here manually.

So you can just do 

```rust
let client: Vec<Client> = ...;
client.to_lua(lua)
```

src/compositor/view.rs
```diff
@@ -11,18 +11,31 @@ pub struct PendingMoveResize {
     pub area: Area
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Debug, Default)]
 pub struct View {
     pub shell: Shell,
-    pub origin: Cell<Origin>,
-    pub pending_move_resize: Cell<Option<PendingMoveResize>>
+    pub origin: Mutex<Origin>,
```
Timidger: :( Cell is such a nicer interface, is a shame.

src/awesome/client.rs
```diff
@@ -42,9 +43,60 @@ impl<'lua> ToLua<'lua> for Client<'lua> {
     }
 }
 
-impl UserData for ClientState {}
+impl UserData for ClientState {
+    fn add_methods(methods: &mut UserDataMethods<Self>) {
+        methods.add_method("geometry", |lua, state, (geometry,): (Value,)| {
```
Timidger: If you want to add methods on the object, I suggest you use the [same API we use else where](https://github.com/way-cooler/way-cooler/blob/master/src/awesome/drawable.rs#L120)

src/awesome/client.rs
```diff
@@ -42,9 +43,60 @@ impl<'lua> ToLua<'lua> for Client<'lua> {
...
+                    table.set("x", geometry.origin.x + x)?;
+                    table.set("y", geometry.origin.y + y)?;
+                    table.set("width", geometry.size.width)?;
+                    table.set("height", geometry.size.height)?;
+                    Ok(table.to_lua(lua))
+                }
+            }
+        });
+
+        methods.add_meta_function(MetaMethod::Index, |lua, (class, index): (ClientState, Value)| {
```
Timidger: For this method you can use the same API as the others as well and just [use the special `"__index"` name](https://github.com/way-cooler/way-cooler/blob/master/src/awesome/screen.rs#L144).

src/awesome/client.rs
acrisci: That sets it on the global. Seems like the real api is here:

https://github.com/way-cooler/way-cooler/blob/833db6ad5cbd040f3b124818c65b4fab24d0d729/src/awesome/drawable.rs#L35-L44

Do you really think that's better?

src/awesome/client.rs
acrisci: This api also only works only for the global.

src/awesome/client.rs
Timidger: Well that code adds it to the meta table whereas this adds it to the user data AFAIK. So it's a little different.

For our purposes they _should_ be be equivalent AFAIK? We should probably standardize on one way though. CC @psychon if it makes a difference?

@acrisci Keep it the way it is for now if it works.

psychon: Well... dunno. I copied the approach for the code above from other places of way-cooler, but I vaguely remember that there were some inconsistencies.

Also, this `add_methods` thing comes from rlua, not lua. It is implemented via an `__index` metamethod as well: https://github.com/chucklefish/rlua/blob/71f3dd50a1f396ea2ff36462e3deb6a48a132c43/src/lua.rs#L814

So, I guess way-cooler grew a reimplementation of some rlua feature somewhere along the way...

Timidger: acrisci's implementation puts the methods on the user data and my object implementation puts it on the table associated with the user data IIRC. The default index method ensures that our functions are called if defined on the table.

Timidger:
ping?

acrisci:
@Timidger is this the way we'll implement this with the multiprocess design or will clients go over the wire?

Timidger:
@acrisci Hmm, yes at a second glance most of this code will probably be thrown out since the changes are tied to the current design.

I'll go ahead and close then.

(FYI I'm working on the split up now)

PR #560, opened 2018-06-15
Restart now shuts down the compositor

Timidger:
~Currently Way Cooler segfaults when Lua restarts because GLib has pointers into the old instance. The solution is to restart that GLib main loop as well.~

~In order to do that, we must make Wayland own the glib loop (e.g. it calls out to it). This will require doing some weird things with unsafe stuff.~

~The biggest problem with this patch right now is that we pass `glib::MainLoop` to C code when you need to pass the pointer. Unfortunately that library doesn't actually provide a way to get to the pointer, so even though it "works" now it's because of UB based on the ordering of Rust structs which can change. A patch should be sent to them.~

~The other big problem is we don't know if it works yet because we don't call `refresh_awesome` any more. The C code will need to be modified for that.~

~This patch has been fixed to work. Unfortunately it relies on UB.~

# TODO
- [ ] ~Clean up code~
- [ ] ~Remove the UB by patching glib-rs~

Gave up and now restarting shutsdown the compositor.

In the future there should be another attempt to fix this.

### DIFF
--- a/config/rc.lua
+++ b/config/rc.lua
@@ -19,6 +19,8 @@ local hotkeys_popup = require("awful.hotkeys_popup").widget
 -- when client with a matching name is opened:
 require("awful.hotkeys_popup.keys")
 
+gears.timer.start_new(1, function() print("I live") return true end)
+
 -- {{{ Error handling
 -- @DOC_ERROR_HANDLING@
 -- Check if awesome encountered an error during startup and fell back to
@@ -61,7 +63,7 @@ editor_cmd = terminal .. " -e " .. editor
 -- If you do not like this or do not have such a key,
 -- I suggest you to remap Mod4 to another key using xmodmap or other tools.
 -- However, you can use another modifier like Mod1, but it may interact with others.
-modkey = "Mod4"
+modkey = "Ctrl"
 
 -- @DOC_LAYOUT@
 -- Table of layouts to cover with awful.layout.inc, order matters.
--- a/src/awesome/awesome.rs
+++ b/src/awesome/awesome.rs
@@ -2,7 +2,6 @@
 
 use super::xproperty::{XProperty, XPropertyType, PROPERTIES};
 use super::{signal, XCB_CONNECTION_HANDLE};
-use awesome::lua::NEXT_LUA;
 use cairo::{self, ImageSurface, ImageSurfaceData};
 use gdk_pixbuf::{Pixbuf, PixbufExt};
 use glib::translate::{ToGlibPtr, FromGlibPtrNone};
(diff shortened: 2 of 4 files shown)


psychon:
> we don't know if it works yet

It doesn't, sorry.

The `GMainLoop` does not own the event sources, but the `GMainContext` does. The `None` argument that we are using to construct a `GMainLoop` is the `GMainContext` that it should use. A `NULL` pointer here means that it should use the default main context (`g_main_context_default()`).

As far as I see from a quick look at https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html, there is no way to replace the default main context (the thread default context is something else).

And before some asks why this split exists: So you can write blocking methods that still run the main loop. E.g. a function that opens the "Select a file to open"-dialog would create a new main loop and run it until the user selected a file. Then it could return that file to the caller. Since the new main loop also runs with the default context, "everything" continues to work.

Timidger:
:(

OK, I'll work on this this weekend. Been really busy at work this week so I haven't had a chance to give this bug the time it deserves.

Timidger:
Ok this latest change should be the fix that actually fixes the segfault. I have tested it and it seems like it restarts correctly.

Way to test that:

1) Run in embedded context (e.g. under X11 or Wayland)
2) Change the meta key to something that won't be captured by the host environment (e.g. I changed `Mod4` to `Ctrl`
3) Resize the window so the top bar is slightly hidden
4) Press the mod key, watch the cascade of errors from all our unimplemented functions
5) Bask in the fact that it doesn't crash and the top bar re-renders itself to match its new dimensions.

Unfortunately this code relies on UB because the `glib` crate doesn't provide a way for us to get to the pointer. I'm going to reach out to them and see if they can add that.

Timidger:
Actually no this is still broken...

```
12:57 < psychon> add gears.timer.start_new(5, function() print("I live") return true) to rc.lua and notice that the timer does not run
12:57 < psychon> the whole reason why Lua can add "stuff" to the main loop (well, main context) is that everything uses GLib's default main context
12:57 < timidger> ughhhj
12:58 < psychon> by running a GMainLoop with a different GMainContext, Lua no longer has access to it
12:58 < psychon> (note that I did not actually try this, but I am fairly sure this happens)
12:58 < timidger> is there no way to "replace" the main context?
12:58 < timidger> Oh let me try it then
12:59 < psychon> that's what I meant when I wrote that I see no way to replace the default main context (with a link to the API docs)
13:00 < timidger> oh
13:00 < psychon> https://sources.debian.org/src/glib2.0/2.56.1-2/glib/gmain.c/?hl=684#L684
13:01 < psychon> seems to be a static variable in gmain.c that is only accessed by g_main_context_default() and nothing else
13:01 < psychon> I guess the best way-cooler can do is to make awesome.restart() error out
13:01 < psychon> sorry
13:01 < timidger> well that's dissapointing
```

PR #562, opened 2018-06-18
implement mouse.screen

acrisci:
https://awesomewm.org/apidoc/libraries/mouse.html#screen

```
screen
    The screen under the cursor
    Type: screen
```

### DIFF
--- a/src/awesome/mouse.rs
+++ b/src/awesome/mouse.rs
@@ -1,10 +1,13 @@
 //! TODO Fill in
 
-use rlua::{self, AnyUserData, Lua, MetaMethod, Table, UserData, UserDataMethods, Value};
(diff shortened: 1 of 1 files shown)


src/awesome/mouse.rs
```diff
@@ -106,29 +109,35 @@ fn index<'lua>(_: &'lua Lua,
...
+            if let Some(output) = output {
+                for screen in screens {
+                    let state = screen.state()?;
+                    if state.outputs.contains(&output) {
+                        return screen.to_lua(lua);
+                    }
+                }
+            }
+
+            return Value::Nil.to_lua(lua);
```
Timidger: You can just put `Ok(Value::Nil)`

Timidger: Though FYI Awesome doesn't handle that well. So it's better to just lie and say the first output.

PR #564, opened 2018-06-21
Fix XDG_DATA_DIRS and XDG_CONFIG_DIRS handling

yzhs:
When using multiple directories in the `XDG_DATA_DIRS` or `XDG_CONFIG_DIRS` environment variables, as specified in the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), these variables are currently interpreted as a single directory. This prevents Lua from finding the AwesomeWM modules.

This pull request fixes this problem by splitting the directories and handling each directory separately. In addition, it changes the default for `XDG_DATA_DIRS` to `/usr/local/share:/usr/share`, as specified.


psychon:
More or less unrelated to this PR, but there seems to be rust bindings to libxdg-basedir: https://crates.io/crates/xdg-basedir

Timidger:
Thanks!

PR #566, opened 2018-07-04
Fix immediate issues from base rc.lua

Akulen:
This allows the base rc.lua to be executed without errors, and for the bindings defined in it to be used.
Adds two dummies for `awesome.spawn` and `drawin.buttons`

### DIFF
--- a/src/awesome/key.rs
+++ b/src/awesome/key.rs
@@ -138,7 +138,9 @@ fn set_key<'lua>(_: &'lua Lua,
     if key_name.starts_with('#') && key_name.len() >= 2 {
         let number = key_name[1..].parse::<xkb::Keycode>()
             .map_err(|err| rlua::Error::RuntimeError(format!("Parse error: {:?}", err)))?;
-        key.set_keycode(number-8 as xkb::Keycode)?;
+        // the - 8 is because of xcb conventions, where "#10" is the keysim for 1,
+        // and the keycode of 1 is 0x02 (obviously)
+        key.set_keycode(number - 8 as xkb::Keycode)?;
     } else {
         let keysym = xkb::keysym_from_name(key_name.as_str(), 0);
         key.set_keysym(keysym)?;
--- a/src/awesome/key.rs
+++ b/src/awesome/key.rs
@@ -134,13 +134,13 @@ fn get_key<'lua>(lua: &'lua Lua, obj: AnyUserData<'lua>) -> rlua::Result<Value<'
 fn set_key<'lua>(_: &'lua Lua,
                  (obj, key_name): (AnyUserData<'lua>, String))
                  -> rlua::Result<Value<'lua>> {
-    let keysym = xkb::keysym_from_name(key_name.as_str(), 0);
     let mut key = Key::cast(obj.clone().into())?;
     if key_name.starts_with('#') && key_name.len() >= 2 {
         let number = key_name[1..].parse::<xkb::Keycode>()
             .map_err(|err| rlua::Error::RuntimeError(format!("Parse error: {:?}", err)))?;
-        key.set_keycode(number as xkb::Keycode)?;
+        key.set_keycode(number-8 as xkb::Keycode)?;
     } else {
+        let keysym = xkb::keysym_from_name(key_name.as_str(), 0);
         key.set_keysym(keysym)?;
     }
     Ok(rlua::Value::Nil)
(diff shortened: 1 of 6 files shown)


src/awesome/keygrabber.rs
```diff
@@ -44,8 +44,16 @@ pub fn keygrabber_handle(mods: Vec<Key>, sym: Key, state: wlr_key_state) -> rlua
              })
 }
 
+/// Check is the Lua callback function is set
+pub fn is_keygrabber_set(lua: &Lua) -> bool {
+    match lua.named_registry_value::<Function>(KEYGRABBER_CALLBACK) {
```
Timidger: Just use `is_ok()` here instead of the match

src/awesome/lua/utils.rs
```diff
@@ -29,6 +29,24 @@ pub fn mods_to_lua<'lua>(lua: &'lua Lua, mods: &[Key]) -> rlua::Result<Table<'lu
     lua.create_table_from(mods_list.into_iter().enumerate())
 }
 
+/// Convert a single number to a modifier list.
+pub fn num_to_mods(modifiers: KeyboardModifier) -> Vec<Key> {
+    let mut res = vec![];
+    let mod_types = [(KeyboardModifier::WLR_MODIFIER_SHIFT, KEY_Shift_L),
```
Timidger: Is it possible to make this a global constant? I have a feeling we will use this again somewhere.

src/compositor/input/keyboard.rs
```diff
@@ -93,24 +93,45 @@ fn emit_awesome_keybindings(lua: &Lua,
...
+    // TODO check behavior when event.pressed_keys() isn't a singleton
+    if awesome::keygrabber::is_keygrabber_set(&*lua) {
+        for event_keysym in event.pressed_keys() {
+            let nres = awesome::keygrabber::call_keygrabber(&*lua, (
+                    awesome::lua::mods_to_lua(&*lua,
+                                              &awesome::lua::num_to_mods(event_modifiers))?,
+                    keysym_get_name(event_keysym),
+                    state_string.into()
+            ));
+            res = match nres {
```
Timidger: This will only report the last error, which is a bit of a shame. Can you at least log every time we set the `err` for the `res`?.

Also this line can be more clear like this:

```rust
if let Err(err) = nres {
    // log here
    res = err;
}
```

src/awesome/key.rs
```diff
@@ -134,13 +134,13 @@ fn get_key<'lua>(lua: &'lua Lua, obj: AnyUserData<'lua>) -> rlua::Result<Value<'
 fn set_key<'lua>(_: &'lua Lua,
                  (obj, key_name): (AnyUserData<'lua>, String))
                  -> rlua::Result<Value<'lua>> {
-    let keysym = xkb::keysym_from_name(key_name.as_str(), 0);
     let mut key = Key::cast(obj.clone().into())?;
     if key_name.starts_with('#') && key_name.len() >= 2 {
         let number = key_name[1..].parse::<xkb::Keycode>()
             .map_err(|err| rlua::Error::RuntimeError(format!("Parse error: {:?}", err)))?;
-        key.set_keycode(number as xkb::Keycode)?;
+        key.set_keycode(number-8 as xkb::Keycode)?;
```
Timidger: Space around the '-8'

Can you comment on why this magic number is used?

src/compositor/input/keyboard.rs
Akulen: Now that I think about it, wouldn't it be cleaner if the error was instantly returned when it happened ? Maybe it's not the default behavior we want though

src/awesome/key.rs
```diff
@@ -138,7 +138,10 @@ fn set_key<'lua>(_: &'lua Lua,
     if key_name.starts_with('#') && key_name.len() >= 2 {
         let number = key_name[1..].parse::<xkb::Keycode>()
             .map_err(|err| rlua::Error::RuntimeError(format!("Parse error: {:?}", err)))?;
-        key.set_keycode(number-8 as xkb::Keycode)?;
+        // the - 8 is because of xcb conventions, where "#10" is the keysim for 1,
+        // and the keycode of 1 is 0x02 (obviously)
+        info!("{} {:?}", number, number - 8 as xkb::Keycode);
```
Timidger: Please don't log every key press an it's - 8 equivilant

src/awesome/keygrabber.rs
Timidger: hehe, no I mean use [the result is_ok function](https://doc.rust-lang.org/std/result/enum.Result.html#method.is_ok). `is_keygrabber_set` is a fine name

Issue #335, opened 2017-07-15
man page


platipo:
I loved [online documentation pages](http://way-cooler.org/), but I still stumbled configuring way-cooler because it's missing man page.
Maybe it's just me but for example I didn't find a way to get screenshot working  even though I'm aware of the work done in #278. (Sorted out: I'm using `scrot` instead of ` way-cooler-grab` 🤦‍)
Anyway in my opinion man page lack pushes away beginners from this awesome WM.

Timidger:
Thanks for the insight! I've been meaning to write man pages, especially ones that can be generated from eg markdown and hosted on the site as well (so you can have man pages for offline documentation, and the same documentation on the website). 

Unfortunately I don't really have the time for that right now, especially as I don't know the man page format. I'm pretty busy right now trying to add am awesomeWM compatibility layer to Way Cooler. 

If you, or anyone else, is willing to write documentation though I will happily assist on any questions you have and any pull requests you open to include it. Feel free to ping me on the gitter chat if you need any help.

AdminXVII:
Do you find the man pages satisfactory? If not, what would you have more detailed?

Timidger:
Ah just didn't realize this wasn't closed automatically. 

Feel free to add a "fixes #<issue_number>" in your future PRs that actually fix said issue.

AdminXVII:
I did, I think you overlooked it? It is in the PR's description

PR #573, opened 2018-10-17
Add man pages

AdminXVII:
Add man and html targets to the makefile, responsible for generating two versions of the content of the manpages directory, one for man per se and one for the web.
The man pages are a basic introduction to the window manager.
They are fairly short, as only one flag is supported and the specifications are already done by Awesome.
They are translated in french and english.

Issue: https://github.com/way-cooler/way-cooler/issues/335

### DIFF
--- a/manpages/awesome.1.md
+++ b/manpages/awesome.1.md
@@ -4,7 +4,7 @@ awesome(1)
 NAME
 ----
 
-awesome - Lua controller for Way-Cooler
+awesome - a clone of AwesomeWM as a Wayland client.
 
 SYNOPSIS
 --------
@@ -14,7 +14,7 @@ SYNOPSIS
 DESCRIPTION
 -----------
 
-*way-cooler* is a tiling window manager for WayLand based on AwesomeWM.
+*way-cooler* is a tiling window manager for Wayland based on AwesomeWM.
 
 *awesome* controls *way-cooler*'s behavior by exposing Lua and DBUS APIs. At startup, it reads an RC file and configure *way-cooler* accordingly.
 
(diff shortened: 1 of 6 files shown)


AdminXVII:
On the same track, it could be interesting to make a script to convert the website's documentation into a Texinfo manual. I can do it if you see it fit.

Timidger:
Thanks for doing this work! I have a few nits about the actual content, but otherwise looks great!

> On the same track, it could be interesting to make a script to convert the website's documentation into a Texinfo manual. I can do it if you see it fit.

I'm not against that, though note that the current documentation on way-cooler.org is about the pre-awesome-clone API. Instead we will eventually have a clone of [the awesome docs](https://awesomewm.org/doc/).

manpages/awesome.1.md
```diff
@@ -0,0 +1,46 @@
+awesome(1)
+==========
+
+NAME
+----
+
+awesome - Lua controller for Way-Cooler
```
Timidger: This wording is a bit ambiguous. Here's a clearer definition about what `awesome` does.

```suggestion
awesome - a clone of AwesomeWM as a Wayland client.
```

manpages/awesome.1.md
```diff
@@ -0,0 +1,46 @@
...
+
+SYNOPSIS
+--------
+
+*awesome* [*--version*]
+
+DESCRIPTION
+-----------
+
+*way-cooler* is a tiling window manager for WayLand based on AwesomeWM.
```
Timidger: It's Wayland not WayLand

```suggestion
*way-cooler* is a tiling window manager for Wayland based on AwesomeWM.
```

AdminXVII:
Done
If you have other requests for change in the future, I will be happy to change both versions

Issue #574, opened 2018-10-18
Tests for awesome/root.rs don't use the latest API


AdminXVII:
Way-cooler version: 0.8.0
RC file: default, but not applicable
Cargo version: 1.29.0
Rust version: 1.29.2

Debug log:
```
[master][~/dev/way-cooler]$ cargo test
   Compiling awesome v0.8.0 (file:///home/adminxvii/dev/way-cooler/awesome)                             
   Compiling way-cooler v0.8.0 (file:///home/adminxvii/dev/way-cooler/way-cooler)
error[E0432]: unresolved import `super::super::tag`
   --> awesome/src/root.rs:139:9
    |
139 |     use super::super::tag;
    |         ^^^^^^^^^^^^^^^^^ no `tag` in the root

error[E0432]: unresolved import `super::super::key`
   --> awesome/src/root.rs:140:9
    |
140 |     use super::super::key;
    |         ^^^^^^^^^^^^^^^^^ no `key` in the root. Did you mean to use `Key`?

warning: unused imports: `KeyboardModifier`, `keysym_get_name`
 --> way-cooler/src/input/keyboard.rs:2:15
  |
2 |               keysym_get_name}, Capability, CompositorHandle, KeyboardHandle, KeyboardHandler,
  |               ^^^^^^^^^^^^^^^
3 |               KeyboardModifier, WLR_KEY_PRESSED};
  |               ^^^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

warning: unused import: `OutputHandle`
  --> awesome/src/objects/screen.rs:11:29
   |
11 | use wlroots::{Area, Origin, OutputHandle, Size};
   |                             ^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

warning: unused import: `lua::setup_lua`
  --> awesome/src/main.rs:40:5
   |
40 | use lua::setup_lua;
   |     ^^^^^^^^^^^^^^

warning: unused variable: `modifiers`
  --> way-cooler/src/input/keyboard.rs:14:13
   |
14 |         let modifiers = dehandle!(
   |             ^^^^^^^^^ help: consider using `_modifiers` instead
   |
   = note: #[warn(unused_variables)] on by default

warning: unreachable statement
  --> awesome/src/objects/mouse.rs:95:13
   |
95 | /             let mut screens: Vec<Screen> = lua.named_registry_value::<Vec<AnyUserData>>(SCREENS_HANDLE)?
[...]

PR #575, opened 2018-10-18
Fix the include statements in awesome/root.rs

AdminXVII:
The include statements in the tests in awesome/root.rs weren't using the latest code organisation, importing from super::super.
Now they import from ::root and ::objects.

Fix for: https://github.com/way-cooler/way-cooler/issues/574


Timidger:
Awesome, thanks! :)

PR #577, opened 2018-10-19
cargo fmt

AdminXVII:
Use the standard rust style within the codebase


Timidger:
AFAIK rustfmt isn't stable yet. Until it is, I don't want to use it as I ran into too many logistical issues when the results kept changing.

Once it is stable I'm more than happy to make it part of the submit process.

AdminXVII:
Ok
It is almost stable and, except for macros, the unstable parts don't concern this repo

Timidger:
> It is almost stable and, except for macros, the unstable parts don't concern this repo

Thanks for doing the research. I will still wait for it to be stabilized if only for my own peace of mind.

PR #582, opened 2018-10-25
Implement the clients property for tag, stored in an associate value.

AdminXVII:
Implement the public API for the client property, except for the signals.
To avoid the Rust-side cloning, values are stored as-is in an associate table (`__clients`) on the object.
This allows the clients to be shared amongst tags as the same Lua-side tables.
A bottleneck occurs when setting the new clients, as one needs to iterate over the old and new client lists to check for common values. This should not be a performance problem though, as it is not used frequently.

### DIFF
--- a/awesome/src/objects/tag.rs
+++ b/awesome/src/objects/tag.rs
@@ -50,15 +50,14 @@ impl<'lua> Tag<'lua> {
         //   Since it does not generally treat big arrays, this may be acceptable,
         //   However a faster algorithm would not hurt if someone finds one
 
-        let prev_clients: HashSet<_> = self.get_clients()?
-                                           .iter()
-                                           .map(|obj| Client::cast(obj.clone().into())
-                                                              .unwrap())
-                                           .collect();
-        let clients: HashSet<_> = clients.iter()
-                                         .map(|obj| Client::cast(obj.clone().into())
-                                                            .unwrap())
-                                         .collect();
+        let prev_clients = self.get_clients()?
+                               .iter()
+                               .map(|obj| Client::cast(obj.clone().into()))
+                               .collect::<rlua::Result<HashSet<_>>>()?;
+        let clients = clients.iter()
+                             .map(|obj| Client::cast(obj.clone().into()))
+                             .collect::<rlua::Result<HashSet<_>>>()?;
+
             
         for client in clients.difference(&prev_clients) {
             // emit signal
@@ -68,23 +67,23 @@ impl<'lua> Tag<'lua> {
             // TODO: emit signal and garbage if not referenced anymore
         };
 
-        let new_clients: Vec<_> = clients
-                            .difference(&prev_clients)
-                            .cloned()
-                            .collect();
-        self.0.table()?.set("__clients", new_clients)?;
+        self.0.table()?.set("__clients", clients.into_iter().collect::<Vec<_>>())?;
         Ok(Value::Nil)
     }
 
-    pub fn client_index(&self, client: &Client) -> Option<usize> {
-        self.get_clients().unwrap().iter().position(|c| {
-            Client::cast(c.clone().into()).unwrap() == *client
-        })
+    pub fn client_index(&self, client: &Client) -> rlua::Result<Option<usize>> {
+        // TODO: remove the chaining of collect and into_iter
+        Ok(self.get_clients()?
+                          .iter()
+                          .map(|obj| Client::cast(obj.clone().into()))
+                          .collect::<rlua::Result<Vec<Client>>>()?
+                          .into_iter()
+                          .position(|c| c == *client))
     }
 
     pub fn tag_client(&mut self, obj: AnyUserData<'lua>) -> rlua::Result<Value> {
         let client = Client::cast(obj.clone().into())?;
-        if let Some(_) = self.client_index(&client) { // if it is already part of the clients
+        if let Some(_) = self.client_index(&client)? { // if it is already part of the clients
             return Ok(Value::Nil);
         }
         let mut clients: Vec<AnyUserData> = self.0.table()?.get("__clients")?;
@@ -97,7 +96,7 @@ impl<'lua> Tag<'lua> {
         let client = Client::cast(obj.clone().into())?;
         let mut clients: Vec<AnyUserData> = self.0.table()?.get("__clients")?;
 
-        match self.client_index(&client) {
+        match self.client_index(&client)? {
             Some(index) => {
                 clients.remove(index);
                 self.0.table()?.set("__clients", clients)?;
@@ -431,6 +430,25 @@ assert(#t:clients() == 0, "Cannot get the clients")
              r#"
 local c = client{}
 local t = tag{}
+t:clients({ c })
+assert(c, "client doesn't exists")
+assert(#t:clients() == 1, "Tag doesn't have the clients")
+assert(t:clients()[1] == c, "Pass by value, not by reference")
+"#,
+             None
+        ).unwrap()
+    }
+
+    #[test]
+    fn tag_set_client_double() {
+        let lua = Lua::new();
+        tag::init(&lua).unwrap();
+        client::init(&lua).unwrap();
+        lua.eval(
+             r#"
+local c = client{}
+local t = tag{}
+t:clients({ c })
 t:clients({ c })
 assert(c, "client doesn't exists")
 assert(#t:clients() == 1, "Tag doesn't have the clients")
--- a/awesome/src/objects/client.rs
+++ b/awesome/src/objects/client.rs
@@ -10,7 +10,7 @@ use rlua::{self, Lua, Table, ToLua, UserData, Value};
 use common::{class::{self, Class, ClassBuilder},
              object::{Object, Objectable}};
 
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, Hash)]
 pub struct ClientState {
     // TODO Fill in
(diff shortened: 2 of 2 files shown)


awesome/src/objects/tag.rs
```diff
@@ -8,10 +8,12 @@ use rlua::{self, AnyUserData, Integer, Lua, Table, ToLua, UserData,
            UserDataMethods, Value};
 
 use common::{class::{self, Class, ClassBuilder},
-             object::{self, Object, Objectable},
+             object::{self, Object, ObjectBuilder, Objectable},
              property::Property,
              signal};
 
+use objects::client::{ClientState, Client};
```
Timidger: Group these together with the common:: group above.

I haven't been _real_ strict on this, but generally I've been separating the imports like this:

```
<std imports>

<external crate imports>

<project imports>
```

awesome/src/objects/tag.rs
```diff
@@ -34,8 +36,87 @@ impl Default for TagState {
...
+
+    pub fn set_clients(&mut self, clients: rlua::Table) -> rlua::Result<Value> {
+        // TODO: this is a really inefficient solution ( O(n*m) for the search )
+        //   Since it does not generally treat big arrays, this may be acceptable,
+        //   However a faster algorithm would not hurt if someone finds one
+
+        let prev_clients = self.get_clients()?;
+
+        // Indexes refering to prev_clients
+        let mut not_common_clients: Vec<u32> = (1..(prev_clients.len()? + 1) as u32)
```
Timidger: Use the `..=` syntax here, instead of adding `+1` to the `len`.

awesome/src/objects/tag.rs
```diff
@@ -34,8 +36,87 @@ impl Default for TagState {
...
+                                                    .collect();
+        let mut num_common_clients = 0;
+            
+        for client in clients.clone().sequence_values::<AnyUserData>() {
+            let mut idx = -1;
```
Timidger: Use an `Option<i32>` here instead, default to `None`.

awesome/src/objects/tag.rs
```diff
@@ -34,8 +36,87 @@ impl Default for TagState {
...
+                .handle_constructor_argument(args)?
+                .build())
+    }
+
+    pub fn get_clients(&self) -> rlua::Result<Table<'lua>> {
+        self.0.table()?.get("__clients")
+    }
+
+    pub fn set_clients(&mut self, clients: rlua::Table) -> rlua::Result<Value> {
+        // TODO: this is a really inefficient solution ( O(n*m) for the search )
```
Timidger: Yikes that is complicated...

Is there a reason we can't use [`std::HashSet`](https://doc.rust-lang.org/std/collections/struct.HashSet.html) to filter out the clients to add instead of this convulted logic?

awesome/src/objects/tag.rs
```diff
@@ -34,8 +36,87 @@ impl Default for TagState {
...
+        self.0.table()?.set("__clients", clients)?;
+        Ok(Value::Nil)
+    }
+
+    pub fn client_index(&self, client: &Client) -> rlua::Result<Integer> {
+        let clients = self.get_clients()?;
+        let self_ref = &*client.state()? as *const _;
+        for pair in clients.pairs::<Integer, AnyUserData<'lua>>() {
+            let (i, c) = pair?;
+            if &*c.borrow::<ClientState>()? as *const _ == self_ref {
```
Timidger: Can you not just cast the `client` to an `Object` to get to the `AnyUserData` to compare directly against the list of clients (and thus defeat the need for using this pointer comparison logic).

awesome/src/objects/tag.rs
```diff
@@ -34,8 +36,87 @@ impl Default for TagState {
...
+        Ok(Value::Nil)
+    }
+
+    pub fn untag_client(&mut self, obj: AnyUserData<'lua>) -> rlua::Result<Value> {
+        let client = Client::cast(obj.clone().into())?;
+        let index = self.client_index(&client)?;
+        let clients = self.get_clients()?;
+
+        for i in index..clients.len()? {
+            clients.set(i, clients.get::<_, Value>(i + 1)?)?;
```
Timidger: Don't do this. Store `__clients` as a `Vec<AnyUserdata>`. [Vec has an auto impl ToLua provided by the rlua crate](https://docs.rs/rlua/0.15.3/rlua/prelude/trait.FromLua.html#impl-FromLua%3C%27lua%3E). Then you can simply cast it to a `Vec<ClientState>` and not have to do this weirdness.

Timidger: If you need guidance on this, look at how `screen.rs` stores screen handles in its `SCREEN_HANDLE`. Just grep `SCREEN_HANDLE` in `screen.rs`.

awesome/src/objects/tag.rs
```diff
@@ -171,20 +262,23 @@ fn get_activated<'lua>(_: &'lua Lua, obj: AnyUserData<'lua>) -> rlua::Result<Val
     Ok(Value::Boolean(obj.borrow::<TagState>()?.activated))
 }
 
-fn get_clients<'lua>(lua: &'lua Lua, _obj: AnyUserData<'lua>) -> rlua::Result<Value<'lua>> {
-    // TODO / FIXME: Do this properly.
-    // - Actually return clients.
+fn get_clients<'lua>(_lua: &'lua Lua,  (obj, val): (AnyUserData<'lua>, Value<'lua>)) -> rlua::Result<Table<'lua>> {
+    // TODO:
```
Timidger: Remove outdated TODO?

awesome/src/objects/tag.rs
```diff
@@ -326,4 +420,123 @@ assert(called_selected == 1)
                  None
         ).unwrap()
     }
+
+    #[test]
```
Timidger: Yay tests! I like tests :+1:

awesome/src/objects/tag.rs
```diff
@@ -171,20 +262,23 @@ fn get_activated<'lua>(_: &'lua Lua, obj: AnyUserData<'lua>) -> rlua::Result<Val
...
     // - Right now this is a property that returns a function. Can we get rid of
-    // some of this   indirection?
-    Ok(Value::Function(lua.create_function(|lua, _: ()| {
-                                                lua.create_table()
-                                            })?))
+    // some of this indirection?
+    let mut tag = Tag(obj.into());
+    match val {
+        Value::Table(clients) => tag.set_clients(clients),
+        _ => Ok(Value::Nil),
```
Timidger: Did you mean to return early here? Because this deosn't do anything

awesome/src/objects/tag.rs
```diff
@@ -326,4 +420,123 @@ assert(called_selected == 1)
...
+        tag::init(&lua).unwrap();
+        client::init(&lua).unwrap();
+        lua.eval(
+             r#"
+local c = client{}
+local t = tag{}
+t:clients({ c })
+assert(c, "client doesn't exists")
+assert(#t:clients() == 1, "Tag doesn't have the clients")
+assert(t:clients()[1] == c, "Pass by value, not by reference")
```
Timidger: You should be modifying `c` in between this and the insert call shouldn't you to assure it is actually by reference?

awesome/src/objects/tag.rs
AdminXVII: But neither Object nor AnyUserData implements the Eq trait. One could change the object implementation, but it would come back to reference comparison, only inside the object. Unless you mean comparison by value, I can't see what you mean, sorry.

AdminXVII:
Done.
The Client are now compared by value to avoid comparing pointers, but it should be trivial to compare them by reference if it is deemed necessary in the future

awesome/src/objects/tag.rs
```diff
@@ -49,46 +50,36 @@ impl<'lua> Tag<'lua> {
...
-                return clients.iter().position(|ref c| {
-                    let state_ref = &*state as *const _;
-
-                    &*c.borrow::<ClientState>().unwrap() as *const _ == state_ref
-                })
-            }
-        }
-        None
+        self.get_clients().unwrap().iter().position(|c| {
+            Client::cast(c.clone().into()).unwrap() == *client
```
Timidger: You'll want to collect the casts as well. Cast these to clients separately in a `map`. IIRC it's possible to `collect` these into a `Result<Vec<Client>, rlua::Error>` but I can't remember how.

If this is too annoying to do, then you can leave the unwrap since this is much less likely to happen.

Timidger:
Ah, I made a mistake. You should be doing the pointer comparison logic I just for some reason thought you could do that using a `AnyUserData`. Turns out that doesn't implement `Eq`/`PartialEq`. Oops! Value comparison might still work, but Awesome does it by pointer and we should mirror how they do it.

Sorry for the confusion.

awesome/src/objects/tag.rs
```diff
@@ -49,46 +50,36 @@ impl<'lua> Tag<'lua> {
...
             // TODO: emit signal and garbage if not referenced anymore
-        }
-        self.0.table()?.set("__clients", clients)?;
+        };
+
+        let new_clients: Vec<_> = clients
+                            .difference(&prev_clients)
+                            .cloned()
+                            .collect();
+        self.0.table()?.set("__clients", new_clients)?;
```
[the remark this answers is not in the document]
AdminXVII: Oops yes it should be clients. Not the union, though, as Awesome makes a replace.
The set difference is useful to make sure a client does not receive `untagged` then `tagged` signals, but it is out of place for the return.

Issue #476, opened 2018-02-04
Make AwesomeWM's root not an object


Timidger:
Currently we have `root` as an object in our Awesome code, but that's not correct since it's not an object that can be instantiated! That functionality should be removed.

PR #583, opened 2018-10-25
Change the root module from an object to a simple table

AdminXVII:
In awesomeWM, the root module is a collection of utilities which does not store state except for the globalconf.
The globalconf's properties are already provided by external means, so the root only need to be a table with the corresponding functions as properties.

Resolve: https://github.com/way-cooler/way-cooler/issues/476


Timidger:
Thanks!

PR #587, opened 2018-10-28
Remove a piece of code which was duplicate.

AdminXVII:
https://github.com/way-cooler/way-cooler/blob/2c6e6f17b8e9cb547b913344f9ac9b512e06e555/awesome/src/main.rs#L205-L235
https://github.com/way-cooler/way-cooler/blob/2c6e6f17b8e9cb547b913344f9ac9b512e06e555/awesome/src/keygrabber.rs#L114-L134


Timidger:
Thanks!

PR #589, opened 2018-11-02
Remove all the warnings

AdminXVII:
With all the methods that aren't used, the dead code warnings polute the output of compilation and obfuscate the valuable warnings during development.
This might be a bit controversial, but it it will be easier to grep for the different `allow` annotations after version 1.0 than to fix a lot of warnings which should be caught during the process.

### DIFF
--- a/awesome/src/common/class.rs
+++ b/awesome/src/common/class.rs
@@ -55,6 +55,7 @@ impl<'lua, S: ObjectStateType> ClassBuilder<'lua, S> {
     }
 
     // TODO remove, do right
+    #[allow(dead_code)]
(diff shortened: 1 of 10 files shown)


Timidger:
Agreed

AdminXVII:
In fact, it may be interesting to [deny warnings](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md) during the test phase to make sure the warnings don't crawl back and to make sure all warnings are documented and fully acknowledged in the pull requests.

Timidger:
Yes that would be best. I was going to suggest that before merging this, but forgot about it until I saw this later

PR #590, opened 2018-11-06
Force no warnings in test phase

AdminXVII:
Add annotations to disallow warnings when testing, based upon https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md.
Also change the build script, as `test` already includes `build` and removing `--all` permits to leverage caching if it is ever implemented on the backend.
Fixed a dependency (gcc => cc) to remove warnings, and fixed the way-cooler part.

Related: #589

### DIFF
--- a/awesome/src/main.rs
+++ b/awesome/src/main.rs
@@ -1,5 +1,38 @@
 //! Awesome compatibility modules
 
+#![cfg_attr(test, deny(bad_style,
+       const_err,
+       dead_code,
+       improper_ctypes,
+       legacy_directory_ownership,
+       non_shorthand_field_patterns,
+       no_mangle_generic_items,
+       overflowing_literals,
+       path_statements ,
+       patterns_in_fns_without_body,
+       plugin_as_library,
+       private_in_public,
+       private_no_mangle_fns,
+       private_no_mangle_statics,
+       safe_extern_statics,
+       unconditional_recursion,
+       unions_with_drop_fields,
+       unused,
+       unused_allocation,
+       unused_comparisons,
+       unused_parens,
+       while_true))]
+
+// Allowed by default
+#![cfg_attr(test, deny(missing_docs,
+       trivial_numeric_casts,
+       unused_extern_crates,
+       unused_import_braces,
+       unused_qualifications))]
+
+// May be good to add
+// #![cfg_attr(test, warn(unused_results))]
+
 extern crate cairo;
 extern crate cairo_sys;
 extern crate env_logger;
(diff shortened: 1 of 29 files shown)


awesome/src/main.rs
```diff
@@ -1,5 +1,38 @@
...
+
+// Allowed by default
+#![cfg_attr(test, deny(missing_docs,
+       trivial_numeric_casts,
+       unused_extern_crates,
+       unused_import_braces,
+       unused_qualifications))]
+
+// May be good to add
+// #![cfg_attr(test, warn(unused_results))]
```
Timidger: Lets go ahead and add this to the deny for testing.

way-cooler/src/main.rs
```diff
@@ -1,18 +1,46 @@
...
+// Allowed by default
+#![cfg_attr(test, deny(missing_docs,
+       trivial_casts,
+       trivial_numeric_casts,
+       unused_extern_crates,
+       unused_import_braces,
+       unused_qualifications))]
+
+// May be good to add
+// #![cfg_attr(test, warn(unused_results))]
```
Timidger: Same as in awesome

AdminXVII:
Done.
This implies lots of changes though, some of which you may not like (i.e. using `let _ =` when the value is really useless). It was why I commented it at first.

awesome/src/objects/tag.rs
```diff
@@ -69,29 +69,23 @@ impl<'lua> Tag<'lua> {
...
-                clients.remove(index);
-                self.set_associated_data("__clients", clients)?;
-                Ok(Value::Nil)
-            }
-            None => Err(rlua::Error::RuntimeError("Client to untag was not tagged".into()))
-        }
+    pub fn untag_client(&mut self, client: Client<'lua>) -> rlua::Result<()> {
+        let clients: Vec<_> = self.clients()?
+                              .into_iter()
+                              .filter(|c| *c != client)
```
Timidger: This is not the same logic as there should be an error signaled if the client wasn't in the list

AdminXVII: In fact it is closer to Awesome's bahaviour

```
void
untag_client(client_t *c, tag_t *t)
{
    for(int i = 0; i < t->clients.len; i++)
        if(t->clients.tab[i] == c)
        {
            lua_State *L = globalconf_get_lua_State();
            client_array_take(&t->clients, i);
            banning_need_update();
            ewmh_client_update_desktop(c);
            screen_update_workarea(c->screen);
            tag_client_emit_signal(t, c, "untagged");
            luaA_object_unref(L, t);
            return;
        }
}
```

PR #591, opened 2018-11-11
Awesome top bar as xdg shell

Timidger:
This PR essentially gets Way Cooler back to the point it was at before we split up the compositor and awesome work into two processes.

This adds the xdg protocol to the awesome client side as that is the standard protocol we will use to render the top bar. It's possible we might want to switch to layer shell, however I haven't finished wrapping that in wlroots-rs so Way Cooler currently doesn't support it.

This isn't totally ready though because there is a type check I had to disable to get this to work...and I need to make sure it's something I can't work around before I merge this.

Also removed the unused result check on CI/Test, because that was a mistake.

# TODO
- [x] Fix that type check bug
- [x] Ensure that updating wlroots fixes the segfault when killing `awesome`.
  - It causes Way Cooler to segfault, but it happens in rootston as well (!) and the error reported makes it look like a bug in how they handle their `wl_list`s.
- [ ] ~Make `awesome` work when using `way-cooler` in DRM~ Lets do this in another patch
  - It's not working because we aren't setting up xwayland correctly and it's causing an early disconnect.

### DIFF
--- a/awesome/src/common/object.rs
+++ b/awesome/src/common/object.rs
@@ -196,7 +196,7 @@ impl<'lua, S: ObjectStateType> ToLua<'lua> for Object<'lua, S> {
 impl<'lua, S: ObjectStateType> FromLua<'lua> for Object<'lua, S> {
     fn from_lua(val: Value<'lua>, _lua: &'lua Lua) -> rlua::Result<Self> {
         if let Value::UserData(obj) = val {
-            Object::cast(obj)
+            Ok(obj.into())
         } else {
             Err(rlua::Error::RuntimeError("Invalid data supplied".into()))
         }
--- a/awesome/src/objects/tag.rs
+++ b/awesome/src/objects/tag.rs
@@ -63,12 +63,14 @@ impl<'lua> Tag<'lua> {
         Ok(())
     }
 
+    #[allow(dead_code)]
     pub fn client_index(&self, client: &Client) -> rlua::Result<Option<usize>> {
         // TODO: remove the chaining of collect and into_iter
         Ok(self.clients()?.iter()
                           .position(|c| *c == *client))
     }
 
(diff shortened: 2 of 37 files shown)


awesome/src/common/object.rs
```diff
@@ -196,7 +196,7 @@ impl<'lua, S: ObjectStateType> ToLua<'lua> for Object<'lua, S> {
 impl<'lua, S: ObjectStateType> FromLua<'lua> for Object<'lua, S> {
     fn from_lua(val: Value<'lua>, _lua: &'lua Lua) -> rlua::Result<Self> {
         if let Value::UserData(obj) = val {
-            Object::cast(obj)
+            Ok(obj.into())
```
Timidger: This is the type check I removed in order to get this to work.

if it's left as is this crashes when loading the configuration while setting the index miss handler in `gear/objects/properties.lua` on line 83 or so. When I dug into it a bit, it expects it to be a screen state (i.e. `S` = `ScreenState`) but it's not that. However when using the `into` here it just works oddly enough?

awesome/src/objects/tag.rs
```diff
@@ -95,9 +95,11 @@ impl UserData for TagState {
     }
 }
 
-pub fn init(lua: &Lua) -> rlua::Result<()> {
+pub fn init(lua: &Lua) -> rlua::Result<Class<TagState>> {
```
AdminXVII: The return values where removed in a recent commit because they were never used and caused errors. Did you find this unsatisfying, or are those changes only artifacts of commits on an old state?

Timidger: I found that unsatisfying, and I realise now that I shouldn't have added the unused result error on test. 

Plus it seemed weird to have a builder with no `build` method, even if it's technicaly not needed.

PR #593, opened 2018-11-17
Rewrote glib main loop init code from C to Rust

Timidger:
* :crab: Removed all of C used in Way Cooler :crab:
  - Side note, this probably wasn't necessary because the C code was _fine_. I'm doing this though because the initilization is going to get more complicated after I add dbus to `awesome` which would require extending that C code more than I'm comfortable with.
* Updated dependencies as a side effect

CC @psychon to ensure I'm setting up the glib loop properly (from my testing it does the same thing as before, e.g. it renders a broken status bar in the `way-cooler` compositor but the Rust crate had a weird interface for adding idle sources to the main loop).


psychon:
Quick first look: I would expect this to lead to 100% CPU usage, because glib does not sleep as long as there is an active idle source.

Timidger:
At one point in my changes it did, but then it was fixed I believe by returning false if flush returns zero.

psychon:
Ah, okay. But returning false from the idle callback means that the whole thing is removed and never runs again.

Timidger:
/sigh

We'll just keep the C code. It's not a huge issue. I'm more annoyed at the glib Rust library for being so incomplete, just like many Rust<-C wrapper libraries.

PR #594, opened 2018-11-22
Added style guide and rustfmt back in

### LINKED ISSUE #59: Project standards for code format, linting, etc
Way Cooler and rustwlc both have somewhat poor code formatting standards. While I am personally somewhat opposed to rustfmt and similar projects which limit the shape of source code, what we have right now is not very unified. We need a solid list of rules that we can point new contributors to that will be definitive. We also need to hold ourselves to these standards!
## Code format
- [x] Tabbing must be done at 4 spaces without the use of `\t`, this is final.
- [ ] Maximum line width? I suggest we take inspiration from the Linux kernel and have line width limits. However, in Rust our code can end up with many generic function signatures which can't fit in certain widths. 80 characters is also good for being able to read on two tabs at a time. 90 could also be an option.
- [ ] Import ordering - We've done a good job of (mostly) ordering `use` statements, but I've left a lot of newlines between them. Perhaps these should be cleaned up. Hopefully this becomes easier as editor support increases and we can stop typing them ourselves.
- [ ] `use` within methods - In a couple of places where we have an import with a narrow scope, it has been nice to have a `use` statement there and not at the top of the file. Should this be avoided or discouraged?
- [ ] Naming - should we look through the code and standardize naming?
- [ ] Modules - @Timidger has called me out many a time for placing too much code in `mod.rs`. We should look at other crates and come up with a scheme of when to split modules into multiple files.
- [ ] Rustwlc in particular has code that we wrote with little knowledge of Rust
- [ ] Cover code formatting in `clippy` or `rustfmt` config, prevent badly-formatted code from being pushed via travis? Too extreme? I'd rather have an automated service say no than nit through a PR for formatting the way they do in wlc. This would be cool for GitHub integration (i.e. the rustwlc bot from before)
## Comments
- [ ] Require doc comments all over the project? Use `#![deny(missing_docs)]`?
- [ ] Require more inline comments? Require comments documenting specific algorithms
- [ ] Doc tests - show off syntax of internal APIs, easy to help new contributors, but can slow down development and is really just designed for libraries
## Testing
- [x] Add coveralls badge when ready
- [x] Add travis build badge
- [ ] Consider running travis build through docker image (we may run into travis timeouts)
- [ ] Bring unit test coverage up, try to target edge cases
- [ ] Create a checklist of window manager user interactions in order to test functionality of tiling/keybindings/ipc/window open+close/X support/etc. especially to help with regressions
- [ ] Automate running that checklist using scripting, such as x tools, docker, etc. (very long term)
- [ ] Test code in callbacks
- [ ] Use dummy-rustwlc or other mocks to perform broader integration tests
- [ ] Analyze memory and CPU usage
## Lints
- [ ] Use of unwrap - I have been advocating to remove all use of `.unrap()` and replace with a (not always helpful) `.expect(message)`. Should this stay? Can we standardize error messages so there isn't as much overhead of calling `.expect()`?
- [ ] Error types - internally, we have been fine with matching on enum fields to discern the type of error found. Should we move towards implementing the `Error` type and including messages?
## Structure
- [ ] I know I have to rewrite `registry`, `ipc`, `lua`, and probably `commands` and `keys` to use a struct instead of global methods. How should we consider this for other parts of way-cooler?
## Dependencies
- [X] If we can write a fix or solve a problem internally, there's a turning point somewhere between "don't rewrite the standard libraries" and "using left-pad". I advocate a conservative use of external dependencies - especially ones which are close to features "plain Rust" has already. For example, I don't think we should use `quickerror` or `tempfile`, as we can just write `impl Error for MyErrorType` ourselves, thanks, and `tempfile` is `std::fs::temp_path`. But I do love the `json!` macro because it makes code across the IPC a lot more concise.
- [x] I think the `unix_socket` stuff is in `std` by now, we should get rid of that
## Documentation
- [ ] We need to move docs from a folder in the source to the wiki (and link to it in the readme)
- [ ] We need to move the cargo docs of rustwlc to a subdirectory of Immington Industries GitHub page
- [ ] Can we run `cargo doc` with fancy flags so that non-exported code (i.e. all of it) is documented?
- [ ] High-level view of the structure of our code?
- [ ] Since we are tagging releases now and want to move to a faster schedule, should we start to utilize the milestone feature of Github? e.g: one for 3.2.0?

Feel free to comment with anything I should add (or insert directly into this @Timidger)

Timidger:
Now that rustfmt is _finally_ 1.0 we can use it again. So this is a big format PR.

We've earmarked the CI to only use 1.0.0.

Also added a style guide for some of the other things that rustfmt doesn't catch.

Fixes #59


Timidger:
Not going to merge this until I get the PR reviews fixed _and_ I merge in #596.

PR #596, opened 2018-11-24
Set up Way Cooler to run AwesomeWM tests

Timidger:
Adds the original Awesome as a submodule (in order to get its tests/ directory, in the future that folder should probably become its own repository). Added a special `run_awesome.sh` in that tests/ directory that sets up Way Cooler.

Re-adds D-Bus to Way Cooler, exposed to lua via the global `dbus` object.

You can run these tests from the root of the repository (after initializing the submodules) by running `make test`

Thanks to @psychon helping me figure out why D-Bus wasn't working at first.


Elv13:
Just to point out that for now maybe starting with a copy of the shims might be a possibility given you already have some of their API "really" implemented. The shims can save you time to get the tests to pass and then one by one replacing them with the real deal, but loading all of the shims at once will ~~probably~~ conflict with other code paths.

Timidger:
@Elv13 Yes thats' what I'm doing. Currently it's still crashing, but it does get it closer

Timidger:
`make test` fails, and there's no way to run it headlessly yet so it has not been added as part of the CI.

It fails due to not enough code being setup for even the CI. That will eventually change later, but for now at least the pieces are here.

PR #598, opened 2018-11-24
way-cooler: xdg-shell add view on map

acrisci:
Add a view when it is mapped instead of when the shell surface is
created.

Also fixes a bug where xdg popups were added to the list of views.

Needs map-unamp implementation on wlroots-rs: https://github.com/swaywm/wlroots-rs/pull/218

### DIFF
--- a/way-cooler/src/shells/xdg.rs
+++ b/way-cooler/src/shells/xdg.rs
@@ -1,7 +1,7 @@
 use std::rc::Rc;
 
 use wlroots::{CompositorHandle, Origin, SurfaceHandle, SurfaceHandler,
-              XdgShellSurfaceHandle, XdgShellHandler, XdgShellManagerHandler, XdgShellState::*};
+              XdgShellSurfaceHandle, XdgShellHandler, XdgShellManagerHandler};
 use wlroots::xdg_shell_events::{MoveEvent, ResizeEvent};
 
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


way-cooler/src/shells/xdg.rs
```diff
@@ -1,7 +1,7 @@
 use std::rc::Rc;
 
 use wlroots::{CompositorHandle, Origin, SurfaceHandle, SurfaceHandler,
-              XdgShellSurfaceHandle, XdgShellHandler, XdgShellManagerHandler};
+              XdgShellSurfaceHandle, XdgShellHandler, XdgShellManagerHandler, XdgShellState::*};
```
Timidger: Don't do a glob import for enums at the top level, put this use statement in the function you use it (e.g `map_request`)

way-cooler/src/shells/xdg.rs
```diff
@@ -100,6 +100,59 @@ impl XdgShellHandler for Xdg {
...
+    fn map_request(&mut self,
+                   compositor: CompositorHandle,
+                   _: SurfaceHandle,
+                   shell_surface_handle: XdgShellSurfaceHandle) {
+        let is_toplevel = with_handles!([(shell_surface: {&shell_surface_handle})] => {
+            match shell_surface.state().unwrap() {
+                TopLevel(_) => true,
+                _ => false
+            }
+        }).unwrap();
```
Timidger: `dehandle!` returns a value, so you can instead do:

```rust
        let is_toplevel = dehandle!(
            @shell_surface = {&shell_surface_handle};
            match shell_surface.state().unwrap() {
                TopLevel(_) => true,
                _ => false
            }
        );
```

and there's no need to do the ugly `unwrap()` at the end.

Can you also put a comment about the shell surface handle being used in `seat.focus_view` which is why it can't be combined with the next `dehandle!` block? I rant into that pitfall just now while reviewing your code.

way-cooler/src/shells/xdg.rs
```diff
@@ -100,6 +100,59 @@ impl XdgShellHandler for Xdg {
...
+            let ::Server { ref mut seat,
+                           ref mut views,
+                           ref cursor,
+                           ref mut xcursor_manager,
+                           .. } = *server;
+            let destroyed_shell = shell_surface.into();
+            views.retain(|view| view.shell != destroyed_shell);
+
+            if views.len() > 0 {
+                seat.focus_view(views[0].clone(), views);
```
Timidger: This would be better written as:

```rust
if let Some(view) = views.get(0).cloned() {
    seat.focus_view(view, views);
} else ...

```

PR #599, opened 2018-11-24
way-cooler seat: end keyboard grab before focus

acrisci:
If a surface has the keyboard grab at the time we switch focus, it will
get the keyboard event instead of the intended surface. End the grab to
reset keyboard focus before sending enter notify.


acrisci:
To reproduce the bug:

Open two windows. Open a popup menu (like the "file" menu on gnome-terminal) then click into the other window. The previous window will still have keyboard focus (if you type, the keyboard keys will go to the unfocused window).

Timidger:
Thanks, sorry I got distracted and this took a while to verify.

PR #601, opened 2018-12-08
Fix `make run`

Hi-Angel:
`make run` command resulted in `awesome` failing because it couldn't find the shims. Fix this by taking into account that path of shims is relative to the path where binaries are.

### DIFF
--- a/awesome/src/lua/mod.rs
+++ b/awesome/src/lua/mod.rs
@@ -55,9 +55,9 @@ fn load_shims(lua: &Lua) {
     let globals = lua.globals();
     let package: Table = globals.get("package").unwrap();
     let mut path = package.get::<_, String>("path").unwrap();
-    path.push_str(&format!(";{0}/?.lua;{0}/?/init.lua", SHIMS_PATH));
+    let shims_path: PathBuf = std::env::current_exe().unwrap().parent().unwrap().join(SHIMS_PATH);
+    path.push_str(&format!(";{0}/?.lua;{0}/?/init.lua", shims_path.to_str().unwrap()));
     package.set("path", path).unwrap();
-    let shims_path = PathBuf::from(SHIMS_PATH);
     for shim in SHIMS.iter() {
         let mut path = shims_path.clone();
         path.push(format!("{}.lua", shim));


awesome/src/lua/mod.rs
```diff
@@ -52,12 +37,26 @@ thread_local! {
 ///
 /// Best way to help out: comment out one of these lines, fix what breaks.
 fn load_shims(lua: &Lua) {
+    let shims_path: PathBuf = std::env::current_exe().unwrap().parent()
+        .ok_or(()).unwrap().join("../../tests/awesome/tests/examples/shims/");
```
Timidger: Why did you put `ok_or(())` here? This drops important error information for no benefit.

awesome/src/lua/mod.rs
```diff
@@ -52,12 +37,26 @@ thread_local! {
...
+                               "button",
+                               "keygrabber",
+                               "mousegrabber",
+                               "key"];
     let globals = lua.globals();
     let package: Table = globals.get("package").unwrap();
     let mut path = package.get::<_, String>("path").unwrap();
-    path.push_str(&format!(";{0}/?.lua;{0}/?/init.lua", SHIMS_PATH));
+    path.push_str(&format!(";{0}/?.lua;{0}/?/init.lua",
+                           shims_path.to_str().ok_or(()).unwrap()));
```
Timidger: Same here, remove this `ok_or`

awesome/src/lua/mod.rs
```diff
@@ -52,12 +37,26 @@ thread_local! {
 ///
 /// Best way to help out: comment out one of these lines, fix what breaks.
 fn load_shims(lua: &Lua) {
+    let shims_path: PathBuf = std::env::current_exe().unwrap().parent()
+        .ok_or(()).unwrap().join("../../tests/awesome/tests/examples/shims/");
+    /// Shims to load
+    const SHIMS: [&str; 11] = ["awesome",
```
Timidger: Please keep this at the top of the file, where it was before.

awesome/src/lua/mod.rs
Timidger: Please keep the shims path as a constant at the top of the file as well.

Hi-Angel: Sorry, I'm not very versed in Rust yet. What should I've used instead?

awesome/src/lua/mod.rs
Hi-Angel: But it's confusing. When it's in the top, it looks like something important, so you need to find all usages of it if you debug something relevant to it.

Whereas when the scope is limited to where it's used, it's obvious that it's a one-time constant.

awesome/src/lua/mod.rs
Timidger: Just remove the `ok_or(())` call.

awesome/src/lua/mod.rs
Timidger: It is important, its all the shims that we still need for Way Cooler to "work". 

Any editor worth its salt will have a "jump to definition" that is language agnostic (or you can try to get the Rust Language Server working, good luck) so it shouldn't be hard to look at the definition if you need it.

Finally, I didn't detail it in the style guide but constants should go at the top. That' standard in most language default "idiomatic" style which is why I forgot it.

Hi-Angel: Ok, I'll change it. That said, you misunderstood me, I didn't mean that it's hard to find the definition. I meant that it's easier to understand the codebase when stuff has as limited visibility scope as possible, because you don't need to check every place where something is used.

But okay, I'll count it as a specific of that particular project, and will stick to that style.

Hi-Angel:
Thanks for your comments, I addressed everything in the commit.

Timidger:
Thanks!

PR #602, opened 2018-12-08
README: fix link to irc channel

Hi-Angel:
I don't think this ever meant to point to freenode since the channel
there is almost empty, don't seem to be related to Way-Cooler or
Awesome, and doesn't have Timidger :)


Timidger:
Ah thanks, that is what I meant

PR #603, opened 2018-12-18
docs: contributing: IRC link should point to oftc

Hi-Angel:
Just found another place in docs where freenode was referred to instead of oftc.


Timidger:
Thanks!

PR #604, opened 2019-01-01
awesome/build: rebuild when a wayland protocol changed or added

Hi-Angel:
It's a small change, but I think it's pretty useful nevertheless.


Timidger:
TIL about that flag.

Thanks!

PR #605, opened 2019-01-04
Restore way-cooler.desktop

ilpianista:
It looks like this went lost in 207d2e74 or is there any reason behind its removal?


Timidger:
The startup is more complex now. Both awesome and Way Cooler binaries need to be started. Also Way Cooler doesn't use dbus any more.

PR #607, opened 2019-04-25
wlroots: update

groxxda:
Updates wlroots to current master.

Please review the keygrabber.rs change since I'm fairly new to rust :)


Timidger:
Thanks! Unfortunately I have [another patchset](https://github.com/way-cooler/way-cooler/tree/galvanization) I'm working on right now that fixes up the wlroots dependency on the client side so this will conflict with those changes. Hopefully I can get it merged soon, but that means these changes aren't going to be very helpful unfortunately. 

Sorry about that! Otherwise it looked good.

PR #608, opened 2019-04-27
Update wayland, remove wayland-scanner

groxxda:
* Update to the current version of wayland-rs
  * Use the EventHandler struct instead of single callback methods with `match event` inside
  * Use GlobalsImplementor to provide `NewProxy` with the implementation

* Drop xdg_shell protocol and use the one from wayland-rs
  * Drop the protocol generation code for now

Make user_data code safe (hopefully :))

The fields inside their wrapper types are still named `proxy` even though they don't refer to `Proxy` instances anymore. They should probably be renamed to a more descriptive name.


Timidger:
Hmm, both the build server and my local machine can't build due to an error in buildng wayland-sys.

Timidger:
[Opened up an issue on wayland-rs bug tracker](https://github.com/Smithay/wayland-rs/issues/261), not sure what the problem is. Probably on their end

awesome/src/wayland_obj/output.rs
```diff
@@ -25,98 +37,109 @@ struct OutputState {
...
+    }
+
+    pub fn name(&self) -> String {
+        unwrap_state(self.as_ref()).name.clone()
+    }
+}
+
+impl GlobalImplementor<WlOutput> for WlOutputManager {
+    fn new_global(&mut self, new_proxy: NewProxy<WlOutput>) -> WlOutput {
+        let res = new_proxy.implement(WlOutputEventHandler {}, Mutex::new(OutputState::default()));
```
Timidger: We are single threaded, is there a way to avoid this mutex?

awesome/src/wayland_obj/xdg_shell.rs
```diff
@@ -1,19 +1,21 @@
...
-
-use crate::wayland_protocols::xdg_shell::{
-    xdg_surface::{self, RequestsTrait as XdgSurfaceTrait, XdgSurface},
-    xdg_toplevel::{self, RequestsTrait as XdgToplevelTrait},
-    xdg_wm_base::{RequestsTrait as XdgWmBaseTrait, XdgWmBase}
+use std::{
+    cell::RefCell,
+    fmt,
+    os::unix::io::AsRawFd,
+    sync::{Mutex, MutexGuard}
```
Timidger: This mutex too?

awesome/src/wayland_obj/xdg_shell.rs
```diff
@@ -102,103 +92,123 @@ impl XdgToplevel {
...
-                Box::from_raw(user_data);
-            }
-        }
+        // unsafe {
+        //     let user_data = self.proxy.get_user_data() as *mut XdgToplevelState;
+        //     if !user_data.is_null() {
+        //         Box::from_raw(user_data);
+        //     }
+        // }
+        // TODO(ried): leak? free user data?
```
Timidger: This can just be deleted I believe, since the user data is owned by the proxy object itself I think.

awesome/src/wayland_obj/xdg_shell.rs
```diff
@@ -102,103 +92,123 @@ impl XdgToplevel {
...
-        .map(|toplevel| XdgToplevel::new(surface, xdg_surface, toplevel))
+        .get_toplevel(|new_toplevel| {
+            let state = XdgToplevelState {
+                size: Size::default(),
+                wl_surface: wl_surface.clone(),
+                xdg_surface: xdg_surface.clone(),
+                buffer: None
+            };
+
+            let res = new_toplevel.implement(XdgToplevelHandler {}, Mutex::new(state));
```
Timidger: This mutex too?

awesome/src/wayland_obj/xdg_shell.rs
```diff
@@ -102,103 +92,123 @@ impl XdgToplevel {
...
+        })
+        .map(|toplevel| XdgToplevel { proxy: toplevel })
+}
+
+struct XdgSurfaceHandler {
+    wlsurface: WlSurface
+}
+
+impl xdg_surface::EventHandler for XdgSurfaceHandler {
+    #[allow(unused)]
```
Timidger: Is this annotation necessary?

awesome/src/wayland_obj/output.rs
groxxda: Changed to RefCell

awesome/src/wayland_obj/xdg_shell.rs
groxxda: I think so too - removed.

Timidger:
Fix build and squash commits and should be good to go

Timidger:
Thanks

PR #609, opened 2019-04-27
Rewrite the Way Cooler compositor in C

Timidger:
The compositor part of Way Cooler is now written in C. The client portion (i.e. the side that implements the AwesomeWM functionality) is still written in Rust.

Ultimately, wlroots-rs was too difficult to write. The mental overhead of attempting to wrap complicated C libraries with Rust is too demanding. This complexity often leads to a RiiR mindset, which I am strongly against. So, the compositor is now written in C.

---

The functionality of the compositor is exactly the same as [tinywl](https://github.com/swaywm/wlroots/tree/master/tinywl), which for now is all that is necessary to work on the awesome portions. More will be added as needed.

The client has been fixed up so that it can now display again in Wayland compositors (including this one). It still isn't usable, obviously, however it being able to display the status bar is nice for debugging.

The shims have been disabled, as they were causing more harm than good.


compositor/meson.build
```diff
@@ -0,0 +1,57 @@
+cc = meson.get_compiler('c')
+
+wayland_server = dependency('wayland-server')
+wayland_client = dependency('wayland-client')
+wayland_cursor = dependency('wayland-cursor')
+wayland_egl    = dependency('wayland-egl')
+wayland_protos = dependency('wayland-protocols', version: '>=1.14')
+xkbcommon      = dependency('xkbcommon')
+xcb            = dependency('xcb', required: true)
```
groxxda:
```suggestion
xcb            = dependency('xcb')
```
required is the default, right?

compositor/meson.build
```diff
@@ -0,0 +1,57 @@
...
+    'cursor.c',
+    'input.c',
+    'keyboard.c',
+    'main.c',
+	'output.c',
+	'pointer.c',
+	'seat.c',
+	'server.c',
+	'view.c',
+)
```
groxxda: mixed indent

meson.build
```diff
@@ -0,0 +1,39 @@
...
+
+        '-Wno-unused-parameter',
+        '-Wno-unused-result',
+        '-Wundef',
+        '-Wvla',
+    ],
+    language: 'c'
+)
+
+git            = find_program('git', native: true, required: false)
```
groxxda:
```suggestion
git = find_program('git', native: true, required: false)
```

compositor/meson.build
Timidger: New to meson, yes it is :)

Contributing.md
```diff
@@ -1,11 +1,89 @@
...
+```
+
+### Import Groups
+
+Imports should be separated into 3 sections: standard library, 3rd party
+imports, and local lib/binary imports. Within these groups the imports should be
+sorted in alphabetical order. Each of these groups must be separated by a single
+empty line.
+
+Import statemnts should be grouped using the nested `{}` syntax.
```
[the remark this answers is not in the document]
Timidger: Sorting is required (or at least, rustfmt does it I believe)

compositor/meson.build
Timidger: ack!

nifker:
One step forward - And two steps back.

Elv13:
> One step forward - And two steps back.

Compared to "all steps back" because the project would be dead forever. So I take it.

lxbarbosa:
@nifker all we care is having an awesome wayland window manager.

PR #611, opened 2019-04-27
Update rlua -> 0.16

groxxda:
**Should be ready to merge.**

~~I left the individual commits that roughly correspond to the changed API.~~

~~They should be squashed before merge.~~

~~This introduces a regression in config.rs: I found no way to replace the existing LUA instance while it is borrowed.~~

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -15,7 +15,7 @@ lazy_static = "0.2"
 log = "0.4"
 env_logger = "0.5"
 exec = "0.3"
-rlua = { version = "0.12.0", default-features = false }
+rlua = { version = "0.16", default-features = false, features = ["system-lua"] }
 bitflags = "1.0"
 nix = "0.6"
 clap = "2"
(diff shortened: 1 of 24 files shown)


awesome/src/lua/config.rs
```diff
@@ -43,24 +43,39 @@ pub fn load_config(mut lua: &mut Lua, cmdline_config: Option<&str>, lib_paths: &
...
+                .set_name("init.lua")
+                .unwrap()
+                .exec()
                 .map(|_: ()| info!("Read init.lua successfully"))
                 .or_else(|err| {
                     log_error(err);
                     info!("Defaulting to pre-compiled init.lua");
                     unsafe {
-                        *lua = Lua::new_with_debug();
+                        // TODO: overwrite global lua!
```
Timidger: I'm guessing the way you will have to achieve this is that instead of passing in a `Context` here for `lua` you should pass in a `Lua`. Then you can replace the lua instance properly.

Note that the reason we replace the whole Lua instance here (in case you didn't already guess) is that this is the case where the init script was unsucessfully executed. However, the previous state from that execution is still hanging around and so we need to clear it out by making an entirely new instance.

Timidger:
Can you rebase on master and squash these commits? Should be good to go after that.

Timidger:
Excellent, thanks!

PR #612, opened 2019-04-30
Fix init file lookup

groxxda:
- The command line argument was ignored
- The path that was added to package.path was wrong:
For an init file in `/root/init.rc` it added `/root/init.rc/?.lua` to package.path.

### DIFF
--- a/client/src/lua/config.rs
+++ b/client/src/lua/config.rs
@@ -2,6 +2,7 @@
 
 use std::{
     env,
+    ffi::OsStr,
     fs::{File, OpenOptions},
     io::{self, Read},
     path::{Path, PathBuf},
@@ -15,10 +16,10 @@ const INIT_FILE_FALLBACK_PATH: &'static str = "/etc/way-cooler/rc.lua";
 
 pub const DEFAULT_CONFIG: &'static str = include_str!("../../../config/rc.lua");
 
-/// Finds the configuration file and returns (filename, content)
+/// Reads the init file and returns it's contents
 pub fn load_config(lua: Context, cmdline_config: Option<&str>) -> io::Result<(String, String)> {
-    let (init_dir, mut init_file) = get_config(cmdline_config)?;
-    if init_dir.components().next().is_some() {
+    let (init_path, mut init_file) = get_config(cmdline_config)?;
+    if let Some(init_dir) = init_path.parent() {
         // Add the config directory to the package path.
         let globals = lua.globals();
         let package: Table = globals.get("package").expect("package not defined in Lua");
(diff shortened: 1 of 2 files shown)


client/src/lua/config.rs
```diff
@@ -15,10 +16,10 @@ const INIT_FILE_FALLBACK_PATH: &'static str = "/etc/way-cooler/rc.lua";
 
 pub const DEFAULT_CONFIG: &'static str = include_str!("../../../config/rc.lua");
 
-/// Finds the configuration file and returns (filename, content)
+/// Reads the init file and returns it's contents
 pub fn load_config(lua: Context, cmdline_config: Option<&str>) -> io::Result<(String, String)> {
-    let (init_dir, mut init_file) = get_config(cmdline_config)?;
-    if init_dir.components().next().is_some() {
+    let (init_path, mut init_file) = get_config(cmdline_config)?;
+    if let Some(init_dir) = init_path.parent() {
```
groxxda: I could not figure out what the previous if-condition was doing here

Timidger: err...yeah I'm not sure what that was doing.

This looks more correct :+1:

client/src/lua/config.rs
```diff
@@ -28,13 +29,16 @@ pub fn load_config(lua: Context, cmdline_config: Option<&str>) -> io::Result<(St
...
             .expect("init_dir not a valid UTF-8 string");
+        trace!("Adding location of init file to package.path: {}", init_dir);
         package
             .set("path", format!("{};{}", paths, init_dir))
             .expect("Failed to set package.path");
     }
     let mut init_contents = String::new();
     init_file.read_to_string(&mut init_contents)?;
-    Ok(("init.lua".to_string(), init_contents))
+    let file_name = init_path.file_name().map(OsStr::to_str).unwrap_or(Option::None);
```
Timidger: Use `and_then` instead of `map(...).unwrap_or(Option::None)`.

[Option.and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then)

This is also called "flat map" in other languages (haskell? OCaml? At least one of those I think)

client/src/lua/config.rs
```diff
@@ -57,9 +64,9 @@ fn get_config(cmdline_path: Option<&str>) -> io::Result<(PathBuf, File)> {
         Some(INIT_FILE_FALLBACK_PATH.into())
     ];
 
-    paths[0] = env::var("WAY_COOLER_INIT_FILE").ok().map(PathBuf::from);
+    paths[1] = env::var("WAY_COOLER_INIT_FILE").ok().map(PathBuf::from);
```
Timidger: :scream: nice catch :+1:

client/src/lua/config.rs
groxxda: Thanks, I knew there had to be a function that does this but must have overlooked it..

Timidger:
Thanks!

Timidger:
Thanks!

PR #615, opened 2019-05-04
[WIP] Stubbed out awful.spawn

SethBarberee:
Most functions are dummy functions and copied awesome.exec to
awful.spawn.with_shell. Stil need to set up awful table to do
awful.spawn but for testing, do spawn.with_shell('command') in rc.lua
until that's set up.

### DIFF
--- a/client/src/awful/mod.rs
+++ b/client/src/awful/mod.rs
@@ -1,0 +1,1 @@
+pub mod spawn;
--- a/client/src/awful/spawn.rs
+++ b/client/src/awful/spawn.rs
@@ -1,0 +1,36 @@
+use std::{
+    process::{Command, Stdio},
+    thread
+};
+use rlua::{self};
+pub fn init(lua: rlua::Context) -> rlua::Result<()> {
+    // TODO Do properly
+    use crate::objects::dummy;
+
+    let spawn = lua.create_table()?;
+    spawn.set("with_shell", lua.create_function(with_shell)?)?;
+    spawn.set("with_line_callback", lua.create_function(dummy)?)?;
+    spawn.set("easy_async", lua.create_function(dummy)?)?;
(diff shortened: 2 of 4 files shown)


psychon:
`awful.spawn` uses `awesome.spawn` for the actual spawning. Last I checked (which is quite a while ago) that one was already stubbed.

So... am I missing something here or what is this actually adding?

Timidger:
As @psychon said, these functions are implemented in the Lua library (specifically [this file it seems](https://github.com/awesomeWM/awesome/blob/master/lib/awful/spawn.lua)) and the only thing that Way Cooler needs to implement are the calls to the `capi` interface functions. Apart from emitting and waiting on some signals all the work is done by the `spawn` function in `capi.awesome`.

However the `spawn` function is not implemented, so all that is needed to be done is implement it [here](https://github.com/way-cooler/way-cooler/blob/master/client/src/awesome.rs#L126) (as in write a function and then replace `crate::objects::dummy` with that function name like the other lines)

SethBarberee:
> So... am I missing something here or what is this actually adding?

Mainly my goal was to just stub out this so I can work on `awesome.spawn`. I'll mark this WIP then as I try to implement the `awesome.spawn` part then. 

From my naive approach currently, glib-rs hasn't wrapped `g_spawn_async_with_pipes` yet similar to what Awesome's C-API uses.

Timidger:
> From my naive approach currently, glib-rs hasn't wrapped g_spawn_async_with_pipes yet similar to what Awesome's C-API uses.

Either we should just use `unsafe` call directly then or send a patch to glib-rs to wrap it safely

SethBarberee:
Gonna close this as I haven't really had much time to delve deeper and with my internship about to start. I'll let someone else take a crack if they want.

PR #619, opened 2019-05-05
Rework startup, replace xdg_shell with layer-shell, introduce facade for wayland

groxxda:
This PR touches many places, I tried splitting it into smaller parts that are independent but that's too much work right now.

The main changes are:
### Rework startup
The startup previously relied on the initialization code for wayland and lua being in a specific order and interleaved. This added extra complications when the lua startup failed and was retried with the bundled init file.
This PR removes the interleaved parts. To accomplish this, lua is started before wayland. So the awesome code will not see any screens on initial startup. This does not work out of the box and will require changes to wibox to handle the headless case gracefully. Until then I added a simple fix to the init file that puts the affected components behind lazy initialization. 
I also moved the glib-loop from the lua module to the top level main module since it will soon have to interface with the wayland parts.
 
### Replace xdg_shell with layer-shell
layer-shell is the natural choice for the standard awesome widgets.
At the moment the implementation is very basic and does not support anything besides drawing the widget with the correct size (no Anchors, Struts or anything like that).
I removed the xdg_shell code for now. If it turns out it's needed for example for popups, it should be readded.

### Put wayland code behind a facade
While reworking the startup code, I tried my best to improve encapsulation of the wayland parts. This introduces a single facade object that delegates to the individual globals instead of each global being accessible individually.

### Open question
I have not decided what class should generate the signals that are emitted. Have a look at the drawable / drawin classes to see the possibilities: (DrawinState, Drawin or the adapter-functions) 

This also fixes the infinite loop my previous signal PR caused.

**Please do not merge as is. The trace outputs have to be removed before and maybe also the signal generation be streamlined**


client/src/area.rs
```diff
@@ -51,3 +55,48 @@ impl Into<Area> for Size {
         }
     }
 }
+
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Margin {
```
groxxda: layer-surface calls it margin, awesome calls it strut. I think margin is more natural.
If someone knows why the layer-surface protocol allows for negative values ( https://github.com/swaywm/wlr-protocols/blob/bc1364930c5f7a309b791403c05d0ca7489b4b07/unstable/wlr-layer-shell-unstable-v1.xml#L171-L174 ) this should be added as a comment here!

client/src/lua/mod.rs
```diff
@@ -85,16 +81,47 @@ fn load_shims(lua: rlua::Context) {
...
-///
-/// This environment is also necessary for some Wayland callbacks,
-/// so it should be called as soon as possible.
-pub fn init_awesome_libraries(lib_paths: &[&str]) {
-    info!("Setting up Awesome libraries");
-    LUA.with(|lua| {
-        lua.borrow()
-            .context(|ctx| register_libraries(ctx, lib_paths).expect("Could not register lua libraries"));
-    });
+pub struct WaylandHandler {}
```
groxxda: `WaylandCallbacks`?  `WaylandAdapter`? `WaylandEventHandlers`?

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+                error!(
+                    "WAYLAND_DISPLAY={}",
+                    env::var("WAYLAND_DISPLAY").unwrap_or_default()
+                );
+            },
+            InvalidName => error!("Invalid socket name provided in WAYLAND_SOCKET"),
+            XdgRuntimeDirNotSet => error!("XDG_RUNTIME_DIR must be set"),
+            InvalidFd => error!("Invalid socket provided in WAYLAND_SOCKET")
+        }
+        exit(1);
```
groxxda: Sorry for the `exit()` outside of `main.rs`.... This is unexpected and should be fixed. I think it's time for our own Result & Error types.... I should at least add a TODO here

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+    WAYLAND.with(|wayland| {
+        let wayland = &mut wayland.borrow_mut();
+        let way_man = WaylandManager::new(crate::lua::WaylandHandler {});
+        (*wayland).replace(way_man);
+    });
+
+    (display, event_queue)
+}
+
+impl WaylandManager {
```
groxxda: `WaylandManager`? `WaylandFacade`? `WaylandGlobalImplementor`?

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+            compositor_manager: WlCompositorManager::default(),
+            output_manager: WlOutputManager::new(handler),
+            seat_manager: WlSeatManager::new(),
+            shell_manager: LayerShellManager::new(),
+            shm_manager: WlShmManager::new()
+        }
+    }
+
+    fn new_global(&mut self, registry: WlRegistry, id: u32, interface: &str, version: u32) {
+        match interface {
```
groxxda: Originally I wanted a function like this:
```
fn get_handler(interface: String) -> (min_version: u32, implementor: &mut GlobalImplementor<Interface>) {
  match interface {
    WlCompositor::NAME => (WL_COMPOSITOR_VERSION, &mut self.compositor_manager),
    ....
  }
}
```
That could spare duplicate code when `GlobalEvent::Removed` gets implemented..
I have the impression that rust does not allow this, or did I miss something?

Timidger: This would return a pointer to a stack allocated object (since it was constructed in the body of the function and its drop will be called at the end of the stack frame). In particular this is a trait object (I believe some version of clippy should complain that this doesn't use `dyn`, which would make this more obvious...I should review how the CI is  set up). In order to return a trait object you need to box it explicitly, e.g return `Box<GlobalImplementor<Interface>>`. You could also use `impl GlobalImplementor<Interface>` IIRC instead.

client/src/wayland_obj/manager.rs
Timidger: What it is so far doing is managing the globals advertised by the Wayland compositor, so maybe `WaylandGlobalManager`?

client/src/wayland_obj/manager.rs
Timidger: Very much agree, for now you can just return a `Result<..., ()>` with a `// TODO` here.

I have had a good experience lately using [failure](https://rust-lang-nursery.github.io/failure/), so I was probably going to use that for errors.

client/src/lua/mod.rs
Timidger: FYI you can drop the `{}` and just put `pub struct WaylandHandler;`. To construct it you simply type `WaylandHandler`. Both should have the same memory representation IIRC (i.e. zero sized type).

I'd prefer `OutputHandler`, since that is what it seems like it is doing? Will we need to centralize _all_ of the handlers under one state structure? It seems better to keep them disjoint, unless you forsee a problem sharing data between them.

client/src/wayland_obj/output.rs
```diff
@@ -1,35 +1,52 @@
...
-pub const WL_OUTPUT_VERSION: u32 = 2;
+pub(crate) const WL_OUTPUT_VERSION: u32 = 2;
 
 /// Wrapper around WlOutput.
 #[derive(Clone, Eq, PartialEq)]
 pub struct Output {
     output: WlOutput
 }
 
+pub trait OutputEventHandler {
```
groxxda: Naming is a bit tricky here, `OutputEventHandler` is the trait that will be 'implemented' by the lua side, while `WlOutputEventHandler` is the implementation that is provided to the wayland side.... This should probably be documented somewhere

client/src/wayland_obj/output.rs
```diff
@@ -82,52 +91,31 @@ impl wl_output::EventHandler for WlOutputEventHandler {
...
 
 impl fmt::Debug for Output {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{:?}", self.output.as_ref().c_ptr())
+        write!(f, "{:?}", self.name())
+    }
+}
+
+impl AsRef<WlOutput> for Output {
+    fn as_ref(&self) -> &WlOutput {
```
groxxda: General rust question: Use `#[inline]` or not?

client/src/wayland_obj/wl_seat.rs
```diff
@@ -0,0 +1,63 @@
...
+        let seat = new_seat.implement(SeatHandler {}, Rc::new(RefCell::new(SeatData::new())));
+        if self.seat.replace(seat.clone()).is_some() {
+            panic!("Seat already registered. Multiple seats are not supported");
+        }
+
+        seat
+    }
+}
+
+struct SeatHandler {}
```
groxxda: todo: be consistent with these class names and rename to `WlSeatEventHandler`

client/src/wayland_obj/output.rs
Timidger: Nah, trust the optimizer to do it (or not). There's too much voodoo with `#[inline]`. If this really is a problem then we would first:

1) Profile, and ensure that a callstack really is too much overhead (doubtful)
2) Enable LTO, which would probably enable automatic inlining
3) Add `#[inline]`, _not_ `#[inline(always)]` which could mess with debugging IIRC.

Generally though, don't worry about it.

client/src/wayland_obj/manager.rs
groxxda: I believe the problem was that `Interface` is also a trait.

client/src/wayland_obj/wl_shm.rs
```diff
@@ -1,59 +1,91 @@
...
+        self.shared_memory
+            .write(&*data)
+            .expect("Could not write data to buffer");
+        self.shared_memory.flush().expect("Could not flush buffer");
+    }
 }
 
-pub struct WlShmManager {}
+impl AsRef<WlBuffer> for Buffer {
+    #[inline]
```
groxxda: todo: don't inline as recommended in the other comment

client/src/lua/mod.rs
groxxda: I don't think that the handlers must share the same instance, let's keep them separate until the need arises (if ever)

client/src/area.rs
Timidger: According to @emersion on #sway-devel "the original idea was to allow slide-out animations"

client/src/objects/drawin.rs
```diff
@@ -4,31 +4,93 @@
...
 
 pub const DRAWINS_HANDLE: &'static str = "__drawins";
 
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub struct DrawinState {
-    // Note that the drawable is stored in Lua.
-    // TODO WINDOW_OBJECT_HEADER??
-    ontop: bool,
+    // ontop: bool,
```
Timidger: Remove commented out code

client/src/objects/drawin.rs
```diff
@@ -4,31 +4,93 @@
...
+}
+// buttons : Vec<Button>,
+// border_needs_update : bool,
+// border_color : Color,
+// border_width : u16,
+// type : WindowType,
+// border_width_callback : Fn(_ , old_width : u16, new_width : u16)
+
+impl DrawinState {
+    /// Update the geometry of this drawin and return the signals to be emitted.
```
Timidger: Is there a reason we send the signals at the end all at once instead of just doing it one by one?

client/src/objects/drawin.rs
```diff
@@ -80,54 +182,86 @@ impl<'lua> Drawin<'lua> {
...
+    fn set_struts(&mut self, lua: rlua::Context<'lua>, struts: Margin) -> rlua::Result<()> {
+        trace!("set_struts({:?})", struts);
+        if struts == self.state()?.struts {
+            return Ok(());
+        }
+
+        self.state_mut()?.struts = struts;
+
+        // TODO(ried): emit screen change workarea event?
+        // TODO(ried): remove since awesome does not emit this event?
```
Timidger: If awesome doesn't emit the event then neither should we. Later once we are compatible we can see about potentially fixing these mistakes

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,169 @@
...
+                trace!("Attaching buffer ({:p}) @ ({},{})", &buffer, x, y);
+                self.wl_surface.attach(Some(&buffer), x, y);
+            }
+        }
+    }
+}
+
+struct LayerSurfaceEventHandler {}
+
+#[allow(unused_variables)]
```
Timidger: Potentially unneeded, if it is use `_`

client/src/wayland_obj/wl_seat.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn new() -> Self {
+        WlSeatManager { seat: None }
+    }
+}
+
+impl GlobalImplementor<WlSeat> for WlSeatManager {
+    fn new_global(&mut self, new_seat: NewProxy<WlSeat>) -> WlSeat {
+        let seat = new_seat.implement(SeatHandler {}, Rc::new(RefCell::new(SeatData::new())));
+        if self.seat.replace(seat.clone()).is_some() {
+            panic!("Seat already registered. Multiple seats are not supported");
```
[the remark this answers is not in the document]
groxxda: I should have paid more attention to the protocol. Looks like I assumed that seat was a global singleton like compositor, shm., etc. Good catch!

client/src/objects/drawin.rs
groxxda: I will remove the update-logic from the ..State objects, then the signal can be emitted one-by-one.

client/src/wayland_obj/wl_seat.rs
```diff
@@ -0,0 +1,63 @@
+use std::{cell::RefCell, rc::Rc};
+use wayland_client::{protocol::wl_seat as protocol, GlobalImplementor, NewProxy, Proxy};
+
+use protocol::{Capability, WlSeat};
+
+/// The minimum version of the wl_seat global to bind to.
+pub(crate) const WL_SEAT_VERSION: u32 = 1;
+
+#[derive(Debug)]
+pub(crate) struct SeatData {
```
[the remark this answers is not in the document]
groxxda: I reduced the visibilities in the wayland_obj module, let me know if they are still too permissive

client/src/wayland_obj/wl_shm.rs
```diff
@@ -1,59 +1,90 @@
...
+        wl_shm::{Format, WlShm}
     },
     NewProxy
 };
 
 use crate::area::Size;
 
-/// The minimum version of the wl_shm global to bind to.
-pub const WL_SHM_VERSION: u32 = 1;
+// TODO(ried): the WlBuffer internal should be internal and Buffer should be Clone
```
groxxda: This would require cloning the File handle. std::fs::File provides a try_clone method, any concerns with this approach?

Timidger: Why does this need to be clone? If so, then does it make sense that both buffers refer to the same file/contents, even though they are "different" buffers?

I think instead that a new temp file should be created on `clone`. However, if we don't need this to be clone it should not be because it's going to be expensive to do so.

If whatever needs to clone this is instead trying to get a reference to this buffer either borrows or `Rc` should be used.

client/src/main.rs
```diff
@@ -229,86 +241,35 @@ fn main() {
...
-            ConnectError::XdgRuntimeDirNotSet => error!("XDG_RUNTIME_DIR must be set"),
-            ConnectError::InvalidFd => error!("Invalid socket provided in WAYLAND_SOCKET")
-        }
-        exit(1);
+    let (display, _globals) = run_wayland(display, &mut event_queue);
+
+    init_glib(display, event_queue, session_fd, system_fd);
+
+    // TODO(ried): hold off refresh until at least one screen is available
+    // (awesome does not like running headless)
```
Timidger: If you could detail what exactly is necessary to make awesome run headless a patch can be constructed that can be sent to them to fix this issue.

Making it in the AwesomeWM bug tracker probably makes the most sense, as then we can easily get feedback before too much work is attempted (e.g. in case they are adverse to such a patch, I don't know why they would, or if they have some insight)

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+
+struct WaylandManager {
+    compositor_manager: WlCompositorManager,
+    output_manager: WlOutputManager,
+    seat_manager: WlSeatManager,
+    shell_manager: LayerShellManager,
+    shm_manager: WlShmManager
+}
+
+// pub fn create_surface() -> Result<WlSurface, ()> {
```
Timidger: This should be removed if it's not necessary

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+    WAYLAND.with(|wayland| {
+        let wayland = &mut wayland.borrow_mut();
+        let wayland = wayland.as_mut().expect("Wayland not initialized");
+        match event {
+            GlobalEvent::New {
+                id,
+                interface,
+                version
+            } => wayland.new_global(registry, id, &interface, version),
+            _ => unimplemented!("GlobalEvent")
```
Timidger: At the very least log this. I would not panic, as I don't see why we need to.

client/src/objects/drawable.rs
```diff
@@ -58,6 +56,7 @@ impl<'lua> Drawable<'lua> {
 
     pub fn get_surface(&self) -> rlua::Result<Value<'lua>> {
         let drawable = self.state()?;
+        trace!("get_surface -> {}", drawable.surface.is_some());
```
Timidger: This log is unnecessary.

(Btw you might find the `dbg!` macro helpful in the future for debugging, it also makes it more clear that "this is logging that shouldn't actually be in the final solution")

client/src/objects/drawable.rs
```diff
@@ -79,55 +78,60 @@ impl<'lua> Drawable<'lua> {
...
-                        .map_err(|_| RuntimeError(format!("Could not set surface for drawable")))?;
-                }
-                drawable.temp_file = temp_file;
-                Object::emit_signal(lua, &obj_clone, "property::surface".into(), Value::Nil)?;
-            }
+
+            drawable.buffer = Some(create_buffer(size).expect("Could not create buffer"));
         }
+        drop(drawable);
+        // TODO(ried): conditionally emit signals only for changed properties
```
Timidger: Can you do that in this patch set? It should be a pretty simple comparison logic flow.

client/src/objects/drawable.rs
```diff
@@ -79,55 +78,60 @@ impl<'lua> Drawable<'lua> {
...
-        let mut drawable = self.state_mut()?;
-        let drawable = &mut *drawable;
-        if let Some(data) = drawable.surface.as_mut().map(get_data) {
-            drawable
-                .temp_file
-                .write(&*data)
-                .expect("Could not write data to buffer");
-            drawable.temp_file.flush().expect("Could not flush buffer");
+    pub fn refresh(&mut self, _lua: Context) -> rlua::Result<()> {
+        trace!("drawable::refresh");
```
Timidger: unnecessary log

client/src/objects/drawable.rs
```diff
@@ -79,55 +78,60 @@ impl<'lua> Drawable<'lua> {
...
+        if self.state()?.refreshed {
+            warn!("Drawable is already refreshed. Skipping work");
```
Timidger: unnecessary log

client/src/objects/drawable.rs
```diff
@@ -79,55 +78,60 @@ impl<'lua> Drawable<'lua> {
...
+            if let Some(data) = drawable.surface.as_ref().map(get_data) {
+                drawable.buffer.as_mut().expect("No buffer available").write(data);
+            }
             drawable.refreshed = true;
         }
-        Ok(())
+
+        let geo = self.state()?.geo;
+
+        error!("calling callback");
```
Timidger: unnecessary log

client/src/objects/drawable.rs
```diff
@@ -79,55 +78,60 @@ impl<'lua> Drawable<'lua> {
...
+        let res = self
+            .get_associated_data::<Drawin>("drawin")?
+            .refresh_pixmap(self.state()?.buffer.as_ref().unwrap(), geo);
+
+        info!("callback done");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -43,31 +96,71 @@ impl<'lua> Drawin<'lua> {
...
+        drawin.create_shell()?;
+        drawin
+            .drawable()?
+            .set_associated_data::<Drawin>("drawin", drawin.clone())?;
         drawins.push(drawin.clone());
         lua.set_named_registry_value(DRAWINS_HANDLE, drawins.to_lua(lua)?)?;
         Ok(drawin)
     }
 
+    pub(crate) fn refresh_pixmap(&mut self, buffer: &Buffer, geometry: Area) -> rlua::Result<()> {
```
Timidger: `pub`

client/src/objects/drawin.rs
```diff
@@ -43,31 +96,71 @@ impl<'lua> Drawin<'lua> {
...
+            let mut state = self.state_mut()?;
+            let shell = state.surface.as_mut().unwrap();
+            shell.set_size(geometry.size);
+
+            shell.set_buffer(buffer, geometry.origin);
+
+            shell.commit();
+        }
+
+        error!("refresh_pixmap done :)");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -43,31 +96,71 @@ impl<'lua> Drawin<'lua> {
...
         Ok(drawin)
     }
 
+    pub(crate) fn refresh_pixmap(&mut self, buffer: &Buffer, geometry: Area) -> rlua::Result<()> {
+        {
+            let mut state = self.state_mut()?;
+            let shell = state.surface.as_mut().unwrap();
+            shell.set_size(geometry.size);
+
+            shell.set_buffer(buffer, geometry.origin);
```
Timidger: Should probably only set the buffer and commit if the size has changed, no? Or is this supposed to handle all updates to the drawin? AFAIK the `refreshed` boolean already covers that though.

client/src/objects/drawin.rs
```diff
@@ -43,31 +96,71 @@ impl<'lua> Drawin<'lua> {
...
     ///
     /// It has the surface that is needed to render to the screen.
-    #[allow(dead_code)]
     pub fn drawable(&mut self) -> rlua::Result<Drawable<'lua>> {
         self.get_associated_data::<Drawable>("drawable")
     }
 
+    fn create_shell(&mut self) -> rlua::Result<()> {
+        let mut state = self.state_mut()?;
+        if state.surface.is_some() {
```
Timidger: If this should only be called in `new` then just put this code in `new`

client/src/objects/drawin.rs
```diff
@@ -43,31 +96,71 @@ impl<'lua> Drawin<'lua> {
...
+
     fn update_drawing(&mut self, lua: rlua::Context<'lua>) -> rlua::Result<()> {
-        let mut drawable: Drawable = self.get_associated_data("drawable")?;
-        {
-            let mut state = self.state_mut()?;
-            if state.geometry_dirty {
-                drawable.set_geometry(lua, state.geometry)?;
-                state.geometry_dirty = false;
-            }
+        trace!("drawin::update_drawing");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
     }
 
     fn set_visible(&mut self, lua: rlua::Context<'lua>, val: bool) -> rlua::Result<()> {
-        {
-            let mut drawin = self.state_mut()?;
-            drawin.visible = val;
+        let mut state = self.state_mut()?;
+        trace!("set_visible(): {}, old: {}", val, state.visible);
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
...
+
         if val {
-            self.map(lua)
+            self.map(lua)?
         } else {
-            self.unmap()
+            self.unmap(lua)?
         }
+
+        trace!("will signal property::visible");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
...
         // TODO other things
+        trace!("drawin::map");
         self.update_drawing(lua)?;
         Ok(())
     }
 
-    fn unmap(&mut self) -> rlua::Result<()> {
+    fn unmap(&mut self, lua: rlua::Context<'lua>) -> rlua::Result<()> {
         // TODO?
+        trace!("drawin::unmap");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
...
     }
 
     pub fn get_geometry(&self) -> rlua::Result<Area> {
         Ok(self.state()?.geometry)
     }
 
+    /// Move and/or resize a drawin
     fn resize(&mut self, lua: rlua::Context<'lua>, geometry: Area) -> rlua::Result<()> {
-        {
+        trace!("drawin::resize");
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
...
+            state.update_geometry(geometry)
+        };
+
+        if signals.len() > 0 {
+            self.update_drawing(lua)?;
+        }
+
+        // TODO update screen workareas like in awesome? Might not be necessary
+        for (signal, args) in signals {
+            trace!("drawin::resize: signaling {}", signal);
```
Timidger: unnecessary log

client/src/objects/drawin.rs
```diff
@@ -78,54 +171,87 @@ impl<'lua> Drawin<'lua> {
...
+            Object::emit_signal(lua, self, &signal, args)?;
         }
-        self.update_drawing(lua)
+
+        Ok(())
+    }
+
+    fn set_struts(&mut self, _lua: rlua::Context<'lua>, struts: Margin) -> rlua::Result<()> {
+        trace!("set_struts({:?})", struts);
```
Timidger: unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
+use std::{cell::RefCell, rc::Rc};
+
+use wayland_client::{protocol::wl_buffer::WlBuffer, GlobalImplementor, NewProxy, Proxy};
+
+pub(super) use wayland_protocols::wlr::unstable::layer_shell::v1::client::{
```
Timidger: i dunno about the rest of these, but this can just be `use`.

`pub(super)` makes whatever is on the right usable by the parent module. Generally you should just be able to use `pub`, the amount of times you will need `pub(super)` seems fairly rare.

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    pub fn create_layer_surface(
+        &self,
+        wl_surface: Surface,
+        output: Option<&Output>,
+        layer: Layer
+    ) -> Result<LayerSurface, ()> {
+        use wayland_client::Interface; // for ::NAME
+        trace!("create_layer_surface");
```
Timidger: unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+        LayerShellManager { shell: None }
+    }
+
+    pub fn create_layer_surface(
+        &self,
+        wl_surface: Surface,
+        output: Option<&Output>,
+        layer: Layer
+    ) -> Result<LayerSurface, ()> {
+        use wayland_client::Interface; // for ::NAME
```
Timidger: Move `use` to top of file

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+        let shell = self.shell.as_ref().expect(&format!(
+            "No LayerShell registered. Make sure your compositor supports the {} protocol",
+            WlrLayerSurface::NAME
+        ));
+
+        let layer_surface = shell.get_layer_surface(
+            &wl_surface,
+            output.map(AsRef::as_ref),
+            layer,
+            Default::default(),
```
Timidger: Please prefer `StructName::default`, as it makes it clearer what is actually being defaulted.

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+            |new_layer_surface| {
+                let state = Rc::new(RefCell::new(LayerSurfaceState::new(wl_surface.clone())));
+                new_layer_surface.implement(LayerSurfaceEventHandler {}, state)
+            }
+        )?;
+
+        let res = LayerSurface::new(layer_surface);
+
+        trace!("committing layersurface");
```
Timidger: Unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+        let mut state = unwrap_state(self.as_ref()).borrow_mut();
+        if let Some((_old_buffer, _)) = state.buffer.replace((buffer.buffer.clone(), origin)) {
+            // TODO(ried): cleanup?
+            warn!("Replacing uncommitted buffer, cleanup?");
+        }
+        // Assigning the Buffer before the first configure event is a protocol error.
+        if state.configuration_serial.is_some() {
+            state.attach_buffer();
+        } else {
+            info!("LayerSurface is not yet configured. Postponing layer attachment");
```
Timidger: This should probably be bumped up to an `error!`, as otherwise this could be lost in the noise.

Also this should probably be represented as an error on the Lua side too, so it should eventually bubble up to an `rlua::Error` (potentially a different error could be returned here, but ultimately to communicate to Lua it must be an `rlua::Error`)

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    pub fn set_size(&mut self, Size { width, height }: Size) {
+        self.layer_surface.set_size(width, height);
+    }
+
+    pub fn commit(&self) {
+        trace!("layersurface commit");
```
Timidger: Unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+            wl_surface,
+            buffer: None,
+            configuration_serial: None
+        }
+    }
+    fn attach_buffer(&mut self) {
+        match self.buffer.take() {
+            None => panic!("Cannot attach buffer to LayerSurface: no buffer available"),
+            Some((buffer, Origin { x, y })) => {
+                trace!("Attaching buffer ({:p}) @ ({},{})", &buffer, x, y);
```
Timidger: Unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+    }
+}
+
+struct LayerSurfaceEventHandler {}
+
+impl wlr_layer_surface::EventHandler for LayerSurfaceEventHandler {
+    fn configure(&mut self, layer_surface: WlrLayerSurface, serial: u32, width: u32, height: u32) {
+        info!("layer_surface::configure {}x{}", width, height);
```
Timidger: Unnecessary log

client/src/wayland_obj/layer_shell.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        let mut state = unwrap_state(layer_surface.as_ref()).borrow_mut();
+        state.configuration_serial = Some(serial);
+        if state.buffer.is_some() {
+            state.attach_buffer();
+            state.wl_surface.commit();
+        }
+    }
+    fn closed(&mut self, _layer_surface: WlrLayerSurface) {
+        unimplemented!();
```
Timidger: This should probably be implemented. I doubt it has to do much (if anything)

client/src/wayland_obj/manager.rs
Timidger: Hm, that shouldn't matter. It should still be possible to box the outer trait.

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+                Self::handle::<WlShm>(registry, WL_SHM_VERSION, version, id, &mut self.shm_manager)
+            },
+            WlrLayerShell::NAME => Self::handle::<WlrLayerShell>(
+                registry,
+                WLR_LAYER_SHELL_VERSION,
+                version,
+                id,
+                &mut self.shell_manager
+            ),
+            _ => info!("unhandled global: {}", interface)
```
Timidger: I don't think this is worth logging. For debugging, it's already possible to see all globals using `weston-info`

client/src/wayland_obj/manager.rs
```diff
@@ -0,0 +1,168 @@
...
+        min_version: u32,
+        version: u32,
+        id: u32,
+        implementor: &mut GlobalImplementor<I>
+    ) {
+        if version < min_version {
+            implementor.error(version);
+        } else {
+            registry
+                .bind::<I, _>(version, id, |newp| implementor.new_global(newp))
```
Timidger: This can be `implementor.new_global` instead of a closure

client/src/wayland_obj/output.rs
```diff
@@ -1,35 +1,52 @@
...
+
+impl OutputEventHandler for Fn(Output) {
+    fn output_changed(&self, output: Output) {
+        (*self)(output);
+    }
+}
+
 // Provides new WlOutputs with an implementation.
-pub struct WlOutputManager {}
+pub(super) struct WlOutputManager {
```
Timidger: `pub`

client/src/wayland_obj/output.rs
Timidger: Why does this trait need to exist in the first place?

client/src/wayland_obj/output.rs
```diff
@@ -1,35 +1,52 @@
...
-pub struct WlOutputEventHandler {}
+struct WlOutputEventHandler {
+    handler: Rc<dyn OutputEventHandler>
+}
 
 /// The cached state for the WlOutput.
 #[derive(Debug, Default, Clone, Eq, PartialEq)]
 struct OutputState {
     name: String,
+    // TODO(ried): use area::Size
```
Timidger: :+1:

client/src/wayland_obj/wl_compositor.rs
```diff
@@ -1,37 +1,43 @@
...
-    static WL_COMPOSITOR: RefCell<Option<WlCompositor>> = RefCell::new(None);
+#[derive(Default)]
+pub(super) struct WlCompositorManager {
+    compositor: Option<WlCompositor>
 }
 
-pub struct WlCompositorManager {}
+impl WlCompositorManager {
+    pub fn create_surface(&self) -> Result<Surface, ()> {
+        use wayland_client::Interface; // for ::NAME
```
Timidger: Put import at top

client/src/wayland_obj/wl_compositor.rs
```diff
@@ -1,37 +1,43 @@
...
-use std::cell::RefCell;
-
 use wayland_client::{
     protocol::{wl_compositor::WlCompositor, wl_surface::WlSurface},
     GlobalImplementor, NewProxy
 };
 
 /// The minimum version of the wl_compositor global to bind to.
-pub const WL_COMPOSITOR_VERSION: u32 = 3;
+pub(super) const WL_COMPOSITOR_VERSION: u32 = 3;
```
Timidger: `pub`

client/src/wayland_obj/wl_compositor.rs
```diff
@@ -1,37 +1,43 @@
...
     fn new_global(&mut self, new_proxy: NewProxy<WlCompositor>) -> WlCompositor {
         let res = new_proxy.implement_dummy();
 
-        WL_COMPOSITOR.with(|wl_compositor| {
-            *wl_compositor.borrow_mut() = Some(res.clone());
-        });
+        if self.compositor.replace(res.clone()).is_some() {
+            use wayland_client::Interface; // for ::NAME
+            panic!(
+                "{} already registered. Multiple compositors are not supported",
```
Timidger: This should honestly never happen, and we shouldn't bother dealing with this. Just replace it always and don't mind what was there before.

Timidger:
Ping? If you don't have time to finish this up I can, it's nearly there and I have enough time to push it over the finish line if need be.

groxxda:
Please go ahead and finish this if you have spare time. I might get to it this weekend but I can not guarantee I'll find the time.

Timidger:
Ok, thank you for your work so far it is very appreciated! You did the hard part :)

PR #623, opened 2019-06-17
Fix incorrect formatting of pointer declarations

ddevault:


Timidger:
Build is broken for different reasons.

If I continue the project I'm just going to move to a formatter because I generally don't care about these stylistic opinions, but I've been informed my way is non-standard and I'd rather conform than be contrarian.

PR #631, opened 2019-07-14
Wlroot 0.6

SethBarberee:
Fixes #631 
- [x] Update frame submission 
- [x] Update build
- [x] Remove lx and ly as wlr-output no longer provides it. <-- **iffy on this one, please check**

### DIFF
--- a/compositor/view.c
+++ b/compositor/view.c
@@ -93,10 +93,11 @@ void wc_view_damage(struct wc_view *view, pixman_region32_t *damage) {
 
 	for (size_t i = 0; i < 4; i++) {
 		struct wlr_output *output = outputs[i];
-                struct wlr_box *output_box = wlr_output_layout_get_box(view->server->output_layout, output);
+		struct wlr_box *output_box =
+				wlr_output_layout_get_box(view->server->output_layout, output);
 
 		if (output) {
-                        // TODO replace lx and ly with xdg-output
+			// TODO replace lx and ly with xdg-output
 			struct wlr_box view_output_geo = {.x = view->geo.x - output_box->x,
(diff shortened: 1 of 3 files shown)


Timidger:
Looks good, I will need to test this later today when I have access to a multi-monitor setup (to ensure the output layout change affects anything).

compositor/view.c
```diff
@@ -93,9 +93,13 @@ void wc_view_damage(struct wc_view *view, pixman_region32_t *damage) {
 
 	for (size_t i = 0; i < 4; i++) {
 		struct wlr_output *output = outputs[i];
+		struct wlr_box *output_box =
+				wlr_output_layout_get_box(view->server->output_layout, output);
+
 		if (output) {
-			struct wlr_box view_output_geo = {.x = view->geo.x - output->lx,
-					.y = view->geo.y - output->ly,
+			// TODO replace lx and ly with xdg-output
```
Timidger: Ok tested it out, and not only does it work but these values aren't necessary here at all. Can you just remove the box access altogether then?

compositor/view.c
```diff
@@ -93,9 +93,13 @@ void wc_view_damage(struct wc_view *view, pixman_region32_t *damage) {
...
+			struct wlr_box view_output_geo = {.x = view->geo.x - output_box->x,
+					.y = view->geo.y - output_box->y,
 					.width = view->geo.width,
 					.height = view->geo.height};
```
Timidger: Also while you're at it, can you add a trailing comma here after `view->geo.height`? clang-format has a bug (at least I hope it's a bug) where it does ugly formatting of this inline struct definition unless you add a trailing comma.

PR #633, opened 2019-07-17
Added mousegrabber api

Timidger:
This adds the mousegrabber api support. It should do _exactly_ what it does in AwesomeWM - e.g. being able to arbitrary set the cursor and not send input to other clients.

# TODO
- [x] ~There's a use double free causing a segfault~ allocating a word size not the size of the struct
- [x] Double check the button is getting to Lua correctly.

### DIFF
--- a/client/build.rs
+++ b/client/build.rs
@@ -3,9 +3,12 @@ extern crate pkg_config;
 
 use std::{env, fs, io::Write, path::Path, process::Command};
 
+use wayland_scanner::{generate_code, Side};
+
 fn main() {
     dump_git_version();
     build_wayland_glib_interface();
+    generate_custom_protocols();
 }
 
 /// Writes the current git hash to a file that is read by Way Cooler
@@ -72,4 +75,15 @@ fn build_wayland_glib_interface() {
         .file("src/wayland_glib_interface.c")
         .compile("wayland_glib_interface");
     println!("cargo:rustc-flags=-l wayland-client");
+}
+
+fn generate_custom_protocols() {
+    let out_dir_str = env::var("OUT_DIR").unwrap();
+    let out_dir = Path::new(&out_dir_str);
+
+    let protocol = "../protocols/way-cooler-mousegrabber-unstable-v1.xml";
+
+    generate_code(protocol, out_dir.join("mouse_grabber_api.rs"), Side::Client);
+
+    println!("cargo:rerun-if-changed={}", protocol);
 }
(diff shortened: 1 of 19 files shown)


client/build.rs
```diff
@@ -73,3 +76,14 @@ fn build_wayland_glib_interface() {
         .compile("wayland_glib_interface");
     println!("cargo:rustc-flags=-l wayland-client");
 }
+
+fn generate_custom_protocols() {
+    let out_dir_str = env::var("OUT_DIR").unwrap();
+    let out_dir = Path::new(&out_dir_str);
+
+    let protocol = "../protocols/way-cooler-mousegrabber-unstable-v1.xml";
```
Timidger: This should just grab all the protocols in the project and build them

client/src/main.rs
```diff
@@ -316,6 +316,30 @@ fn init_wayland() -> (WaylandState, GlobalManager) {
             ));
         }
     });
+    globals.instantiate_exact(
```
Timidger: This is getting repeaty, you should add a macro or something to clean this up (or can we do it with a function?)

config/rc.lua
```diff
@@ -588,3 +588,12 @@ end)
...
 -- }}}
+
+mousegrabber.run(
+  function(mouse, button)
+    print(mouse["x"], button)
+    if (mouse["x"] < 100) then
+      return false
+    end
+    return true
+  end, "cross")
```
Elv13: You might want to avoid having this by default? Maybe it is time to add an `awesome-x11` like `tests/` directory with a headless wayland server? Assuming the GLib event loop can be started without deadlocking eveything, GDbus should work? This is "all" it takes to inject the tests.

Timidger: Yeah that was for testing that I forgot about, I'll remove that.

Timidger: What I was testing with this was somewhat difficult to test.

There's not really a way to ensure that when this executes the following happens:
1) Any current mouse grabs (e.g. resizing a window) stops
2) Input is no longer sent to any other client but `way-cooler-client`
3) Keyboard input still works however and new clients take focus
4) The cursor does not change (this is not how Wayland works "by default" - normally clients expose their own cursor and so we have to actively ignore it while in this mode)

The tests will be useful in other cases. E.g setting up fake outputs or inputs and ensuring `way-cooler-client` gets it. But it will take a lot of machinery that is tangential to getting this actually working right now. It's not useful to build regression tests before it even works.

I'm not against tests at all, it's just a lot of work and right now I already have a lot on my plate considering I'm the only contributor. Patches welcome.

Issue #634, opened 2019-07-19
Doesn't build the client, build.rs problem


EleDiaz:
Hi, I was trying to build the project on Archlinux, but I wasn't able, because how build.rs works generating **mouse_grabber_api.rs**.

It depends on how **fs::read_dir** list directories, and in my case it gets overwrite by wlr protocol

https://github.com/way-cooler/way-cooler/blob/d32cae33755a6923f83a4393f7756424dbfc9bc7/client/build.rs#L82

I guess the solution, it's something like this
```rust
fn generate_custom_protocols() {
    let out_dir_str = env::var("OUT_DIR").unwrap();
    let out_dir = Path::new(&out_dir_str);
    let path = Path::new(PROTOCOL_DIR).join("way-cooler-mousegrabber-unstable-v1.xml");
    generate_code(path, out_dir.join("mouse_grabber_api.rs"), Side::Client);
    println!("cargo:rerun-if-changed={}", PROTOCOL_DIR);
}
```

PR #635, opened 2019-07-19
A possible fix for #634

EleDiaz:


Timidger:
Derp, that was some dumb code I wrote, apologies.

Can you instead change it so that it uses the file name as part of the generation? More protocols will be added and I was trying (poorly) to future proof this so I can just add a protocol and it will pick it up and generate it.

I think [this function](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name) is what we want.

Timidger:
Thanks!

PR #637, opened 2019-07-28
Topbar work

Timidger:


Timidger:
Great, thanks for checking. Will need to fix that CI issues with rustfmt being installed..

Was hoping to do more with this patch set but it seems more and more likely that dumbly implementing the API is a better way to fix things. For example as it stand now if the output sizes changes a million errors are thrown from something unrelated. Everything is so interconnected it's probably best to hook everything up (mocked or otherwise) and then fix any problems that arise.

Bit of a shame, as smaller piecemeal success is more encouraging. Eg. getting the bar to sorta work. Though there are a few use cases that should be easy to set up (Sending all keyboard input to this client and using that to spawn programs, for example)

Elv13:
args, too late :-(

Timidger:
rip

I'm just going to target 4.3 (or whatever the highest stable version is at any time). I'm fine with accepting patches for those future changes though, as long they aren't backwards incompatible (since right now I just use the system installed Awesome libraries and I'd rather not make it harder to contribute by forcing the use of a bleeding-edge version)

Elv13:
> as long they aren't backwards incompatible

Well, if I change the name of the CAPI method then add legacy support code in `awful` to "fix" existing user configs, then that is mildly backward/forward incompatible if way-cooler doesn't provide the renamed methods.

The change is a cleaner version of this ugly hack:

```lua

-- was called `:buttons()`

--TODO v6 remove this legacy ****
function client.get_buttons(self)
    return setmetatable(self:_buttons(), {__call = function(self, new_buttons)
        gears.debug.deprecated([[
            buttons is now a property, please stop calling it, it is not a method anymore,
            use `c.buttons = new_buttons` instead of `c:buttons(new_buttons)`
        ]])

        if new_buttons then self:_buttons(new_buttons) end

        return self:_buttons()
    end}
end

--TODO v6 remove this legacy ****
function client.set_buttons(self, new_buttons)
     self:_buttons(new_buttons)
end

```

From the end user point of view, it is fully backward compatible and they wont notice the change until the deprecation warning appears (after v5.0 release). But in any case, if someone uses Awesome libs from v4.4 on a way-cooler without the name change, it will break. The same is true with using the libs from v4.3 on a v4.4 core.

I mean, the public API doesn't break, but the private APIs mildly change in every release (and always has). At some point way-cooler will have to track AwesomeWM git-master Lua libs if we want to do a shared release (AwesomeWM 5.0 / Way-Cooler 1.0?) with Wayland support using Way-Cooler. I don't mind submitting patches on both side once it gets complete-ish enough so I can at least test them. Right now I am announcing my plan to move a lot of the existing core methods away from the public API and into the private one so I can fix some API bugs/inconsistencies in a (Lua) compatibility layer. Someone else is currently working on a framework to enforce the (Lua) API naming guidelines so this never regresses. The end result is hopefully a more consistent API with a lesser learning curve and a better documentation. This is a net win for way-cooler at the cost of having to coordinate better with "downstream" (ie, the AwesomeWM Lua libraries) developers (ie, me).

(as I side note, I started doing those mousegrabber tests I proposed in last week PR, but I am not done yet)

PR #640, opened 2019-08-07
Execute command after X11 starts up

Timidger:
This is done so that the client sees the updated DISPLAY environment
variable.

This is especially important to Awesome because it still uses legacy X11
commands for many of its functionality.

### DIFF
--- a/compositor/xwayland.c
+++ b/compositor/xwayland.c
@@ -2,6 +2,7 @@
 #include "xwayland.h"
 
 #include <stdlib.h>
+#include <unistd.h>
 
 #include <wlr/util/log.h>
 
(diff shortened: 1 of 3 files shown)


compositor/xwayland.c
```diff
@@ -138,13 +139,28 @@ static void wc_xwayland_new_surface(struct wl_listener *listener, void *data) {
...
 }
 
+static void wc_xwayland_ready(struct wl_listener *listener, void *data) {
+	struct wc_server *server =
+			wl_container_of(listener, server, xwayland_ready);
+	if (server->startup_cmd != NULL) {
+		// NOTE Executed here so that DISPLAY is correct for the client
+		wlr_log(WLR_INFO, "Executing \"%s\"", server->startup_cmd);
+		if (fork() == 0) {
+			execl("/bin/sh", "/bin/sh", "-c", server->startup_cmd, NULL);
```
Elv13: Some BSD people will eventually argue that `env` is better. I never saw a system where the POSIX shell isn't in /bin.

psychon: `execlp` would be a simpler way to do that (involves starting one process less).

PR #642, opened 2019-08-08
Fixed mousegrabber

Timidger:
Aborts were happening because of insufficient double lock / double lock
free checks.

Also fixed the type of button sent so it's more inline with what Awesome expects.

Turned off old Rust code from compiling.


Timidger:
Gonna leave this one open a bit longer since apparently I keep finding holes

PR #644, opened 2019-08-13
Keybindings

Timidger:
All keys go to the special Awesome client. The client can register for keys to be not sent to other clients as well.


Timidger:
Build is broken because of libinput being updated by our wlroots being behind, need to repin it

PR #649, opened 2019-10-03
layer-shell: Move surface to desired layer

Timidger:
TODO: Update the submodule version


Timidger:
Yes this is, I just haven't worked on Way Cooler for a little while now. I tend to work on Way Cooler in bursts (as is evident by [this graph](https://github.com/way-cooler/way-cooler/graphs/contributors)) so I plan to get back to this when I feel like it. If this is blocking some work someone else is doing I can get off my ass and merge this though :smile:

